PR #65, opened 2026-03-14
fix(a11y): add visible focus rings on buttons and links

FagnerMartinsBrack:


FagnerMartinsBrack:
@claude address low priority issue

PR #194, opened 2026-04-30
Add unified progress bar with client-side animation

FagnerMartinsBrack:
## Summary
Implements a unified progress bar for the article-body that animates smoothly between server updates by extrapolating linearly from observed crawl and summary pipeline stages. The bar spans 0–100% across both pipelines without regressing when control transfers from crawl to summary.

## Key Changes

- **Progress mapping** (`progress-mapping.ts`): Defines stage-to-percentage mappings for crawl (5–55%) and summary (65–100%) stages, ensuring the bar never regresses when crawl completes and summary takes over.

- **Server-side progress tracking**: 
  - Added `markCrawlStage()` and `markSummaryStage()` methods to persist pipeline stages in DynamoDB
  - Extended `ArticleCrawl` and `GeneratedSummary` types to include optional `stage` field
  - Modified `article-reader.ts` to compute unified progress ticks from crawl/summary state and emit them via `hx-swap-oob`

- **Progress bar component** (`progress-bar.component.ts`): Renders the bar with data attributes carrying stage, percentage, and server timestamp; always emits the element so HTMX OOB swaps have a target.

- **Client-side animation** (`progress-bar.client.ts`):
  - Anchors the bar at SSR-supplied percentage and timestamp
  - Computes linear rate between consecutive server ticks
  - Projects percentage forward on each `requestAnimationFrame` tick, capped at 99% to prevent crowding 100% before server confirmation
  - Re-scans DOM on HTMX `htmx:afterSwap` to absorb fresh server ticks
  - Handles edge cases: empty SSR timestamps, clock skew, worker redelivery regressions

- **Styling** (`article-body.styles.css`): Added 4px progress bar with 150ms linear transition for smooth per-poll jumps and continuous sub-second rAF updates.

- **Build integration** (`build-client-bundles.js`): Bundled `progress-bar.client.ts` as a global `ProgressBar` module initialized on page load with HTMX swap listener.

## Notable Implementation Details

- The bar uses a two-tier percentage scale: crawl stages occupy 5–55%, summary stages 65–100%, with a 10% gap ensuring no regression when pipelines switch.
- Server ticks are absorbed on HTMX `htmx:afterSwap` events, allowing the bar to update without separate polling.
- Rate computation floors at 0 to handle worker redeliveries that might report a lower percentage.
- The bar is always rendered (visible or hidden via CSS class) so OOB swaps targeting `#article-body-progress` always succeed.
[...]

PR #201, opened 2026-04-30
feat(hutch,save-link): unified page-loading progress bar

FagnerMartinsBrack:
## Summary
- Adds a single 0–100% progress bar that animates across the crawl and summary pipelines, anchored on server polls and projected forward via `requestAnimationFrame` between polls.
- Server emits a `hx-swap-oob` bar on every reader/summary poll response so the bar updates without a separate poll cycle.
- Reimplementation of #194 against current `main`, with the three low-priority review observations fixed in the implementation rather than left as follow-ups.

## Low-priority issues addressed (vs #194)
1. **OOB attribute injection no longer regex-based.** `progress-bar.template.html` carries an optional `{{#if oob}} hx-swap-oob="outerHTML"{{/if}}`; `renderProgressBarOob` calls `render()` with `oob: true` instead of `.replace()`-ing emitted HTML.
2. **Dropped redundant `markCrawlStage("crawl-ready")`** after `markCrawlReady` in `save-link-work.ts`. Once `crawl.status === "ready"` the bar already switches to summary, so the stage write was unobservable. `crawl-ready` removed from the `CrawlStage` union and percentage table.
3. **Dropped `summary-content-loaded`** — the original PR wrote it back-to-back with `summary-generating`, so a 3s poll never observed it. Removed both the write and the mapping entry.
4. **Dropped `summary-complete`** — written after `saveGeneratedSummary` flips status to `ready`, at which point the bar is already hidden. Removed for the same reason.

## Mapping after dropping unobservable stages
- Crawl (5 stages): `crawl-fetching` 5, `crawl-fetched` 17, `crawl-parsed` 29, `crawl-metadata-written` 41, `crawl-content-uploaded` 53
- Summary (2 stages): `summary-started` 65, `summary-generating` 90
- Client-side cap at 99% so the bar never crowds 100% before the server confirms a terminal state

## Notable implementation details
- Worker writes are unconditional (`SET crawlStage = :stage` / `SET summaryStage = :stage`); SQS redelivery just rewrites the same monotonic sequence.
- `buildUnifiedProgress` returns `undefined` when both pipelines are terminal, or when the crawl has failed (rendering a half-full bar that never advances would look broken).
- Empty-`tickAt` SSR fallback is honored: the client anchors against `deps.now()` so the first real server tick still produces a positive elapsed window for `computeRate`.
- `markCrawlPending` preserves any previously recorded stage in the in-memory provider (mirrors the DDB `UpdateExpression` which only touches `crawlStatus`), so the legacy-stub healing path doesn't reset the bar.

## Test plan
- [x] `pnpm check` — 16 projects, all coverage thresholds met (100% functions, overall 99.51% statements / 97.16% branches)
- [ ] Manual: save a fresh URL, watch the bar march from 5% through ~53% as crawl progresses, jump to 65% when crawl flips ready, project toward 99% during the summary wait, and disappear when the summary lands
- [ ] Manual: save a URL that fails to crawl — bar should hide once crawl flips `failed`
- [ ] Manual: re-save a URL with a cached summary — bar should not appear (both pipelines already terminal)

### DIFF
--- a/projects/hutch/scripts/build-client-bundles.js
+++ b/projects/hutch/scripts/build-client-bundles.js
@@ -45,6 +45,29 @@ const BUNDLES = [
 			"}).attach();",
 		].join("\n"),
 	},
+	{
+		entry: path.join(
+			PROJECT_ROOT,
+			"src/runtime/web/shared/article-body/progress-bar.client.ts",
+		),
+		outfile: path.join(OUT_DIR, "progress-bar.client.js"),
+		globalName: "ProgressBar",
+		footer: [
+			// HTMX fires htmx:afterSwap on the container around any swapped node,
+			// including OOB swaps targeting #article-body-progress. The listener
+			// re-anchors the bar against the new (tickAt, pct) so the rAF loop
+			// projects forward from there.
+			"ProgressBar.initProgressBars({",
+			"  document: window.document,",
+			"  now: function () { return Date.now(); },",
+			"  requestAnimationFrame: function (cb) { return window.requestAnimationFrame(cb); },",
+			"  cancelAnimationFrame: function (id) { window.cancelAnimationFrame(id); },",
+			"  addSwapListener: function (listener) {",
+			"    window.document.body.addEventListener('htmx:afterSwap', listener);",
+			"  }",
+			"});",
+		].join("\n"),
+	},
 ];
 
 function buildOptions(bundle) {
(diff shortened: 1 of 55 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #201
- **Commit**: 4764e8dcee7ecfdd9e52a43d50217c7d5a95e92e

FagnerMartinsBrack:
@claude addres low prio issues 1 and 2, ignore 3 and 4

FagnerMartinsBrack:

## Context
- **PR**: #201
- **Commit**: 33c31979b8be51fc07b71a25117e66242f044b08

PR #253, opened 2026-05-10
feat: move /view freshness check to background stale-check Lambda

FagnerMartinsBrack:
## Summary
- /view's synchronous `refreshArticleIfStale` call is moved to a new SQS-backed Lambda (`stale-check-requested`). /view now renders immediately from DDB and publishes `StaleCheckRequestedEvent`; the background worker decides whether to republish `SaveAnonymousLinkCommand` to redrive the crawl pipeline.
- New event/handler/composition root in save-link, in-memory + EventBridge publishers in test-fixtures + hutch, and DDB readers (`findArticleFreshness`, `findArticleCrawlStatus`) duplicated in save-link to avoid cross-project imports per CLAUDE.md.
- Phase 1 of 4 in the unstick-articles plan. Later phases bump Lambda budgets, cap auto-heal retries, and tune SQS `maxReceiveCount`.

## Why
Two failure modes were stranding URLs after a successful save:
- Mode A (iana media-types XHTML): structural 30s Lambda timeout overflow → DLQ flips row to `crawlStatus='failed'`. Phase 2 fixes this.
- Mode B (Medium-hosted articles like fagnerbrack.com): `/view` blocks 5–30s on every visit because the freshness/conditional-GET path runs inline on the request. **This PR fixes Mode B.**

The introducing commit for Mode A's loop is 2947e396 — its "reprime" branch turned a one-time `failed` into an infinite `/view click → republish → timeout × 3 → DLQ` loop. Moving the freshness check to a background worker decouples the user request path from that loop.

## What does NOT change here
- iana stays `failed/failed`. Lambda budget is unchanged → Phase 2.
- The three Medium URLs stay in their current ready/ready or ready/failed-summary state (no regression expected).
- `refreshArticleIfStale` still runs inline for `/queue`, `/import`, and `/save` — those flows weren't reported as a UX hang.

## Test plan
- [x] Local: lint + typecheck + tests pass for save-link, hutch, test-fixtures, hutch-infra-components
- [x] Local: coverage thresholds met (save-link 100% across 57 files; hutch 99.51/96.68/100/99.51 vs targets 99/96/100/99)
- [x] New `stale-check-handler.test.ts`: 6 specs covering reprime / new / skip / unchanged / refreshed / batch / invalid-detail
- [x] `view.route.test.ts`: 5 affected specs rewritten to assert on the new "render-immediately + publish-stale-check" contract
- [ ] After merge: deploy to prod
[...]

PR #259, opened 2026-05-10
feat(save-link,hutch-infra-components): SQS partial batch failure reporting in stale-check handler

FagnerMartinsBrack:
## Summary

Addresses the low-priority follow-up flagged on the merged stale-check Lambda PR (#253):

> Consider SQS partial batch failure reporting in stale-check handler — `projects/save-link/src/save-link/stale-check-handler.ts:15-35`

The handler now returns `SQSBatchResponse` with `batchItemFailures`, so when one record in a batch errors the others are still settled and only the failed record is retried. Previously the whole batch was retried on a single failure (safe, given upstream idempotency, but redundant).

## What changed

- **`projects/save-link/src/save-link/stale-check-handler.ts`** — wraps the per-record processing in a `try/catch`, pushes failed `messageId`s into `batchItemFailures`, and returns `{ batchItemFailures }` instead of throwing on the first failure.
- **`src/packages/hutch-infra-components/src/infra/hutch-sqs-backed-lambda.ts`** — adds an opt-in `reportBatchItemFailures?: boolean` flag that wires `functionResponseTypes: ["ReportBatchItemFailures"]` on the `EventSourceMapping`. Without this flag set, AWS would silently treat the response as a successful invocation and skip the DLQ entirely — so the handler change and the infra change must land together.
- **`projects/save-link/src/infra/index.ts`** — passes `reportBatchItemFailures: true` to the stale-check `HutchSQSBackedLambda`. All other SQS-backed Lambdas keep their existing throw-to-redrive semantics (unchanged).
- **`projects/save-link/src/save-link/stale-check-handler.test.ts`** — updates the existing batch test + the zod-validation test to assert the response shape, and adds a new spec that exercises the partial-failure path: two records, the second one rejects from `refreshArticleIfStale`, and the response reports only `msg-2`.

## Why both layers

`batchSize: 1` is hardcoded today on `HutchSQSBackedLambda`, so the runtime impact for the stale-check Lambda is currently nil. The handler change is correctness-preserving for the for-loop pattern (so a future `batchSize > 1` does not regress) and the EventSourceMapping config is what actually instructs AWS to honor the response — without `functionResponseTypes` set, returning `{ batchItemFailures: [...] }` is treated as success and the failed record would be silently dropped before the DLQ alarm fires.

## Test plan

- [x] `pnpm nx run save-link:test --testPathPattern=stale-check-handler` → 8 specs pass (5 existing + 1 updated batch + 2 new failure-reporting specs)
- [x] `npx jest --coverage --collectCoverageFrom='dist/save-link/stale-check-handler.js'` → 100% statements/branches/functions/lines on the handler
- [x] `pnpm nx run save-link:lint` and `pnpm nx run @packages/hutch-infra-components:lint` → clean
- [x] Pre-commit `pnpm check` (run-many across 18 projects) → green
- [ ] After merge: confirm the deployed `stale-check-requested` EventSourceMapping has `FunctionResponseTypes: ["ReportBatchItemFailures"]` (`aws lambda list-event-source-mappings --function-name stale-check-requested-handler`)
- [ ] After merge: force a handler error (e.g. transient DDB throttle) and verify only the failing message is retried + DLQ alarm still fires after `maxReceiveCount`

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hc9GFNQBHbfHFhdRCbZ6wz)_

### DIFF
--- a/projects/save-link/src/infra/index.ts
+++ b/projects/save-link/src/infra/index.ts
@@ -287,6 +287,7 @@ const staleCheckRequestedLambdaWithSQS = new HutchSQSBackedLambda("stale-check-r
 	lambda: staleCheckRequestedLambda,
 	queue: staleCheckRequestedQueue,
 	alertEmailDLQEntry: alertEmail,
+	reportBatchItemFailures: true,
 });
 
 eventBus.subscribe(StaleCheckRequestedEvent, staleCheckRequestedLambdaWithSQS);
--- a/projects/save-link/src/save-link/stale-check-handler.test.ts
+++ b/projects/save-link/src/save-link/stale-check-handler.test.ts
@@ -56,10 +56,15 @@ describe("initStaleCheckHandler", () => {
 			logger: noopLogger,
 		});
 
-		await handler(createSqsEvent({ url: URL_UNDER_TEST }), stubContext, () => {});
+		const result = await handler(
+			createSqsEvent({ url: URL_UNDER_TEST }),
+			stubContext,
+			() => {},
+		);
 
 		expect(publishSaveAnonymousLink).toHaveBeenCalledTimes(1);
 		expect(publishSaveAnonymousLink).toHaveBeenCalledWith({ url: URL_UNDER_TEST });
+		expect(result).toEqual({ batchItemFailures: [] });
 	});
 
 	it("re-publishes SaveAnonymousLinkCommand when refreshArticleIfStale returns 'new' (e.g. row was evicted between view and worker)", async () => {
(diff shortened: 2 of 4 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #259
- **Commit**: 8719cb3905e30deee0fcead681b300c661392ac9

FagnerMartinsBrack:
@claude analyse low prio issues and consider if they need to be addressed

FagnerMartinsBrack:
@claude does stale-check-handler handles only one URL? Why is batch necessary here?

PR #268, opened 2026-05-10
fix(hutch): re-show onboarding when the extension is uninstalled

FagnerMartinsBrack:
## Summary

Once a user installed the extension, dismissed onboarding, then uninstalled (or wiped their profile / switched browser / cleared part of their cookie jar), the install + save-first-article checkboxes stayed ticked for up to a year and the onboarding banner stayed hidden. The product silently lied about what was actually installed.

## Why the bug existed

`hutch_ext_installed` had a 365-day TTL and was renewed by **two** independent paths:
- The extension's content script writing `document.cookie = "hutch_ext_installed=1; max-age=31536000"` on every Readplace page load (deployed code, not changeable without republishing the Chrome and Firefox extensions — weeks of store review).
- The hutch middleware setting it on every Siren request.

Neither path cleared the cookie when the extension was uninstalled. `hutch_ext_saved` had the same shape (365 days, no liveness check). The dismissal logic at `queue.page.ts:156` is gated on `extensionInstalled && dismiss === ONBOARDING_VERSION` — that branch was already correct, but `extensionInstalled` never flipped back to `false` in production.

## Fix (server-only, no extension republish)

1. **New `hutch_ext_alive` cookie** — `httpOnly`, **30-day TTL**, set by the existing `mark-extension-installed.middleware.ts` on every Siren request. Because it's `httpOnly`, the extension's content script (or any other JS) cannot forge or renew it via `document.cookie`. The only way it stays set is for an installed extension to keep making Siren requests.
2. `isExtensionInstalled(req)` now reads `hutch_ext_alive` instead of `hutch_ext_installed`. The legacy cookie is still written by deployed extensions but is intentionally ignored on the server.
3. **Cut `hutch_ext_saved` from 365d → 30d** and renew it from the same middleware while present, so the "save first article" milestone tracks extension liveness instead of persisting for a year past uninstall.
4. The dismiss cookie can stay 1-year — once `extensionInstalled` flips back to `false`, the dismissal is implicitly ignored and the onboarding re-renders.

## Behaviour change worth flagging

Users who install the extension but never open the popup (or use the Cmd/Ctrl+D shortcut) will see the install step show as **incomplete** until they actually use the extension once. Today the content-script's page-load write makes the step flip to complete just from loading any Readplace page — even if the user never touched the toolbar. Under this PR, the trigger is "the user actually invoked the extension" instead of "the extension was loaded by the browser". This is arguably more honest, and aligns the install step with the save-first-article step (which has always required a real extension action).

## Test plan

- [x] `pnpm nx run hutch:check` (lint + 100% coverage) — passes
- [x] `pnpm nx run hutch:test-integration` — passes
- [x] **New regression test**: `re-renders onboarding after uninstall (dismiss + legacy cookie present, alive cookie missing)` — pins the exact bug scenario
- [x] **New regression test**: `does not mark install-extension complete when only the legacy hutch_ext_installed cookie is present` — pins that the server stops trusting the content-script-written cookie
- [x] **New middleware tests**: `sets the alive cookie as httpOnly`, `renews the hutch_ext_saved cookie on a Siren request when it is already present`, `does not set hutch_ext_saved on a Siren request when the cookie is absent`
- [ ] Verified locally: install extension, dismiss onboarding, uninstall, return to `/queue` → onboarding visible with install step incomplete (within 30 days of last Siren call)

https://claude.ai/code/session_018razYuhVfZf5t8DtETBK8T

---
_Generated by [Claude Code](https://claude.ai/code/session_018razYuhVfZf5t8DtETBK8T)_

### DIFF
--- a/projects/hutch/src/e2e/queue-flow/onboarding-actions.ts
+++ b/projects/hutch/src/e2e/queue-flow/onboarding-actions.ts
@@ -1,7 +1,7 @@
 import { expect } from '@playwright/test'
 import {
-  COOKIE_NAME,
-  COOKIE_VALUE,
+  ALIVE_COOKIE_NAME,
+  ALIVE_COOKIE_VALUE,
   SAVE_COOKIE_NAME,
   SAVE_COOKIE_VALUE,
 } from '@packages/onboarding-extension-signal'
@@ -29,8 +29,8 @@ export function createOnboardingActions(
         await expect(step).toHaveAttribute('data-test-onboarding-complete', 'false')
 
         await page.context().addCookies([{
-          name: COOKIE_NAME,
-          value: COOKIE_VALUE,
+          name: ALIVE_COOKIE_NAME,
+          value: ALIVE_COOKIE_VALUE,
           path: '/',
           domain: new URL(page.url()).hostname,
         }])
--- a/projects/hutch/src/runtime/web/api/api.routes.test.ts
+++ b/projects/hutch/src/runtime/web/api/api.routes.test.ts
@@ -490,8 +490,8 @@ describe("POST /queue/:id/delete (Siren)", () => {
 	});
 });
 
-describe("extension-installed cookie middleware", () => {
-	it("sets the extension-installed cookie on the Siren entry point before login", async () => {
+describe("extension-alive cookie middleware", () => {
+	it("sets the alive cookie as httpOnly on the Siren entry point before login", async () => {
 		const testApp = createTestApp(createDefaultTestAppFixture(TEST_APP_ORIGIN));
 
 		const response = await request(testApp.app)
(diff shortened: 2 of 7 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #268
- **Commit**: 35f0ba505e54472b66aa1bfbf6ef5ff2b8e88673

FagnerMartinsBrack:
@claude address low prio issues and consider if the design can be improved to reduce the number of comments by centralising the logic in such a way it's clear to understand by looking up the places it's using the methods

FagnerMartinsBrack:

## Context
- **PR**: #268
- **Commit**: 930a06769d33e8724baf5e1d57c3d3f9a618c2fc

PR #285, opened 2026-05-11
feat: introduce typed Article aggregate, migrate refresh-article-content (Phase 1)

FagnerMartinsBrack:
## Summary

Phase 1 of the article aggregate redesign. Adds a typed Article aggregate plus a `load → transition → save → dispatch` orchestrator in `@packages/domain` and migrates the `refresh-article-content` writer (the trigger of the 2026‑05‑10 forever‑pending bug) end‑to‑end onto it.

The transition function returns `{ article, effects }` as one value, so a writer that resets summary state without dispatching `GenerateSummaryCommand` is a missing‑effect unit‑test failure rather than a stuck production row.

Phase 1 is intentionally narrow:
- Only `refresh-article-content` is migrated. The other 16 writers still issue their own `UpdateExpression`s.
- No `version` attribute, no concurrency check — last‑writer‑wins, per the design discussion.
- The DDB store's `save` writes only the fields the `refreshContent` transition mutates (metadata, freshness, estimatedReadTime, summary). Crawl state is loaded into the aggregate but not written back so a concurrent crawl-state writer cannot be clobbered. Future transitions that mutate crawl will extend the `UpdateExpression`.
- No watchdog and no architecture snapshot needed — wire-format `defineEvent`/`defineCommand` declarations are unchanged. The Lambda still subscribes to `RefreshArticleContentCommand` and dispatches `GenerateSummaryCommand`; only the internal handler structure changed.

## What changed

| Layer | Files |
|---|---|
| Aggregate types + transitions + orchestrator (pure) | `src/packages/domain/src/article-aggregate/{article.types,effects.types,storage.types,effect-dispatcher.types,transition-and-persist,transitions/refresh-content,index}.ts` |
| In-memory test doubles | `src/packages/test-fixtures/src/providers/article-aggregate/{in-memory-article-store,in-memory-effect-dispatcher,index}.ts` |
| DynamoDB store + Lambda effect dispatcher | `projects/save-link/src/article-aggregate/{dynamodb-article-store,lambda-effect-dispatcher}.ts` |
| Migrated handler + composition root | `projects/save-link/src/save-link/refresh-article-content-handler.ts`, `projects/save-link/src/runtime/refresh-article-content.main.ts` |
| Replaced (deleted) | `projects/save-link/src/save-link/refresh-article-content.ts` (the inline `UpdateExpression`) |

## Why this fixes the 2026‑05‑10 regression

The previous handler issued an inline `UpdateExpression` that simultaneously REMOVEd every summary attribute and SET `summaryStatus = "pending"`, then separately dispatched `GenerateSummaryCommand` from the handler body. A change to one limb without the other left rows stuck on "Generating summary…". The aggregate folds both into a single transition return value: the orchestrator runs `save` then `dispatch`, and unit tests assert that `refreshContent({...}).effects` contains the dispatch — so a future writer that drops it fails before the change reaches production.

## Test plan

- [x] `pnpm check` passes for all 18 projects (the 3 pre-existing failures on `main` — `@packages/article-resource-unique-id`, `@packages/test-phase-runner`, `@packages/crawl-article` — are unrelated to this PR; verified by stashing changes and re-running on a clean tree)
- [x] 100 % statement / branch / function / line coverage on every new/changed file
- [x] In-memory end-to-end test demonstrates: seed → handle SQS event → aggregate flips summary to pending → `generate-summary` effect dispatched (in that order)
- [x] Save → dispatch ordering test guards the worker race the previous handler protected against
- [x] Handler returns batch failure (does not dispatch) on Zod parse failure, store load returning `undefined`, and store save throwing
- [x] DynamoDB store unit-tested with fake `DynamoDBDocumentClient`: `UpdateExpression` writes metadata/freshness/estimatedReadTime/summary fields and never touches crawl attributes

https://claude.ai/code/session_01TzuH2JW3jkq7DchELqki5C

---
_Generated by [Claude Code](https://claude.ai/code/session_01TzuH2JW3jkq7DchELqki5C)_


FagnerMartinsBrack:

## Context
- **PR**: #285
- **Commit**: c571bf78091c142e501643d50d9990861c4e3639

FagnerMartinsBrack:
@claude address low prio 1. Exhaustive dispatch in lambda-effect-dispatcher.ts

FagnerMartinsBrack:
@claude ## Adversarial cases to check


These should be in the unit-test fixtures up front so a future relaxation doesn't silently re-open a hole.

- **Effect dispatch fails after store.save() succeeds.** Designed-for behaviour: handler throws → SQS retries → next invocation re-runs the transition against the now-persisted state and re-dispatches the effects. Effects MUST be idempotent in their consumers (already true for the current EventBridge events). Test the redelivery path against the in-memory dispatcher.

- **Tier-0 (extension) vs Tier-1 (server) racing on the same URL.** submitLinkContent and submitLink both produce a valid transition; the second sees the version conflict and rebases against the now-Tier-0-populated aggregate. selectMostCompleteContent runs unchanged (this plan does not touch the selector logic).

- **DLQ handler runs against a row that already reached terminal state via a sibling path.** markCrawlExhausted transition is only valid from {kind:"pending"} and {kind:"failed"} — receiving it on {kind:"ready"} returns a no-op {article: unchanged, effects: []} rather than throwing, so a late DLQ arrival can't regress a ready row.

- **Client-abuse: repeated submitLink for the same URL by a script.** TTL gate dedups; recrawl-via-TTL is the only way to force a re-fetch, and it requires /admin/recrawl (logged-in admin). No new abuse surface vs today.

- **Server-decoy: submitLinkContent from extension with garbage HTML.** Tier-0 still goes through selectMostCompleteContent; Tier-1 wins on quality. Same as today.


Report on these after reviewing the code

FagnerMartinsBrack:

## Context
- **PR**: #285
- **Commit**: fb4ca90e4fd7575a3fb188cd75b70f78fe7eead8

FagnerMartinsBrack:
@claude address this low prio only: 2. Redelivery-cycle integration test (noted in adversarial analysis)

FagnerMartinsBrack:

## Context
- **PR**: #285
- **Commit**: 1804b2926ea200f597f8599b2072d8ddaeb6aba0

PR #286, opened 2026-05-11
refactor(@packages/check-stuck-articles): only flag pending states

FagnerMartinsBrack:
## Summary

- Strip the legacy-stub, writer-contract (`summaryStatus='ready'` AND `attribute_not_exists(summary)`), `originalUrl`-skip, and domain-exclusion paths from the canary.
- Reduce `classifyRow` reasons to `summary-pending` / `crawl-pending` and drop the `summary` field from the row inputs of `classifyRow` / `checkTerminalState` / the DDB projection.
- Make `originalUrl` and `savedAt` required in the row schema (no more `dynamoField` opt-in); drop `summary`, `summaryFailureReason`, `crawlFailureReason` from projection and `StuckRow`.
- Delete `exclude-patterns.ts` and the `skippedNoOriginal` / `excludedDomain` counters.
- Trim the failing-test message and the GitHub Actions issue body line to drop the now-always-empty `failure:` segment.

The goal is the simplest possible "stuck in non-terminal state" detector: surface rows whose `summary` or `crawl` state machine never moved past `pending` past the 20-min age gate, and nothing else. Legacy data shape is no longer the canary's concern.

## Test plan

- [x] `pnpm nx run @packages/check-stuck-articles:check` — 42/42 green locally (tsc lint + biome + knip + 4 node:test files)
- [ ] First scheduled run on `main` after merge — expect the report to either come back clean or surface any pending rows that the legacy-stub / no-originalUrl filters were previously masking. The last `main` run masked 1,165 no-`originalUrl` and 4 domain-excluded rows; with this PR those filters are gone, so an over-pending row that fell into either bucket would now page.
- [ ] Confirm no row currently has `summaryStatus='pending'` AND missing `originalUrl` in prod — the simplified Zod schema is strict and a row like that would hard-fail `schema.parse()` on the first run instead of being silently skipped (loud failure is the desired behaviour, but worth flagging before merge).

PR #274, opened 2026-05-11
feat: validate URLs at intake with shared SaveableUrlSchema

FagnerMartinsBrack:
## Summary

Closes class #5 sub-types 1–3 from `article-aggregate-design-gaps-2026-05-11.md`: reject `chrome://`, `*.home.arpa`, and malformed URLs **before** they become article rows.

A URL string that the system cannot meaningfully crawl never gets a DDB row, never enters the SQS pipeline, and never fires the canary. The user/operator is told *why* at the HTTP boundary.

## Design

Single shared `validateSaveableUrl()` in `@packages/domain/article` returns a discriminated `Result`:

```ts
type SaveableUrlResult =
  | { status: "SUCCESS"; url: SaveableUrl }
  | { status: "ERROR"; error: { code: SaveableUrlErrorCode; message: string } };
```

Error codes are stable contracts: `unsupported_scheme`, `private_network`, `malformed_url`. `SaveableUrl` is a Zod-branded type so callers can't bypass the validator and hand a raw string to `saveArticleFromUrl`.

The matching `SaveableUrlSchema` re-exposes the validator as a Zod schema, carrying the error code via `params.saveableUrlCode` on custom issues for callers that need to branch on failure kind.

## Wiring (Phase 1 + 2)

| Entry point | Behaviour on rejection |
|---|---|
| `POST /queue/save` (form) | 422 — re-renders `/queue` with the human-readable message in the existing save-error banner. |
| `POST /queue` (Siren API) | 422 — when client opts in via `Prefer: return=representation`, returns the current collection with `properties.warning: { code, message }` so the extension can drop the user into the list view *and* render a warning. Legacy chrome-extension v1.0.66 (no `Prefer`) keeps the pre-existing stub-save behaviour via a separate `saveUnsaveableUrlStub`. |
| `GET /view/<url>` / `GET /view?url=` | Existing save-error page; never creates the anonymous stub row. |
| `POST /queue/save-html` (URL + URL-only fallback) | 422 `invalid-save-html` for non-saveable URLs. |
| `POST /import/:id/commit` | Filters selected URLs through the validator; commits the saveable ones, stores the skipped URLs in a read-once `import_skipped` cookie. `/queue` renders them as a *"Couldn't import these links"* banner with the per-URL reason. |

## Extension client

[...]

PR #300, opened 2026-05-13
chore(@packages/check-stuck-articles,hutch): retire exclude-patterns debris and add form-level URL-validation regression

FagnerMartinsBrack:
## Goal

Close out the two remaining items from the URL-validation hardening exercise after PR #274 and the existing 31-case `saveable-url.test.ts`:

1. Retire the last stale `exclude-patterns` reference in source control.
2. Cover the canary-historical rejection classes at the form-submit round-trip level so a future refactor that breaks the typed-error branch fails a regression instead of breaking the user-facing flow silently.

## Context

Class #5 of `docs/article-aggregate-design-gaps-2026-05-11.md` listed four sub-types of unsaveable URL: browser-internal (`chrome://`), local-network (`*.home.arpa`), malformed (`[www.theinformation](https://www.theinformation)....`), and permanently non-HTML (`*.pdf`). The first three are now blocked at intake — confirmed by `saveable-url.test.ts`'s `describe("legacy canary exclude patterns (the ones intake should now block)")`. The fourth stays with the aggregate via `markCrawlUnsupported`.

Two loose ends remained:

- `src/packages/check-stuck-articles/knip.config.ts:8` still listed `scripts/exclude-patterns.ts` as an entry — the only `exclude-patterns` reference left in source control after the source module and dist artefacts went away. A reader grepping for `EXCLUDE_PATTERNS` would still find it and think the canary still maintains a runtime exclude list.
- The unit tests in `saveable-url.test.ts` cover `validateSaveableUrl`'s return value but not the form HTTP round-trip. A future refactor that breaks the 422 status or the typed-code branch passes unit tests while silently breaking the user-facing flow.

## What changed

- **knip cleanup**: removed the stale `scripts/exclude-patterns.ts` entry from `@packages/check-stuck-articles`'s knip config — the source file is gone and the build doesn't regenerate the corresponding dist artefacts (verified by running `pnpm nx run @packages/check-stuck-articles:check` and checking `dist/scripts/` after).
- **Typed error code on the error pill**: surfaced `SaveableUrlErrorCode` through `QueueViewModel.saveErrorCode` → `queue.component.ts` → `queue.template.html`, rendering a new `data-test-saveable-url-code="<code>"` attribute on the existing `[data-test-save-error]` pill. Behaviour unchanged; the attribute is only present when a saveable-url validation produced the error (not for `error_code=save_failed` redirects).
- **Form-level regression**: extended `queue.route.test.ts` with a `describe("form-level URL-validation regression (canary-historical inputs)")` block that walks each of the brief's eight inputs (`chrome://extensions/`, `about:blank`, `https://cd.home.arpa/x`, `http://localhost:3000/x`, `https://192.168.1.1/x`, `www.theinformation....`, `https://server`, `""`) and asserts the form-submit round-trip:
  - HTTP status is 422,
  - the error pill carries the matching typed code,
  - no article row was stored against the user.

## Implementation note: integration over e2e

The original brief proposed a Playwright e2e spec. After investigation, the form-level regression is covered by integration tests instead because:

1. CLAUDE.md says **"Use integration tests for comprehensive filter/query functionality testing, not E2E tests"** — and the existing `queue.route.test.ts` already supertest-tests the chrome:// and localhost cases at the form level, so the patterns were already in place.
2. The e2e server at `projects/hutch/src/e2e/e2e-server.main.ts:41` deliberately overrides `validateSaveableUrl` so the local server can save its own `localhost` test URLs — which means an e2e Playwright spec would silently see the three `private_network` cases (`*.home.arpa`, `localhost`, `192.168.1.1`) succeed with 201, not 422, and would never have been a real regression.

The integration test runs in <2s and exercises the same handler, viewmodel, component, and template as the production form post.

## Verification

- `pnpm nx run @packages/check-stuck-articles:check` → 51 tests pass, `dist/scripts/exclude-patterns.*` does not regenerate.
- `pnpm nx run hutch:check` → 108 tests pass in `queue.route.test`, all coverage thresholds met (172 files: 99.69 / 96.15 / 100 / 99.69).
- Pre-commit hook ran the full monorepo `pnpm check` (18 projects), all green.

## Out of scope (deliberate)

- Adding new rejection classes — content-type rejection stays with the crawler/aggregate via `markCrawlUnsupported`.
- Refactoring `validateSaveableUrl` itself.
- Re-introducing the canary's exclude-pattern list (PR #286 retired it).
- Touching `ArticleResourceUniqueId.parse` or the tracking-param stripper.
- Adding more unit tests — `saveable-url.test.ts` already covers every documented rejection class with 31 cases.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RsJrZBjgzruGeAMkxWVhCA)_

### DIFF
--- a/projects/hutch/src/runtime/web/pages/queue/queue.component.ts
+++ b/projects/hutch/src/runtime/web/pages/queue/queue.component.ts
@@ -16,6 +16,7 @@ interface QueueDisplayModel {
 	total: number;
 	pluralSuffix: string;
 	saveError?: string;
+	saveErrorCode?: string;
 	importFlash?: string;
 	hasImportSkipped: boolean;
 	importSkippedEntries: ReadonlyArray<{ url: string; reasonLabel: string }>;
@@ -67,6 +68,7 @@ function toQueueDisplayModel(vm: QueueViewModel, options: { extensionInstalled: 
 		total: vm.total,
 		pluralSuffix: vm.total !== 1 ? "s" : "",
 		saveError: vm.saveError,
+		saveErrorCode: vm.saveErrorCode,
 		importFlash: vm.importFlash,
 		hasImportSkipped: Boolean(vm.importSkipped && vm.importSkipped.entries.length > 0),
 		importSkippedEntries: vm.importSkipped?.entries ?? [],
--- a/projects/hutch/src/runtime/web/pages/queue/queue.page.ts
+++ b/projects/hutch/src/runtime/web/pages/queue/queue.page.ts
@@ -408,6 +408,7 @@ export function initQueueRoutes(deps: QueueDependencies): Router {
 			]);
 			const vm = toQueueViewModel(result, urlState, {
 				saveError: validation.error.message,
+				saveErrorCode: validation.error.code,
 				unreadCount,
 				summaryByUrl,
 				crawlByUrl,
--- a/projects/hutch/src/runtime/web/pages/queue/queue.route.test.ts
+++ b/projects/hutch/src/runtime/web/pages/queue/queue.route.test.ts
@@ -146,6 +146,42 @@ describe("Queue routes", () => {
 			expect(doc.querySelector("[data-test-save-error]")?.textContent).toMatch(/[Pp]rivate-network/);
 		});
 
+		describe("form-level URL-validation regression (canary-historical inputs)", () => {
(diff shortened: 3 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #300
- **Commit**: e5ddbd3a4a924d3cb9bb45eb77a79be614ea6436

PR #351, opened 2026-05-18
fix(hutch): hand off polling between reader and summary slots so progress bar can hide

FagnerMartinsBrack:
## Summary

Found while investigating a stuck progress bar on `/admin/recrawl/<cutlefish-substack-url>` — the DDB row was healthy (`crawlStatus=ready`, `summaryStatus=ready`) but the admin recrawl page kept rendering a half-filled bar that `/view/<same-url>` correctly hid.

**Root cause.** The recrawl pipeline transitions a row through `crawl: pending → ready` and (when canonical hash changes) `summary: ready → pending → ready`. The reader poll observes both axes for its OOB progress bar but only continued its OWN polling chain. When the reader poll's stop condition (`crawl=ready` + content present) coincided with summary being mid-regeneration, the bar's last update showed summary stage VISIBLE, then no further poll ever observed summary completing — because the summary slot's HTMX trigger was never armed at page load (summary was `ready` then). The chain broke silently and the bar stranded visible.

**Fix.** Each poll response now emits the sibling slot as an `hx-swap-oob` fragment with the appropriate poll URL, so any axis that transitions to `pending` while the primary axis is settling re-arms its own polling chain. Symmetric in both directions (reader → summary and summary → reader).

**Type-level enforcement.** A bare runtime fix would regress: both poll handlers now route through a shared `renderPollResponseBody(input)` whose input shape requires every axis' state, both poll URLs, and a `primary: 'reader' | 'summary'` discriminator. There is no path through `article-reader.ts` that can emit a poll response containing only one slot. `HandlePollParams.extensionInstallUrl` is required (`string | undefined`, not optional) so all three reader pages — admin recrawl, public `/view`, private `/queue/:id/read` — must explicitly wire it into both `handleSummaryPoll` and `handleReaderPoll`. Every slot variant template renders with a stable `id` (`article-body-summary-slot` / `article-body-reader-slot`) so HTMX OOB can target the slot regardless of current variant.

## Test plan

- [x] Unit tests cover the cross-axis handoff (5 new tests under `cross-axis poll handoff`):
  - Reader-poll observes summary going pending → emits OOB summary slot with `hx-get`
  - Summary-poll observes crawl going pending → emits OOB reader slot with `hx-get`
  - Both poll handlers emit terminal sibling OOB (no `hx-get`) when sibling axis is already settled
  - Full admin-recrawl scenario: reader poll about to settle while summary just flipped to pending hands off
- [x] Full hutch suite: 1280 tests passing, 99.65% line / 100% function coverage
- [x] `nx affected -t test --base=main` clean
- [x] Lint + typecheck + knip + unused-css clean
- [ ] Manual verification on staging: visit `/admin/recrawl/<some-url>` and watch the bar disappear once both pipelines settle (does NOT require waiting on a refresh)


FagnerMartinsBrack:

## Context
- **PR**: #351
- **Commit**: 5ee789a53a8c843808ffc03f633f6499ff92b0cf

PR #450, opened 2026-05-28
fix(hutch): default https:// for schemeless /admin/recrawl paths

FagnerMartinsBrack:
## Summary
- Pasting `/admin/recrawl/host/path` (no scheme) into the admin recrawl path rendered the `SaveErrorPage` ("No article URL provided") and redirected back to the landing page. Root cause: `req.params[0]` arrives without a scheme, `z.url()` rejects it, and the existing `https:/` → `https://` regex only fires for collapsed-scheme paths from API Gateway.
- `handleRecrawlArticle` now prepends `https://` when the wildcard segment is missing a scheme, so admins can paste the bare article URL. The downstream `findArticleByUrl` check still 404s for URLs that aren't in the DB — no risk of stub-row creation.
- Added a regression test that hits `/admin/recrawl/example.com/post` (mirrors the bug report) and asserts the page renders and `publishRecrawlLinkInitiated` is called with the canonical `https://example.com/post`.

## Test plan
- [x] `pnpm nx run hutch:check` — lint, type-check, unit + integration tests, 100% coverage gate
- [x] New test covers the schemeless path case
- [x] Existing tests for encoded-scheme path (`/admin/recrawl/https%3A%2F%2F…`) still pass
- [ ] Manual smoke after deploy: visit `/admin/recrawl/<host>/<path>` (no scheme) on prod and confirm the recrawl page renders instead of the error

### DIFF
--- a/projects/hutch/src/runtime/web/pages/admin/recrawl.page.ts
+++ b/projects/hutch/src/runtime/web/pages/admin/recrawl.page.ts
@@ -96,7 +96,13 @@ function handleRecrawlArticle(
 		const rawPath = req.params[0];
 		// API Gateway v2 HTTP API decodes %2F to /, restore https:/ → https://
 		// (same normalisation as /view).
-		const normalizedUrl = rawPath.replace(/^(https?):\/(?!\/)/i, "$1://");
+		const restoredScheme = rawPath.replace(/^(https?):\/(?!\/)/i, "$1://");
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #450
- **Commit**: 9b5b69f14ff40dc58c12eb4d1682dbb5a00c6afe

PR #508, opened 2026-06-06
feat(reader): keep same-host in-article links in the reader tab

FagnerMartinsBrack:
## What

Inside any reader view, an in-article link that points back at our **own host** (readplace.com in prod, the staging host in staging) now loads in the **same page** instead of opening a new tab.

## Why

The reader body renders inside a sandboxed iframe whose `<base target="_top">` already sends target-less links to the parent tab. But captured article HTML frequently bakes `target="_blank"` into its anchors — and the iframe sandbox (`allow-popups`, `allow-popups-to-escape-sandbox`) deliberately keeps those `_blank` links usable. So a link **back to Readplace** would pop a *second* Readplace tab rather than navigating in place.

## How

- New pure helper `same-host-links.ts` parses the reader content with `linkedom` and rewrites **only** anchors whose absolute href host equals the deployment host to `target="_top"`. External links, relative/`mailto:`/`#fragment` hrefs, and non-anchor tags are left untouched (the same-host match is case-insensitive and port-aware, so a staging subdomain is correctly treated as a different host from prod).
- The deployment origin (`appOrigin`, already resolved from `APP_ORIGIN` at the composition root) is threaded from every reader entry point — queue reader, public `/view`, admin recrawl — **and their poll handlers** down to the iframe builder. The rewrite therefore applies both on the initial SSR render and on the poll OOB swap, so an article that finishes crawling while the reader is open also gets the treatment.
- The transform is gated on `appOrigin` being present, so it's a no-op (verbatim passthrough) when omitted — existing iframe-shell tests are unchanged.

Host detection is per-environment rather than baked in at save time, so the same saved article classifies its links correctly whether served from prod or staging.

## Testing

- New thorough unit tests for the helper (internal vs external, blank/no-target, case/port sensitivity, staging-vs-prod, relative/fragment/mailto, mixed documents, no-op, and non-anchor tag preservation).
- New tests at the iframe-builder level and an end-to-end `ReaderPage` test proving the `appOrigin` threading reaches the rendered iframe.
- `pnpm check` passes across all 25 projects, including 100% coverage on the new code (enforced via the pre-commit hook).

https://claude.ai/code/session_01837kfCN1wy15PTmp5gXKmC

---
_Generated by [Claude Code](https://claude.ai/code/session_01837kfCN1wy15PTmp5gXKmC)_

### DIFF
--- a/projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe-srcdoc.ts
+++ b/projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe-srcdoc.ts
@@ -10,12 +10,11 @@ const READER_IFRAME_CSS = readFileSync(
 export interface ReaderIframeSrcdocInput {
 	content: string;
 	/**
-	 * The deployment's own origin (e.g. https://readplace.com). When provided,
-	 * in-article links back to this host are rewritten to navigate the reader's
-	 * own tab rather than open a new one. Omitted by unit tests that only assert
-	 * the document shell.
+	 * The deployment's own origin (e.g. https://readplace.com). In-article links
+	 * back to this host are rewritten to navigate the reader's own tab rather
+	 * than open a new one.
 	 */
-	appOrigin?: string;
+	appOrigin: string;
 }
 
 /**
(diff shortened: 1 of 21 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #508
- **Commit**: f7da7ef6b882d534d9a9e4dd9113f4755a563d5b

projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe-srcdoc.ts
```diff
@@ -25,5 +33,11 @@ export interface ReaderIframeSrcdocInput {
 export function buildReaderIframeSrcdoc(
 	input: ReaderIframeSrcdocInput,
 ): string {
-	return `<!doctype html><html><head><meta charset="utf-8"><meta name="viewport" content="width=device-width, initial-scale=1"><base target="_top"><style>${READER_IFRAME_CSS}</style></head><body class="article-body__content">${input.content}</body></html>`;
+	const content = input.appOrigin
```
FagnerMartinsBrack: @claude input.appOrigin is not optional, all readers that use the iframe MUST specify it

FagnerMartinsBrack:

## Context
- **PR**: #508
- **Commit**: 2ddc1caffbf21cd7d4d6fc40ea2f4e592c367369

PR #524, opened 2026-06-10
fix(view): noindex public /view and point canonical at the original source

FagnerMartinsBrack:
## Why

`readplace.com/view/<url>` renders third-party article text. If Google indexes those wrapper pages, the domain starts looking like a scraper proxy, which drags first-party product pages down with it. HN readers also resent interstitial wrappers. The fix is to disclaim the wrapper in our SEO signals and attribute the content to the primary source — direct citations to original sources raise trust with both search and answer engines.

## What changed

Every public `/view` article page now:

- **Emits `robots: noindex, follow`** (was `index, follow`). The page stays crawlable so engines can read the canonical and follow through to the source, but the wrapper itself is kept out of the index.
- **Sets `<link rel="canonical">`, `og:url`, and JSON-LD `url` to the original publisher URL** (was the `https://readplace.com/view/…` wrapper URL). This matches what the `text/markdown` variant of `/view` already did (`Canonical: <source>`).
- **Drops the now self-referential JSON-LD `isBasedOn`** — with `url` already pointing at the source, the Article schema cleanly describes the original.

## How

`base.component.ts`'s `normalizeCanonicalUrl` deliberately forces every canonical onto `readplace.com` (guarded by an existing test, to stop staging URLs leaking into canonicals). Rather than weaken that shared guard, I added an explicit opt-in:

- `SeoMetadata.canonicalIsExternal` — when set, the base layout preserves the cross-origin canonical instead of forcing the origin. Only `/view` sets it.

## Scope / non-goals

- **Share links are unchanged.** The share balloon still produces `…/view/<url>` links against the visitor's `appOrigin` — only the indexing/attribution signals move to the source.
- The `/view` **landing form** (the paste box, no third-party text) stays `index, follow`.
- No sitemap change needed — `/view` paths were never listed in `sitemap.xml`.

## Verification

`pnpm nx run hutch:check` passes (type-check, biome, knip, unused-css, tests, coverage). `view.component.ts` is at 100% coverage; all thresholds met. Updated the affected route/component tests to assert the source URL, `noindex, follow`, and the JSON-LD `url`/absent `isBasedOn`.

https://claude.ai/code/session_01LabRr5Axuqcqz37dttUvEz

---
_Generated by [Claude Code](https://claude.ai/code/session_01LabRr5Axuqcqz37dttUvEz)_

### DIFF
--- a/projects/hutch/src/runtime/web/base.component.ts
+++ b/projects/hutch/src/runtime/web/base.component.ts
@@ -178,7 +178,9 @@ function renderBaseTemplate(body: PageBody, state: BannerState): string {
 		staticBaseUrl: STATIC_BASE_URL,
 		title: seo.title,
 		description: seo.description,
-		canonicalUrl: normalizeCanonicalUrl(seo.canonicalUrl),
+		canonicalUrl: seo.canonicalIsExternal
+			? new URL(seo.canonicalUrl).href
+			: normalizeCanonicalUrl(seo.canonicalUrl),
 		ogType,
 		ogImage: seo.ogImage,
 		ogImageAlt: seo.ogImageAlt,
--- a/projects/hutch/src/runtime/web/page-body.types.ts
+++ b/projects/hutch/src/runtime/web/page-body.types.ts
@@ -2,6 +2,11 @@ export interface SeoMetadata {
 	title: string;
 	description: string;
 	canonicalUrl: string;
+	/** When true, `canonicalUrl` is kept as a cross-origin URL instead of being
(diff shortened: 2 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #524
- **Commit**: a8bd69f84513bf9eabd8dc5f163956b3f9b0f6d5

FagnerMartinsBrack:
@claude create a plan to fix the low priority issuesand explain the reasoning

PR #527, opened 2026-06-10
fix(hutch): mark session, visitor and click cookies Secure on https

FagnerMartinsBrack:
## Why

The `Secure` attribute was missing from all three app-set cookies (`hutch_sid`, `hutch_vid`, `hutch_click`), so they could be sent in cleartext on any non-HTTPS hop (downgrade, mixed content, misrouted plain HTTP). `Secure` must be set by the application — the CDN/proxy doesn't add it. Low severity, standard defence-in-depth.

## What

- New `web/cookie-options.ts` with `baseCookieOptions(secure)` (`httpOnly: true, sameSite: "lax", path: "/", secure`) and `isHttpsOrigin(origin)`. All three cookies — session (`auth.page.ts` / `google-auth.page.ts`, incl. the Google OAuth state cookie that mirrors the session options), visitor-id middleware, and click-attribution middleware — now build their options from this single helper, so the attributes can never drift between them.
- The `secure` flag is derived in `createApp` from the already-injected `appOrigin` (`https:` scheme → `secure: true`) and injected into the auth routers and both middlewares. Deriving from the serving origin rather than `NODE_ENV` means the attribute follows the actual transport, and prod/staging (https `APP_ORIGIN`) get Secure while local dev (`http://localhost:3000`) and the e2e server (`http://127.0.0.1`) still receive cookies — login keeps working over plain http.
- No other cookie attribute changes; `HttpOnly`, `SameSite=Lax`, `Path=/`, and the existing `maxAge` values are untouched.

## Tests

- Unit: `cookie-options.test.ts` pins both helper branches with exact-shape assertions; middleware unit tests now pin `secure: false`/`true` alongside the unchanged attributes.
- Integration: new `secure-cookies.route.test.ts` boots the app with an https origin (all three cookies carry `Secure; HttpOnly; SameSite=Lax`) and with the http test origin (cookies still set, no `Secure`, login redirect works).
- `queue.share-balloon.route.test.ts` logged in via the supertest cookie jar against an `https://staging.…` origin; the jar correctly refuses to replay the now-Secure session cookie over plain http, so it authenticates with an explicit `Cookie` header instead (its subject is share-URL rendering, not cookie transport).

`pnpm check` passes across all 25 projects.

https://claude.ai/code/session_01UZZUBT6BQoxbAXREUM6FUw

---
_Generated by [Claude Code](https://claude.ai/code/session_01UZZUBT6BQoxbAXREUM6FUw)_

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -103,6 +103,7 @@ import { createClickAttributionMiddleware } from "./web/click-attribution.middle
 import { createVisitorIdMiddleware } from "./web/visitor-id.middleware";
 import { initGoogleAuthRoutes } from "./web/auth/google-auth.page";
 import { SESSION_COOKIE_NAME } from "./web/auth/session-cookie";
+import { isHttpsOrigin } from "./web/cookie-options";
 import { initForgotPasswordRoutes } from "./web/auth/forgot-password.page";
 import { initQueueRoutes } from "./web/pages/queue/queue.page";
 import { initImportSessionRoutes } from "./web/pages/import/import.page";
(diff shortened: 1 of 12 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #527
- **Commit**: d26db532a63c09f175d37bcbad74e823ce42aec4

PR #534, opened 2026-06-11
ci: rename all-tests job to web-tests

FagnerMartinsBrack:
## Why

The `all-tests` job name overstated its coverage: it runs only the Node suite (`pnpm check`), while the iOS test bundle runs in the separate `ios-tests` job. Integrating the two into one job was considered first (per the request) but isn't feasible:

- A GitHub Actions job runs on a single runner OS — `pnpm check` runs on `ubuntu-latest`, while the iOS bundle needs Xcode + an iPhone simulator on `macos-15`.
- `ios-tests` is deliberately path-gated to iOS changes on main pushes (via `detect-projects`, which doesn't run on PRs); folding it into the always-on job would either run macOS runners (~10× Linux cost) on every push/PR or change the gating semantics.

So this takes the agreed fallback: rename `all-tests` → `web-tests` and keep `ios-tests` separate.

## Changes

- Rename the `all-tests` job to `web-tests` in `ci.yml`, including all `needs:` lists and `if: needs.all-tests.result` expressions (`deploy-platform`, `pipeline`, `ios-testflight-publish`).
- Document on the job why the iOS suite lives separately.

No behaviour change — same steps, same triggers, same gating. Downstream automation (`claude-PR-*` workflows) keys off the workflow name `"CI"`, not job names, so it is unaffected.

## ⚠️ Action that may be needed outside this PR

If branch protection / rulesets list **`all-tests`** as a required status check, update the required check to **`web-tests`** — otherwise PRs (including this one) will wait forever on a check that no longer reports.

https://claude.ai/code/session_01C2yqm6cgM9UKLX3FBGuSkL

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2yqm6cgM9UKLX3FBGuSkL)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,10 @@ permissions:
   contents: read
 
 jobs:
-  all-tests:
+  # Lint/type/test/coverage suite for every Node project (web app, extensions,
+  # packages). The iOS suite cannot live in this job: it needs Xcode on a macOS
+  # runner and is path-gated to iOS changes (see ios-tests below).
+  web-tests:
     runs-on: ubuntu-latest
     environment: staging
     env:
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #534
- **Commit**: 2581b78742c172dbe7f04564bca5276c959a3d24

PR #540, opened 2026-06-12
fix(view): point og:url at the readplace wrapper, assert canonicalIsExternal is absolute

FagnerMartinsBrack:
Follow-up to the four low-priority review notes on #524.

## Summary

- **Decouples `og:url` from `<link rel="canonical">`** via a new optional `SeoMetadata.ogUrl`. `/view` now sets `ogUrl` to the readplace wrapper URL so social shares preserve readplace's Open Graph object identity — meaning Facebook/LinkedIn render the **downloaded thumbnail + reader content** rather than walking off to the source. `<link rel="canonical">` stays on the publisher for attribution. Pages that don't set `ogUrl` keep their existing behaviour (`og:url === canonical`).
- **Hardens the `canonicalIsExternal` branch** with `assert(/^https?:\/\//i.test(...))` via a new `externalCanonicalUrl()` helper. The non-external branch tolerates relative URLs (`normalizeCanonicalUrl` resolves them against `https://readplace.com`); the external branch did not, and a future caller passing a relative URL would have surfaced as a raw `ERR_INVALID_URL` 500'ing the whole page. Today's only caller (`/view`) already passes an absolute URL from `validateSaveableUrl`, so behaviour is unchanged — this is a fail-fast guard for the next caller.
- **Pins the contracts at the base layer**: positive `canonicalIsExternal` (preserves cross-origin canonical), negative (assert fires fast on relative URL), `og:url` default-to-canonical, and `og:url` decoupled-from-canonical-when-`ogUrl`-set. Sit next to the existing staging-URL canonical-rewrite guard at `base.component.test.ts:434` so the new behaviour is documented at the layer that interprets it.
- Updates two `/view` tests that asserted `og:url === canonical` to reflect the new decoupling (canonical still pegs to the publisher; `og:url` normalizes the wrapper onto the prod readplace host).

## Notes on the original review

Notes ① (noindex + cross-domain `rel=canonical`) and ②'s "decouple or leave it" are confirmed-intent — Google honours the noindex (PR's primary goal) and the cross-origin canonical is only ever a hint, so the canonical serves non-Google consumers (social unfurls, answer engines parsing the markup). Note ② is now implemented because the product call is "preserve readplace's OG object identity on shares so the downloaded thumbnail/content stays on us."

## Test plan

- [x] `pnpm nx run hutch:check` (lint + type-check + tests + coverage + knip + unused-css) — green at 100% coverage across 245 files
- [x] `assert`'s false path lives inside `node:assert`, so caller-side V8 block coverage stays clean — no new ignore comments needed
- [x] Existing `/view` SEO tests at `view.component.test.ts:135` and `:518` updated for the new `og:url` semantics
- [x] New base-level tests cover both branches of `canonicalIsExternal` and both branches of `ogUrl ?? canonical`

---
_Generated by [Claude Code](https://claude.ai/code/session_01GkoPQNjHmUtDj8JBfgrKVz)_

### DIFF
--- a/projects/hutch/src/runtime/web/base.component.test.ts
+++ b/projects/hutch/src/runtime/web/base.component.test.ts
@@ -447,6 +447,78 @@ describe("Base component", () => {
 		).toBe("https://readplace.com/queue");
 	});
 
+	it("preserves a cross-origin canonical (no readplace.com rewrite) when canonicalIsExternal is set", () => {
+		const page = createTestPageBody({
+			seo: {
+				title: "T",
+				description: "D",
+				canonicalUrl: "https://example.com/original-post",
+				canonicalIsExternal: true,
+			},
+		});
+		const result = Base(page, GUEST_STATE).to("text/html");
+		const doc = new JSDOM(result.body).window.document;
+
+		expect(
+			doc.querySelector('link[rel="canonical"]')?.getAttribute("href"),
+		).toBe("https://example.com/original-post");
+	});
+
+	it("throws a clear invariant error when canonicalIsExternal is set with a relative URL (the branch tolerates only absolute http(s) URLs)", () => {
+		const page = createTestPageBody({
+			seo: {
+				title: "T",
+				description: "D",
+				canonicalUrl: "/relative-path",
+				canonicalIsExternal: true,
+			},
+		});
+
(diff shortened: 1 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #540
- **Commit**: 38d8adb723c6959a5f022655fabbb6187e5fb28a

PR #559, opened 2026-06-14
fix(article-parser): preserve LinkedIn post paragraph structure

FagnerMartinsBrack:
## Problem

A LinkedIn post saved via the browser extension (to bypass the login wall) rendered on readplace with **all line breaks and paragraphs collapsed** into continuous text — every word present, but the structure gone.

Example (prod): `readplace.com/view/www.linkedin.com/posts/ivan-vitiaev_…activity-7471119261797756928-OoS_`

## Root cause

LinkedIn renders a post body as a single inline `<span dir="ltr">` with **no block elements** — a soft line break is one `<br>` and a paragraph break is `<br><br>`. Mozilla Readability's `_replaceBrs` only converts a *chain* of 2+ `<br>` into a `<p>`, and leaves the text preceding the first chain as an **orphan text node** that is never wrapped. The result: the leading sentence butts directly against the next paragraph and the post's paragraph rhythm is lost.

Every other stage was reproduced as lossless: the extension sends real `document.documentElement.outerHTML` (`<br>` intact), S3 store/read-back is byte-exact, the linkedom round-trips preserve `<br>`, and the reader iframe renders content raw. The loss is solely in the Readability parse.

## Fix

Add a `linkedinPreParser` (mirroring `mediumPreParser` / `theInformationPreParser`) that runs **before** Readability:

- Locates the element directly hosting the post's `<br>`s (descending through `.feed-shared-update-v2__description` / `.update-components-text` / older text classes).
- Rebuilds explicit block markup via DOM manipulation: **runs of 2+ `<br>` → `<p>` boundaries; single `<br>` → soft breaks**.
- `buildSyntheticHtml` then wraps the result in a clean `<article>`, so `_replaceBrs` never runs and no leading text is orphaned.

It's conservative: returns `undefined` (→ default Readability) for any page that is not a `<br>`-structured feed post — no known container, no `<br>`, or no extractable text — so Pulse articles and non-LinkedIn pages are untouched.

Registered at the save-link (prod crawl) and hutch composition roots, ahead of the permissive `mediumPreParser`.

## Test plan

- `pnpm nx run article-parser:check` → lint + type-check + knip + **138 tests pass**, **100% coverage** on `linkedin-pre-parser.ts`.
- `pnpm nx affected -t check` → all affected projects pass, 100% coverage maintained (run by the pre-commit hook).
- New `linkedin-pre-parser.test.ts` includes an **end-to-end `parseHtml` regression**: asserts multiple `<p>`, no `threat.<p>` orphan, and no residual `<br><br>`.

## Caveat to confirm

The live LinkedIn DOM can't be fetched (login wall), so the container selectors are based on LinkedIn's published class names plus the reproduced structure. The safe fallback degrades to today's behavior rather than breaking. Highest-confidence validation is to run the **actual stored pending HTML** for the affected URL (from the `PENDING_HTML` S3 bucket) through the parser — happy to add a small verification script if wanted.


FagnerMartinsBrack:

## Context
- **PR**: #559
- **Commit**: 5aec5980928501f7d9d7de466cd09be5bc36449b

FagnerMartinsBrack:
@claude is it possible to create a generic rule to convert `<br><br>` into a paragraph without building a linkedin pre-parser in such a way it doesnt break other sites? `<br><br>` seems to be a pretty common pattern for WYSIWYG paragraps

FagnerMartinsBrack:
@claude Can we just make the `<br><br>` work when the content includes it by adding the relevant space for the parsed clean reader display?

FagnerMartinsBrack:
@claude the issue with this PR is that there's a lot fo custom code ansd hooks uinto custom linkedin classes that can change later. Can we do a simple global brbr converted into just 2 brbr in the resulted html if readability misses them? then the brbr inside the span would just render as is?

FagnerMartinsBrack:
@claude swap this PR over to the in-place transform now

FagnerMartinsBrack:

## Context
- **PR**: #559
- **Commit**: 3f8b3e5db3b8e6f8fef23f8ac0a220f3e9025269

PR #631, opened 2026-06-20
test(hutch): cover dynamodb-article-store and drop whole-file c8 ignore

FagnerMartinsBrack:
## What

Removes the whole-file `/* c8 ignore start/stop */` from `projects/hutch/src/runtime/providers/article-store/dynamodb-article-store.ts` and replaces the coverage it was hiding with real unit tests.

## Why

- **Rule:** CLAUDE.md "Allowed `c8 ignore` Cases / Thin AWS SDK Wrappers" + the `test-driven-design` skill, which define the wrapper exemption as "truly trivial (3-5 lines, no branching), tested via production canaries".
- **Source:** `CLAUDE.md`, `.claude/skills/test-driven-design/SKILL.md`
- **File:** `projects/hutch/src/runtime/providers/article-store/dynamodb-article-store.ts`

The ignored file is ~574 lines of branching logic — `findArticlesByUser`/`countArticlesByUser` `do/while` pagination over `LastEvaluatedKey`, conditional `FilterExpression` assembly, the read/unread branch in `updateArticleStatus`, the `excludeContent` projection branch, and `ConditionalCheckFailedException` swallow-vs-rethrow in three writers. It is consumed by `app.ts` and four Lambda mains and has a co-located `.test.ts`, so it is unit-testable. The whole-file ignore was suppressing untested branches, which the coverage rules require to be exercised, not excluded — and silenced coverage on a load-bearing production path.

## Change

- Drop the two `c8 ignore` markers (file open + close).
- Generalise the co-located fake `DynamoDBDocumentClient` to a per-command-type FIFO response queue (plus a default) so pagination loops can be driven, and add cases for every store method and branch.
- Replace the one negative assertion (`.not.toContain("#content")`, which also matched `#contentSourceTier`) with a positive exact-match on the full `ProjectionExpression`, per "Avoid Negative Test Assertions".

## Verification

`pnpm nx run hutch:check` passes: tests green, biome/knip/unused-css clean, and the store file reports 100% lines (572/572), functions (22/22), statements (572/572) and branches (94/94) with no `c8 ignore`.

### DIFF
--- a/projects/hutch/src/runtime/providers/article-store/dynamodb-article-store.test.ts
+++ b/projects/hutch/src/runtime/providers/article-store/dynamodb-article-store.test.ts
@@ -1,36 +1,46 @@
 import { ConditionalCheckFailedException, type DynamoDBDocumentClient } from "@packages/hutch-storage-client";
+import { ArticleResourceUniqueId } from "@packages/article-resource-unique-id";
+import { MinutesSchema, ReaderArticleHashId } from "@packages/domain/article";
 import type { UserId } from "@packages/domain/user";
 import { initDynamoDbArticleStore } from "./dynamodb-article-store";
 
 const USER = "abc123" as UserId;
 const URL = "https://example.com/article";
+const RESOURCE_ID = "example.com/article";
+const ROUTE_ID = ReaderArticleHashId.from(URL).value;
+const TWO_MINUTES = MinutesSchema.parse(2);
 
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #631
- **Commit**: 896e4a50299875282ed8002daed2e1ea129e9e6b

PR #636, opened 2026-06-20
refactor(domain): brand oauth ids via zod instead of `as`

FagnerMartinsBrack:
## What

Removes the `as` type assertions used to brand the four OAuth identifier schemas in `src/packages/domain/src/oauth/oauth.schema.ts`:

```ts
// before
export const OAuthClientIdSchema = z.string().transform((s): OAuthClientId => s as OAuthClientId);
// after
export const OAuthClientIdSchema = z.string().brand<"OAuthClientId">();
export type OAuthClientId = z.infer<typeof OAuthClientIdSchema>;
```

The same change is applied to `AccessTokenSchema`, `RefreshTokenSchema`, and `AuthorizationCodeSchema`. The brand types are now derived from the schemas with `z.infer`. `oauth.types.ts` keeps the `OAuthClient` interface and imports the brand types from `oauth.schema.ts`; `index.ts` re-points the four type re-exports to `./oauth.schema`.

## Why

- **Rule:** Avoid TypeScript Type Assertions (`as`) — **Source:** CLAUDE.md
- `as`-branding silently hides any future mismatch between schema output and the brand type. This is not an allowed exception (`as const` / isolated Node API wrapper).
- Matches the idiom already used in the same package: `saveable-url.ts` (`z.string().brand<"SaveableUrl">()`) and `import-session.schema.ts` (`.brand<"ImportSessionId">()`).

## Impact

The `@packages/domain/oauth` public export surface is unchanged, so no consumers change. `oauth.schema.test.ts` stays green: zod `.brand()` preserves runtime value equality (`.parse("client-abc") === "client-abc"`) and still rejects non-strings.

### DIFF
--- a/src/packages/domain/src/oauth/index.ts
+++ b/src/packages/domain/src/oauth/index.ts
@@ -1,13 +1,13 @@
-export type {
-	OAuthClientId,
-	AuthorizationCode,
-	AccessToken,
-	RefreshToken,
-	OAuthClient,
-} from "./oauth.types";
+export type { OAuthClient } from "./oauth.types";
 export {
 	OAuthClientIdSchema,
 	AccessTokenSchema,
 	RefreshTokenSchema,
 	AuthorizationCodeSchema,
+} from "./oauth.schema";
+export type {
+	OAuthClientId,
+	AuthorizationCode,
+	AccessToken,
+	RefreshToken,
 } from "./oauth.schema";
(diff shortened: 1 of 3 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #636
- **Commit**: da332ec9bb9670db896197738938d299977bd3c0

PR #645, opened 2026-06-20
test(subscription): replace `{} as never` Lambda context with a typed stub

FagnerMartinsBrack:
## What

`handle-subscription-cancellation-scheduled-handler.test.ts` passed `{} as never` as the AWS Lambda `Context` argument at six handler call sites (lines 65, 95, 124, 150, 169, 197).

## Why

- **Rule:** "Avoid TypeScript Type Assertions (`as`)" — *Faking a partial object with `as`* is BAD (the documented `{ status: 200 } as Response` anti-pattern). `as never` is not one of the allowed exceptions (`as const`, isolated Node.js API wrappers).
- **Source:** `CLAUDE.md`
- **File:** `projects/hutch/src/runtime/handle-subscription-cancellation-scheduled/handle-subscription-cancellation-scheduled-handler.test.ts`

`{} as Partial<Context>` would not type-check because the handler's parameter is the full required `Context` interface. The repo already solves this without an assertion: `reader-ready-fanout-handler.test.ts` and `reader-ready-notify-handler.test.ts` declare a fully-typed `const stubContext: Context = { ... }`.

## Change

- Import `Context` alongside `SQSEvent` from `aws-lambda`.
- Add a `stubContext: Context` fixture mirroring the existing convention.
- Replace all six `{} as never` arguments with `stubContext`.

The handler ignores the context object, so all test assertions and behaviour are unchanged. No coverage impact.

### DIFF
--- a/projects/hutch/src/runtime/handle-subscription-cancellation-scheduled/handle-subscription-cancellation-scheduled-handler.test.ts
+++ b/projects/hutch/src/runtime/handle-subscription-cancellation-scheduled/handle-subscription-cancellation-scheduled-handler.test.ts
@@ -1,10 +1,25 @@
 import assert from "node:assert/strict";
-import type { SQSEvent } from "aws-lambda";
+import type { Context, SQSEvent } from "aws-lambda";
 import { UserIdSchema } from "@packages/domain/user";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #645
- **Commit**: beb57613dd9adab5f84636325d74ba2cdd883970

PR #646, opened 2026-06-20
test(handle-subscription-cancelled): drop `{} as never` Lambda context fake

FagnerMartinsBrack:
## What

Replaces the five `{} as never` Lambda `Context` fakes in `handle-subscription-cancelled-handler.test.ts` with a fully-typed `stubContext: Context` object literal.

## Why

- **Rule:** Avoid TypeScript Type Assertions (`as`) — faking a partial object with `as` instead of a properly typed value.
- **Source:** `CLAUDE.md` ("Avoid TypeScript Type Assertions (`as`)" section). Its BAD example is `{ status: 200, ok: true } as Response`; the only allowed exceptions are `as const` and isolated Node.js API wrappers, neither of which covers `{} as never`.
- **File:** `projects/hutch/src/runtime/handle-subscription-cancelled/handle-subscription-cancelled-handler.test.ts` (lines 80, 130, 149, 167, 193).

## Change

- Import `Context` alongside `SQSEvent` from `aws-lambda`.
- Add a `stubContext: Context` literal with every required field populated (mirroring the existing compliant `reader-ready-fanout-handler.test.ts` fixture) — no type assertion.
- Pass `stubContext` to all five handler calls instead of `{} as never`.

Test-only change; the handler ignores the context argument, so behaviour and assertions are unchanged.

### DIFF
--- a/projects/hutch/src/runtime/handle-subscription-cancelled/handle-subscription-cancelled-handler.test.ts
+++ b/projects/hutch/src/runtime/handle-subscription-cancelled/handle-subscription-cancelled-handler.test.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import type { SQSEvent } from "aws-lambda";
+import type { Context, SQSEvent } from "aws-lambda";
 import { UserIdSchema } from "@packages/domain/user";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #646
- **Commit**: da5f081de7dd714418a0b38a969fc46bad2d92e2

PR #647, opened 2026-06-20
test(observability): replace negative assertions in analytics dashboard test with positive query checks

FagnerMartinsBrack:
## What

Replace two negative `.not.toContain()` assertions in `projects/hutch/src/runtime/observability/analytics-dashboard.test.ts` with positive assertions of the actual expected query value.

## Why

The **Avoid Negative Test Assertions** rule (`.claude/skills/test-driven-design/SKILL.md`) flags `.not.toContain()` because it goes stale when the query format changes and asserts an absence rather than the real expected value.

- **Line 149** — `expect(q).not.toContain("logGroups(namePrefix")`: the dashboard renderer requires each subscription log group to be prefixed with its own `SOURCE '<name>'` keyword joined by ` | ` (see `sourceClause`/`logWidget` in `analytics-dashboard.ts`). The replacement asserts each subscription query **starts with** the exact `SOURCE 'g1' | SOURCE 'g2' | … | ` prefix built from the imported `SUBSCRIPTION_DASHBOARD_LOG_GROUPS`. This positively pins the dashboard `SOURCE` form and structurally excludes the CLI-only `logGroups(namePrefix: […])` form (which would occupy that same leading position), so it is strictly stronger than the absence check.
- **Line 202** — `expect(q).not.toContain("visitor_hash not in")`: `excludeVisitorHashesClause` returns `[]` when no hashes are configured, so the empty-hash query must equal the populated-hash query (from the existing `buildBody()` helper, which uses `["deadbeefcafef00d"]`) with the exact exclude clause removed. The replacement asserts per-widget exact equality (`toBe`) against the populated query stripped of the known clause — a positive assertion of the real expected value that also verifies the populated path injects the clause.

## Scope

Test-only change. No production code, exported signatures, or other callers are affected. Assertions stay green because they encode the exact strings emitted by `analytics-dashboard.ts`.

| Rule | Source | File |
|------|--------|------|
| Avoid Negative Test Assertions | `.claude/skills/test-driven-design/SKILL.md` | `projects/hutch/src/runtime/observability/analytics-dashboard.test.ts` |

### DIFF
--- a/projects/hutch/src/runtime/observability/analytics-dashboard.test.ts
+++ b/projects/hutch/src/runtime/observability/analytics-dashboard.test.ts
@@ -142,11 +142,9 @@ describe("buildAnalyticsDashboardBody — drift prevention", () => {
 
 	it("queries spanning subscription Lambda log groups emit one `SOURCE '<name>'` per group joined by `|` — `logGroups(namePrefix: [...])` is a CLI-only form the dashboard renderer rejects", () => {
 		const subscriptionQueries = widgetQueries().filter((q) => q.includes(`"${STREAMS.subscriptions}"`));
+		const expectedSourcePrefix = `${SUBSCRIPTION_DASHBOARD_LOG_GROUPS.map((name) => `SOURCE '${name}'`).join(" | ")} | `;
 		for (const q of subscriptionQueries) {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #647
- **Commit**: 093f9c01626a7c1f1a740d79a77eb69aa71998df

PR #649, opened 2026-06-20
refactor(article-summary): contain fake DynamoDB client cast in factory

FagnerMartinsBrack:
## What

`dynamodb-generated-summary.test.ts` cast its fake client from
`Partial<DynamoDBDocumentClient>` back to the full client type with
`client as typeof client & DynamoDBDocumentClient` at every one of its 11 call
sites.

## Why

**Rule:** Avoid TypeScript Type Assertions (`as`)
**Source:** CLAUDE.md
**File:** `projects/hutch/src/runtime/providers/article-summary/dynamodb-generated-summary.test.ts`

Widening a `Partial<T>` back to the full type with `as` re-introduces the very
assertion the `Partial<T>` pattern exists to avoid, and hides any future
signature mismatch between the fake and the real client. The cast was also
duplicated 11 times.

## Change

`createFakeClient` now returns `DynamoDBDocumentClient` with the single
unavoidable assertion contained in the factory, and the 11 call-site
assertions are removed. This matches the established convention already used at
`src/packages/check-stuck-articles/scripts/collect-stuck-rows.test.ts:38`. The
production signature is unchanged because `deps.client` is forwarded to
`defineDynamoTable`, which requires the full client. No assertion values
change, so tests and coverage are unaffected.

### DIFF
--- a/projects/hutch/src/runtime/providers/article-summary/dynamodb-generated-summary.test.ts
+++ b/projects/hutch/src/runtime/providers/article-summary/dynamodb-generated-summary.test.ts
@@ -1,17 +1,18 @@
 import type { DynamoDBDocumentClient } from "@packages/hutch-storage-client";
 import { initDynamoDbGeneratedSummary } from "./dynamodb-generated-summary";
 
-function createFakeClient(item: Record<string, unknown> | undefined): Partial<DynamoDBDocumentClient> {
-	return {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #649
- **Commit**: 6751e301876067612a89836bb9632e25737abc5f

PR #651, opened 2026-06-20
refactor(stripe-subscriptions): drop `as Record<string,string>` casts in header assertions

FagnerMartinsBrack:
## What

Removes three `as Record<string, string>` type assertions in `projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts` (lines 28, 116, 200) used to read captured fetch headers.

## Why

The captured `receivedInit?.headers` is typed `HeadersInit | undefined`. Casting it with `as Record<string, string>` bypasses the compiler and is forbidden by the **Avoid TypeScript Type Assertions (`as`)** rule in `CLAUDE.md` — it is not one of the allowed exceptions (`as const`, isolated Node.js API wrapper).

## Change

Replace the cast + property access with a `Headers` instance built from the captured init:

```ts
const headers = new Headers(receivedInit?.headers);
assert.equal(headers.get("Authorization"), "Bearer sk_test_abc");
```

`new Headers(...)` accepts `HeadersInit | undefined` directly, so no assertion is needed. `.get()` returns `string | null`, and the assertions remain equivalent because the implementation always sends these headers. No production code or other callers are affected.

- Source: CLAUDE.md
- File: projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts

### DIFF
--- a/projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts
+++ b/projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts
@@ -25,9 +25,9 @@ describe("initStripeSubscriptions", () => {
 
 			assert.equal(receivedUrl, "https://api.stripe.com/v1/subscriptions/sub_to_cancel");
 			assert.equal(receivedInit?.method, "DELETE");
-			const headers = receivedInit?.headers as Record<string, string>;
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #651
- **Commit**: 58be6b59c3ec9a42b57bcf3398f9cf4f33e467e5

PR #653, opened 2026-06-20
test(stripe-webhook): remove `as never` Lambda Context fake

FagnerMartinsBrack:
## What

`projects/hutch/src/runtime/stripe-webhook-receiver/stripe-webhook-receiver-handler.test.ts` faked the Lambda `Context` argument with `{} as never` in six call sites (lines 79, 100, 125, 149, 178, 204).

## Why

CLAUDE.md — **Avoid TypeScript Type Assertions (`as`)** — forbids `as` outside the documented exceptions (`as const`, isolated Node.js API wrappers, `requireEnv` generic). `as never` is not an allowed exception. The guideline prescribes faking a subset of an object's shape rather than asserting it.

`{} as Partial<Context>` would not type-check against the handler's `Context` parameter without a further assertion, so the contained, codebase-consistent fix is to use the fully-typed `stubContext: Context` literal already established in `reader-ready-fanout-handler.test.ts` and `reader-ready-notify-handler.test.ts`.

## Change

- Add `Context` to the `aws-lambda` type import.
- Add a `const stubContext: Context = { ... }` literal.
- Replace all six `{} as never` occurrences with `stubContext`.

The handler under test ignores the context/callback arguments, so behaviour and every assertion are unchanged; tests stay green and coverage is unaffected.

- Rule: Avoid TypeScript type assertions (`as`); fake partial objects with `Partial<T>`
- Source: CLAUDE.md
- File: `projects/hutch/src/runtime/stripe-webhook-receiver/stripe-webhook-receiver-handler.test.ts`

### DIFF
--- a/projects/hutch/src/runtime/stripe-webhook-receiver/stripe-webhook-receiver-handler.test.ts
+++ b/projects/hutch/src/runtime/stripe-webhook-receiver/stripe-webhook-receiver-handler.test.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import type { APIGatewayProxyEventV2 } from "aws-lambda";
+import type { APIGatewayProxyEventV2, Context } from "aws-lambda";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
 import type { StripeEventType } from "@packages/hutch-infra-components";
 import {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #653
- **Commit**: 26fb400cbb57528727756c16ba007800142643b9

PR #654, opened 2026-06-20
test(subscription-charge-failed): replace `as never` Lambda context with a typed stub

FagnerMartinsBrack:
## What

Replaces the five `{} as never` Lambda `Context` fakes in
`subscription-charge-failed-handler.test.ts` with a fully-typed
`const stubContext: Context = { ... }`.

## Why

- **Source:** CLAUDE.md
- **File:** `projects/hutch/src/runtime/subscription-charge-failed/subscription-charge-failed-handler.test.ts`

The test asserted `{} as never` for the Lambda `Context` argument at lines
49, 86, 107, 130, and 157. `as never` is a type assertion that bypasses
the compiler and is not among the rule's allowed exceptions (`as const`,
isolated Node wrappers). The guideline specifically calls out faking a
partial object with `as` rather than a properly-typed object.

## Change

Adds `import type { Context } from "aws-lambda"` and a fully-populated
`stubContext: Context`, then replaces every `{} as never` with
`stubContext`. This mirrors the established, type-safe pattern already
used in the same `hutch` project
(`reader-ready-fanout-handler.test.ts`, `reader-ready-notify-handler.test.ts`)
and across ~30 handler tests in `save-link`. The handler implementation
never reads the context argument, so all behaviour and assertions are
unchanged.

### DIFF
--- a/projects/hutch/src/runtime/subscription-charge-failed/subscription-charge-failed-handler.test.ts
+++ b/projects/hutch/src/runtime/subscription-charge-failed/subscription-charge-failed-handler.test.ts
@@ -1,4 +1,5 @@
 import assert from "node:assert/strict";
+import type { Context } from "aws-lambda";
 import { UserIdSchema } from "@packages/domain/user";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
 import type { PublishCancelSubscriptionCommand } from "@packages/test-fixtures/providers/events";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #654
- **Commit**: b1e6afdc6b91238935262f19f1f4d31746c80cef

PR #660, opened 2026-06-20
refactor(save): serve save-error countdown as a same-origin bundle

FagnerMartinsBrack:
## What

The `/save` error page's 5-second redirect countdown was emitted as a hand-written **inline `<script>` string** injected via `PageBody.scripts`:

```ts
const COUNTDOWN_SCRIPT = `<script>
(function() {
  var el = document.querySelector('.save-error__seconds');
  ...
})();
</script>`;
```

This is the inlining pattern the web skill steers away from.

## Why

`.claude/skills/web/SKILL.md` — **"Browser JS Is Bundled and Served Same-Origin"** / **"Client-Side JavaScript Conventions"**: behaviour htmx cannot express (here, the countdown animation) must live in a compiled `*.client.ts` IIFE bundle referenced via a relative same-origin `<script src="/...">`, shipped atomically with the HTML — not inlined.

## Change

Mirrors the existing, well-tested `pages/view/expiry-counter.client.ts` precedent:

- **New `save-error.client.ts`** — `initSaveErrorCountdown(deps)` with dependency-injected `setIntervalFn`/`clearIntervalFn`/`document`. It reads the start value from a `data-countdown-seconds` attribute (a genuinely-nullable DOM read) so every branch is reachable and 100% coverable — no `??` or assert-throw phantom branches.
- **New `save-error.client.test.ts`** — covers no-element, missing-attribute, non-numeric-attribute, per-tick decrement, stop-at-zero, and `stop()`.
- **Template** gains `data-countdown-seconds="{{seconds}}"` while keeping the visible `{{seconds}}` text, so `save.route.test.ts`'s `countdown.textContent === "5"` assertion stays green.
- **Component** references `<script src="/client-dist/save-error.client.js" defer></script>`.
- **`scripts/build-client-bundles.js`** registers the new bundle (globalName `SaveErrorCountdown`, DOMContentLoaded footer wiring real browser timers) so Lambda packaging + `express.static` serve it same-origin.

No test asserted the old inline script content, so nothing else ripples. `client-dist/` is gitignored and rebuilt by the build step.

## Source

- Rule: Browser JS must be a same-origin bundled `<script src>`, not inlined
- Doc: `.claude/skills/web/SKILL.md`
- File: `projects/hutch/src/runtime/web/pages/save/save-error.component.ts`

### DIFF
--- a/projects/hutch/scripts/build-client-bundles.js
+++ b/projects/hutch/scripts/build-client-bundles.js
@@ -166,6 +166,23 @@ const BUNDLES = [
 	{
 		entry: path.join(
 			PROJECT_ROOT,
+			"src/runtime/web/pages/save/save-error.client.ts",
+		),
+		outfile: path.join(OUT_DIR, "save-error.client.js"),
+		globalName: "SaveErrorCountdown",
+		footer: [
+			"document.addEventListener('DOMContentLoaded', function () {",
+			"  SaveErrorCountdown.initSaveErrorCountdown({",
+			"    document: window.document,",
+			"    setIntervalFn: function (cb, ms) { return window.setInterval(cb, ms); },",
+			"    clearIntervalFn: function (id) { window.clearInterval(id); }",
+			"  });",
+			"});",
+		].join("\n"),
+	},
+	{
+		entry: path.join(
+			PROJECT_ROOT,
 			"src/runtime/web/pages/view/view-paywall.client.ts",
 		),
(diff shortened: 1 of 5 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #660
- **Commit**: 664ec892bce5f4596fe5db6056b5be3e7deec1f5

PR #668, opened 2026-06-20
refactor(domain): brand read-time via MinutesSchema instead of `as`

FagnerMartinsBrack:
## What

`calculateReadTime` in `src/packages/domain/src/article/estimated-read-time.ts` branded both of its computed returns with `as Minutes`:

```ts
return 1 as Minutes;
return Math.max(1, Math.ceil(wordCount / WORDS_PER_MINUTE)) as Minutes;
```

This bypasses the compiler: if the `Minutes` brand changes upstream, the `as` silently hides the mismatch.

## Why

CLAUDE.md — **Avoid TypeScript Type Assertions (`as`)** — requires branded domain values to flow through a validated brand factory rather than `as`. The domain already exposes one: `MinutesSchema` (`z.number().transform((n): Minutes => ...)`) in `article.schema.ts`, which is the single, documented isolated wrapper that contains the cast.

## Change

- Import `MinutesSchema` from `./article.schema` and return `MinutesSchema.parse(...)` for both branches.
- Remove the now-unused `import type { Minutes }` (would otherwise trip knip/biome).

Runtime values are unchanged (the brand is compile-time only), so `estimated-read-time.test.ts` — which asserts numeric values via `.toBe(...)` — stays green, and the two existing coverage branches (`wordCount <= 0` vs not) are preserved.

- Source: CLAUDE.md
- File: `src/packages/domain/src/article/estimated-read-time.ts`

### DIFF
--- a/src/packages/domain/src/article/estimated-read-time.ts
+++ b/src/packages/domain/src/article/estimated-read-time.ts
@@ -1,10 +1,11 @@
+import { MinutesSchema } from "./article.schema";
 import type { Minutes } from "./article.types";
 
 const WORDS_PER_MINUTE = 238;
 
 export function calculateReadTime(wordCount: number): Minutes {
 	if (wordCount <= 0) {
-		return 1 as Minutes;
+		return MinutesSchema.parse(1);
 	}
-	return Math.max(1, Math.ceil(wordCount / WORDS_PER_MINUTE)) as Minutes;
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #668
- **Commit**: 103fdbdac3b7387bf6c3aaf611896d318acccdaa

PR #678, opened 2026-06-20
test(cancel-subscription): type fake Lambda context, remove plan-bound comment

FagnerMartinsBrack:
## What

In `projects/hutch/src/runtime/cancel-subscription/cancel-subscription-handler.test.ts`:

- Add a local `buildLambdaContext(): Context` factory (typed via `import type { Context } from "aws-lambda"`) and replace all nine `{} as never` fakes of the handler's Lambda `Context` argument with `buildLambdaContext()`.
- Remove the plan/task-bound comment `// No immediate Stripe DELETE — that's the regression we're fixing.`

## Why

- **Avoid TypeScript type assertions (`as`)** — `{} as never` is a type assertion that is not one of the documented allowed exceptions (`as const`, isolated Node.js API wrappers). CLAUDE.md prescribes a typed value over `as`; a real `Context` factory removes the assertion entirely. Source: CLAUDE.md.
- **Comments must document why, not what; no task/PR-referencing or moment-in-time comments** — "that's the regression we're fixing" assumes the current moment and references the in-flight change, which rots once the fix lands. The next line's assertion (`cancelledSubscriptionIds()` deep-equals `[]`) already encodes the expected behaviour. Source: CLAUDE.md.

## Safety

- The handler ignores its context argument at runtime, so all assertions are unchanged.
- `**/*.test.ts` is excluded from coverage collection (`enforce-coverage.config.base.js`), so the unused Context methods do not affect the 100% functions gate.
- `@types/aws-lambda` is already a hutch devDependency; empty function bodies are already used and lint-clean in this file.
- Change is contained to the single target file — no exports, callers, or other tests are affected.

### DIFF
--- a/projects/hutch/src/runtime/cancel-subscription/cancel-subscription-handler.test.ts
+++ b/projects/hutch/src/runtime/cancel-subscription/cancel-subscription-handler.test.ts
@@ -9,6 +9,7 @@ import type {
 	PublishSubscriptionCancelled,
 } from "@packages/test-fixtures/providers/events";
 import { buildSqsEvent } from "@packages/test-fixtures/sqs";
+import type { Context } from "aws-lambda";
 import { initCancelSubscriptionHandler } from "./cancel-subscription-handler";
 
 const USER_ID = UserIdSchema.parse("4".repeat(32));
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #678
- **Commit**: 505550d0eac11dd4eb17aa43d78f6b7cb12b198a

PR #681, opened 2026-06-20
refactor(export-user-data): drop `as` casts from handler test

FagnerMartinsBrack:
## What

Removes the TypeScript type assertions in
`projects/hutch/src/runtime/export-user-data/export-user-data-handler.test.ts`
that violate CLAUDE.md.

## Why

**Rule:** Avoid TypeScript Type Assertions (`as`) / Branded Types for Domain IDs
**Source:** CLAUDE.md
**File:** `projects/hutch/src/runtime/export-user-data/export-user-data-handler.test.ts`

The file used `as` casts that bypass the compiler, none of which are an allowed exception (`as const`, isolated Node wrappers, playwright config):

- `{} as never` faked the Lambda `Context` argument.
- `"user-1" as UserId` and `1 as Minutes` cast raw literals into branded domain types instead of using a validated brand factory.
- `upload.parsedBody as {...}` cast an `unknown` value to a shape instead of parsing it at the boundary.

## Change

- Add a fully-typed `stubContext: Context` constant and pass it to every `handler(...)` call, matching the established pattern in `reader-ready-fanout-handler.test.ts`.
- Replace `"..." as UserId` with `UserIdSchema.parse(...)` and `1 as Minutes` with `MinutesSchema.parse(1)` (both exported from `@packages/domain`).
- Replace each `parsedBody as {...}` with a zod `ExportBodySchema.parse(...)` defined in the test.
- Swap the now-unused `type` imports of `UserId`/`Minutes` for the corresponding schema value imports.

No production code changes; the assertions keep the exact values they checked before, so the suite stays green and coverage is unaffected.

### DIFF
--- a/projects/hutch/src/runtime/export-user-data/export-user-data-handler.test.ts
+++ b/projects/hutch/src/runtime/export-user-data/export-user-data-handler.test.ts
@@ -1,8 +1,9 @@
 import { noopLogger, HutchLogger } from "@packages/hutch-logger";
-import type { SQSBatchResponse, SQSEvent, SQSRecord, SQSRecordAttributes } from "aws-lambda";
+import type { Context, SQSBatchResponse, SQSEvent, SQSRecord, SQSRecordAttributes } from "aws-lambda";
+import { z } from "zod";
 import { initInMemoryArticleStore } from "@packages/test-fixtures/providers/article-store";
-import type { Minutes } from "@packages/domain/article";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #681
- **Commit**: 62ed7d66ffcd4dcfce9e0b2b3748e0924b2eadd7

PR #687, opened 2026-06-20
test(hutch): drop `as` casts in api.routes.test in favor of validated UserId factory

FagnerMartinsBrack:
## What

Removes two TypeScript type assertions from `projects/hutch/src/runtime/web/api/api.routes.test.ts` that violate the **Avoid TypeScript Type Assertions (`as`)** rule in `CLAUDE.md`.

## Why

- **`"test-user-123" as UserId`** casts a raw string into a branded domain ID, bypassing the brand's intended guard. Per CLAUDE.md ("Branded Types for Domain IDs" / "Avoid TypeScript Type Assertions"), branded IDs must come from a validated factory. The codebase already exports `UserIdSchema` from `@packages/domain/user` and uses `UserIdSchema.parse(...)` for this exact purpose in many sibling test files.
- **`{ ... } as Client`** asserts a full `Client` from a partial literal, hiding any future required field added upstream. The `Client` interface only requires `id` and `grants` (both present in the literal), so the assertion is unnecessary — letting the `createTestToken(): Token` return annotation structurally check the literal restores the compile-time guard.

## Change

- `TEST_USER_ID` now built with `UserIdSchema.parse("test-user-123")` (same string value, brand applied via the validated factory).
- `as Client` removed from the fake token's `client` literal.
- Imports adjusted so no import goes unused under biome/knip: `import type { Token }` and value import `import { UserIdSchema }`.

No runtime values or test assertions change; the suite stays green.

**Source:** CLAUDE.md
**File:** `projects/hutch/src/runtime/web/api/api.routes.test.ts`

### DIFF
--- a/projects/hutch/src/runtime/web/api/api.routes.test.ts
+++ b/projects/hutch/src/runtime/web/api/api.routes.test.ts
@@ -1,6 +1,6 @@
 import assert from "node:assert";
 import request from "supertest";
-import type { Token, Client } from "@node-oauth/oauth2-server";
+import type { Token } from "@node-oauth/oauth2-server";
 import { useTestServer } from "../../test-app";
 import type { TestAppHarness } from "../../test-app";
 import {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #687
- **Commit**: b07105bef733811797371fb0791a4a0b727c148a

PR #701, opened 2026-06-20
test(queue): type save-content-limit fakes as Partial&lt;T&gt; instead of as unknown

FagnerMartinsBrack:
## What

Replaces two `as unknown as` double-casts in
`projects/hutch/src/runtime/web/pages/queue/save-content-limit-handler.test.ts`
with explicit `Partial<T>` fakes.

- `fakeResponse` now declares the fake as `Partial<Response>` and returns the typed `res` from its chaining methods (`status`/`type`/`json`), so the handler's `.status(422).type(...).json(...)` chain still works.
- `fakeRequest` now asserts its literal `as Partial<Request>` (the explicit subset) before the single boundary narrow, and annotates `get`/`accepts` return types so they actually type-check against Express's overloaded signatures.

## Why

CLAUDE.md, section **"Avoid TypeScript Type Assertions (`as`)"**, forbids faking a partial object with `as` and requires `Partial<T>` to make the subset explicit. The previous `as unknown as Response` / `as unknown as Request` form erases the compiler's shape check completely, hiding any future mismatch if Express's types change. The new shape matches the pattern already sanctioned in this repo (`click-attribution.middleware.test.ts`, `require-admin.middleware.test.ts`).

- Rule: Avoid TypeScript Type Assertions (`as`) — fake partial objects via `Partial<T>`
- Source: CLAUDE.md
- File: `projects/hutch/src/runtime/web/pages/queue/save-content-limit-handler.test.ts`

## Verification

`pnpm nx run hutch:check` passes — tsc, biome, knip, unused-css, 100% coverage, and all tests green.

🤖 Part of the guidelines &amp; skills audit.

### DIFF
--- a/projects/hutch/src/runtime/web/pages/queue/save-content-limit-handler.test.ts
+++ b/projects/hutch/src/runtime/web/pages/queue/save-content-limit-handler.test.ts
@@ -8,25 +8,28 @@ function fakeResponse(): {
 	calls: FakeJsonCalls;
 } {
 	const calls: FakeJsonCalls = {};
-	const res = {
+	const fake: Partial<Response> = {};
+	const res = fake as Response;
+	Object.assign(fake, {
 		status(s: number) { calls.status = s; return res; },
 		type(t: string) { calls.type = t; return res; },
 		json(body: unknown) { calls.body = body; return res; },
-	} as unknown as Response;
+	} satisfies Partial<Response>);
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #701
- **Commit**: 9c36d5937a23628955b30d6c44279e2ea34e8966

PR #702, opened 2026-06-20
refactor(reader-iframe): drop double cast in observer test fake

FagnerMartinsBrack:
## What

Removes the `as unknown as FakeObserverCtor` double type assertion from the reader-iframe client test's `makeObserverFactory` helper.

## Why

CLAUDE.md's **"Avoid TypeScript Type Assertions (`as`)"** rule forbids `as` outside the documented exceptions (`as const`, isolated Node wrappers). The `FakeObserver as unknown as FakeObserverCtor` cast is the explicitly-discouraged pattern of forcing a fake to satisfy an interface instead of declaring the fake to satisfy it directly.

## Change

- Extract the observer instance shape into a named `FakeObserverInstance` interface.
- Declare `class FakeObserver implements FakeObserverInstance` so the compiler verifies the shape.
- Assign `Ctor: FakeObserver` with no cast.

The class is structurally compatible with `new (cb: () => void) => FakeObserverInstance`, so the cast was unnecessary. Verified with a standalone `tsc --strict --noEmit` reproduction that assigns the factory's `Ctor` to both the `ResizeObserver` and `MutationObserver` deps slots — it type-checks clean. No asserted values or signatures change, so the existing test cases remain green.

- Rule source: CLAUDE.md — "Avoid TypeScript Type Assertions (`as`)"
- File: `projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe.client.test.ts`

### DIFF
--- a/projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe.client.test.ts
+++ b/projects/hutch/src/runtime/web/shared/article-body/reader-slot/reader-iframe.client.test.ts
@@ -2,8 +2,13 @@ import assert from "node:assert/strict";
 import { JSDOM } from "jsdom";
 import { initReaderIframes, measureContentHeight } from "./reader-iframe.client";
 
+interface FakeObserverInstance {
+	observe(): void;
+	disconnect(): void;
+}
+
 interface FakeObserverCtor {
-	new (cb: () => void): { observe(): void; disconnect(): void };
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #702
- **Commit**: e3bd05796f83b07ccf4315e4dad864ecad3f367c

PR #705, opened 2026-06-20
test(web-embed): assert exact snippet output instead of negative matchers

FagnerMartinsBrack:
## What

Replaces the three negative `.not.toMatch` assertions in
`projects/web-embed/src/runtime/embed/snippet.component.test.ts` (no
`<script>` tags, no `document.cookie`/`onclick=`/`onload=`, no
`rel="nofollow"`) with a single per-variant assertion of the **exact**
rendered canonical HTML via `.toBe()`.

## Why

The **test-driven-design** skill's *Avoid Negative Test Assertions* rule
(`.claude/skills/test-driven-design/SKILL.md`) states that negative
assertions like `.not.toMatch()` / `.not.toContain()` become stale when
code is refactored and pass for the wrong reason — the fix is to assert the
actual expected value.

`renderCanonicalSnippet` is deterministic (Handlebars substitution of the
fixed canonical origins), so the full rendered output for each variant can
be pinned exactly. The literal expected strings contain no `<script>`, no
cookie/event-handler assignments, and no `rel=nofollow`, so an exact-output
assertion is strictly stronger than the three negative checks combined: the
security and SEO invariants stay enforced, and the test now fails on any
unexpected change instead of silently passing on a stale absence.

## Scope

- File: `projects/web-embed/src/runtime/embed/snippet.component.test.ts`
- The pre-existing positive `toContain` tests (save endpoint, icon URL) and
  the `all` array are left untouched.
- No production code changes; the same already-covered path
  (`renderCanonicalSnippet`) is exercised, so coverage is unaffected.

### DIFF
--- a/projects/web-embed/src/runtime/embed/snippet.component.test.ts
+++ b/projects/web-embed/src/runtime/embed/snippet.component.test.ts
@@ -37,21 +37,16 @@ describe("rendered canonical snippet invariants", () => {
 		expect(html).toContain("https://readplace.com/embed/icon.svg");
 	});
 
-	it.each(all)("snippet %s contains no <script> tags", (_label, html) => {
-		expect(html).not.toMatch(/<script/i);
-	});
+	const expectedCanonicalHtml: Record<SnippetVariant, string> = {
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #705
- **Commit**: fce208dbe6e8086e348d665319183d822c2581dc

PR #712, opened 2026-06-20
refactor(hutch-storage-client): remove `as` casts and what-only comment in batchGet

FagnerMartinsBrack:
## What

Resolve three CLAUDE.md guideline findings in `src/packages/hutch-storage-client/src/define-table.ts`.

### 1. `Keys: batchKeys as Key[]` (line 197) — Avoid TypeScript Type Assertions
The cast existed only to strip `readonly` from `batchKeys` (`readonly Key[]`) to satisfy the SDK's mutable `Keys` field. Replaced with a mutable copy `[...batchKeys]`. `Key` (`Record<string, string | number>`) is structurally assignable to the SDK's `Record<string, NativeAttributeValue>`, so no assertion is required.

### 2. `pending = unprocessed as Key[]` (line 220) — Avoid TypeScript Type Assertions
`unprocessed` is DynamoDB SDK boundary data (`result.UnprocessedKeys?.[tableName]?.Keys`, typed `Record<string, NativeAttributeValue>[]`). Per the guideline, boundary data is validated with a Zod schema rather than cast. Added a module-level `KeySchema` and parse the response; promoted `import type { z }` to a value import (`zod` is already a declared dependency). The line is exercised by the existing `batch-get-from-table.test.ts` UnprocessedKeys retry tests, so coverage is unchanged.

### 3. `/** Raw Update passthrough. */` (line 41) — Comments Document Why, Not What
A pure what-comment restating the method name (unlike the sibling `put` doc, which explains writes bypass the read-side schema). Removed per the rule's "remove freely" stance; adding a why would require human approval.

## Why
- Rule: Avoid TypeScript Type Assertions (`as`) — source: CLAUDE.md
- Rule: Comments Document Why, Not What — source: CLAUDE.md

## Risk
Low. No public signature changes; `define-table.ts` consumers and `batch-get-from-table.test.ts` assert only on key/projection shape, which is preserved.

### DIFF
--- a/src/packages/hutch-storage-client/src/define-table.ts
+++ b/src/packages/hutch-storage-client/src/define-table.ts
@@ -17,9 +17,11 @@ import {
 	ScanCommand,
 	UpdateCommand,
 } from "@aws-sdk/lib-dynamodb";
-import type { z } from "zod";
+import { z } from "zod";
 
 type Key = Record<string, string | number>;
+
+const KeySchema = z.array(z.record(z.string(), z.union([z.string(), z.number()])));
 
 /** Every command input minus `TableName` — the gateway supplies the name. */
 type WithoutTable<T> = Omit<T, "TableName">;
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #712
- **Commit**: d9c4497851781fef76f2b46d32082f0d6afe009d

PR #716, opened 2026-06-21
refactor(web-shell): remove Request type assertions from content-negotiation test

FagnerMartinsBrack:
## What

Removes two `as unknown as Request` casts from `content-negotiation.test.ts` by narrowing `wantsMarkdown`'s parameter to a minimal structural type.

## Why

`CLAUDE.md` → **Avoid TypeScript Type Assertions (`as`)** forbids using `as` to fake partial objects; the rule explicitly applies in test code. The two fakes used `} as unknown as Request;` (line 20) and `... } as unknown as Request;` (line 49), neither of which is an allowed exception (`as const` / isolated Node wrapper).

- **Rule:** Avoid TypeScript Type Assertions (`as`) — fake partials without bypassing the compiler
- **Source:** CLAUDE.md
- **File:** `src/packages/web-shell/src/content-negotiation.test.ts` (+ `src/packages/web-shell/src/content-negotiation.ts`)

## Change

- Add `export type AcceptNegotiable = { get(name: string): string | undefined; accepts(type: string): string | false }` in `content-negotiation.ts` and accept it in `wantsMarkdown` (the only members the function uses).
- Type the test fakes as `AcceptNegotiable`; the factory now returns `AcceptNegotiable`. No `as` anywhere.
- Remove the now-unused `import type { Request }` from both files.

`Pick<Request, "get" | "accepts">` was intentionally avoided because it keeps express's `get(name: "set-cookie"): string[] | undefined` overload, which the string-returning fakes cannot satisfy — that overload is the reason the original cast existed. A full express `Request` remains structurally assignable to `AcceptNegotiable`, so production callers (`send-component.ts`, `server.ts`, `view.page.ts`) are unchanged.

## Verification

- `pnpm nx run web-shell:check` — passes (tsc, biome, knip, tests, 100% coverage).
- `pnpm nx run hutch:compile` — passes (downstream callers compile unchanged).

### DIFF
--- a/src/packages/web-shell/src/content-negotiation.test.ts
+++ b/src/packages/web-shell/src/content-negotiation.test.ts
@@ -1,7 +1,6 @@
-import type { Request } from "express";
-import { wantsMarkdown } from "./content-negotiation";
+import { type AcceptNegotiable, wantsMarkdown } from "./content-negotiation";
 
-function requestWithAccept(accept: string): Request {
+function requestWithAccept(accept: string): AcceptNegotiable {
 	const types = accept.split(",").map(entry => {
 		const [type, ...params] = entry.trim().split(";");
 		const qParam = params.find(p => p.trim().startsWith("q="));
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #716
- **Commit**: b6491a00efc0215187aa9724a2ca9f5c1872168e

PR #721, opened 2026-06-21
refactor(import-session): extract mapping logic out of c8-ignored wrapper and adopt named params

FagnerMartinsBrack:
## What

- Extract the branching selection/mapping logic from `dynamodb-import-session.ts` into a new, non-c8-ignored `import-session-mapping.ts`:
  - `computeDeselected({ allSelected, deselected, totalUrls })` — the allSelected fallback, `Set` construction, and full-deselect index math.
  - `toImportSession(row)` — the stored-row to domain-`ImportSession` mapping.
- Add `import-session-mapping.test.ts` covering every branch (allSelected true/false, deselected present/absent), so the extracted logic meets the project coverage thresholds directly instead of hiding behind a blanket ignore.
- Change `loadOwned(id: string, userId: string)` to `loadOwned(owner: { id: string; userId: string })` and update all five call sites.

## Why

- **Named Parameters Over Positional When Types Repeat** (CLAUDE.md): the two consecutive `string` params (`id`, `userId`) are transposable and easy to swap by accident.
- **Allowed c8 ignore Cases — thin AWS SDK wrapper only** (CLAUDE.md / test-driven-design SKILL.md): the whole-file ignore is justified only for a truly trivial wrapper. SKILL.md directs extracting mapping/parsing logic into a helper and unit-testing it rather than ignore-wrapping branching code. After this change the wrapper is just guards plus single SDK calls, consistent with `dynamodb-article-store.ts`, `dynamodb-auth.ts`, and `dynamodb-pending-signup.ts`.

The factory's public signature and return type are unchanged, so the `app.ts` composition root is unaffected.

### DIFF
--- a/projects/hutch/src/runtime/providers/import-session/dynamodb-import-session.ts
+++ b/projects/hutch/src/runtime/providers/import-session/dynamodb-import-session.ts
@@ -9,12 +9,10 @@ import { z } from "zod";
 import { UserIdSchema } from "@packages/domain/user";
 import {
 	IMPORT_SESSION_TTL_SECONDS,
+	type ImportSessionStore,
 	ImportSessionIdSchema,
 } from "@packages/domain/import-session";
-import type {
-	ImportSession,
-	ImportSessionStore,
-} from "@packages/domain/import-session";
+import { computeDeselected, toImportSession } from "./import-session-mapping";
 
 const SessionRow = z.object({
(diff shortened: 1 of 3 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #721
- **Commit**: 726a65f6075f7a56cbd5b33410a6afb506dafe6d

PR #723, opened 2026-06-21
test(reader-ready-fanout): replace `as UserId` casts with UserIdSchema.parse

FagnerMartinsBrack:
## What

The `reader-ready-fanout-handler.test.ts` test constructed savers with `userId: "viewer" as UserId` (and `"never"`), casting raw strings to the branded `UserId` type via `as`.

## Why

CLAUDE.md's **Avoid TypeScript Type Assertions (`as`)** rule forbids using `as` to cast a raw value to a branded domain ID and prescribes the validated-factory pattern (`UserIdSchema.parse(rawValue)`). The cast bypasses the compiler so a wrong-shaped value would be silently accepted, and the rule lists no test-file exception for branded-ID casts.

## How

- Switched the import from `import type { UserId }` to the runtime `import { UserIdSchema }` from `@packages/domain/user`.
- Replaced all three `"…" as UserId` occurrences with `UserIdSchema.parse("…")`, matching the established pattern used across the hutch suite (e.g. `conversions.test.ts`, `banner-state.test.ts`).

The `UserIdSchema` transform is an identity at runtime, so the existing positive assertions (`toHaveBeenCalledWith({ userId: "viewer", … })`) remain exactly correct. Verified by compiling and running the suite: 4/4 tests pass, no tsc errors for this file.

### DIFF
--- a/projects/hutch/src/runtime/reader-ready-fanout/reader-ready-fanout-handler.test.ts
+++ b/projects/hutch/src/runtime/reader-ready-fanout/reader-ready-fanout-handler.test.ts
@@ -1,6 +1,6 @@
 import type { Context, SQSEvent } from "aws-lambda";
 import { noopLogger } from "@packages/hutch-logger";
-import type { UserId } from "@packages/domain/user";
+import { UserIdSchema } from "@packages/domain/user";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #723
- **Commit**: 0d8c9b5ab26ebd560cd53c356dfff37fca77f421

PR #732, opened 2026-06-21
refactor(crawl-article): drop `as` cast in persona-fallback test headers

FagnerMartinsBrack:
## What

The `persona-fallback.test.ts` captured request headers by spreading
`init?.headers as Record<string, string>` at five call sites (lines 60, 76, 93,
146, 161). The fetch `init.headers` union is `HeadersInit`, so this `as` cast
bypasses the compiler.

## Why

CLAUDE.md "Avoid TypeScript Type Assertions (`as`)" forbids `as` casts, and this
one is not an allowed exception. The production code normalises the same union
via `new Headers(...)` in `callerHeaderOverrides` rather than asserting its type.

## How

Add a `recordHeaders(init)` helper that builds a plain record by iterating
`new Headers(init?.headers)`, mirroring the production approach, and replace all
five casts with calls to it. `new Headers` lowercases header names, which matches
the lowercase keys the existing assertions read (`user-agent`, `if-none-match`,
`accept`), so behaviour and expected values are unchanged.

Verified with `pnpm nx run @packages/crawl-article:check` (compile, tsc, biome,
knip all pass).

### DIFF
--- a/src/packages/crawl-article/src/persona-fallback.test.ts
+++ b/src/packages/crawl-article/src/persona-fallback.test.ts
@@ -5,6 +5,14 @@ import {
 	withPersonaFallback,
 } from "./persona-fallback";
 
+function recordHeaders(init: Parameters<typeof fetch>[1]): Record<string, string> {
+	const out: Record<string, string> = {};
+	new Headers(init?.headers).forEach((value, key) => {
+		out[key] = value;
+	});
+	return out;
+}
+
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #732
- **Commit**: a1cb272b6a7d686119bc5239442ac5783698d55b

PR #739, opened 2026-06-21
refactor(firefox-extension): replace boundary `as` casts with type guards

FagnerMartinsBrack:
## What

Removes the three `as` type assertions in the Firefox extension
background worker (`background.browser.ts`) and replaces them with runtime
type guards that validate external data at the boundary.

## Why

`browser.storage.local.get`, `runtime.onMessage`, and the content-script
`sendMessage` response all return untyped/`unknown` data. Casting that
data to `OAuthTokens`, `PopupMessage`, and `{ rawHtml: unknown }` with
`as` bypasses the compiler and silently accepts malformed shapes, which
the CLAUDE.md "Avoid TypeScript Type Assertions" rule forbids. The rule
normally directs Zod validation at boundaries, but browser extensions do
not depend on Zod (and no install runs in CI), so the browser-safe
equivalent is a hand-rolled `function isX(v: unknown): v is X` guard.

## Changes

- Add `isOAuthTokens`; `getTokens` returns `null` on a malformed stored
  blob instead of casting.
- Add `hasStringType` and `isPopupMessage`; the `onMessage` listener
  validates the discriminant before narrowing and ignores unknown
  messages.
- Drop the redundant `as { rawHtml: unknown }` cast, relying on the
  existing `in` narrowing.

## Verification

`pnpm nx run firefox-extension:check` passes (tsc, biome, knip, 100%
coverage). `background.browser.ts` is excluded from coverage, so no test
changes are required.

### DIFF
--- a/projects/extensions/firefox-extension/src/runtime/background/background.browser.ts
+++ b/projects/extensions/firefox-extension/src/runtime/background/background.browser.ts
@@ -26,12 +26,23 @@ declare const __SERVER_URL__: string;
 const SERVER_URL = __SERVER_URL__;
 const CLIENT_ID = "hutch-firefox-extension";
 
+function isOAuthTokens(value: unknown): value is OAuthTokens {
+	return (
+		typeof value === "object" &&
+		value !== null &&
+		"accessToken" in value &&
+		typeof value.accessToken === "string" &&
+		"refreshToken" in value &&
+		typeof value.refreshToken === "string"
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #739
- **Commit**: ecb44d7050da343f4f6e4568a0cf5575b9e5e369

PR #741, opened 2026-06-21
refactor(chrome-extension): remove `as` casts from popup boundary and navigator probe

FagnerMartinsBrack:
## What

Removes the three guideline violations in
`projects/extensions/chrome-extension/src/runtime/popup/popup.browser.ts`:

1. **Message-boundary `as` cast** (`raw as PopupMessage`). The
   `browser.runtime.onMessage` listener receives `raw: unknown` from the
   runtime messaging boundary and only cares about the `save-progress` message.
   Replaced the cast with two type guards, `isSavePhase` and
   `isSaveProgressMessage`, that narrow `unknown` to
   `{ type: "save-progress"; phase: SavePhase }` with no assertion. The
   extension does not depend on zod (and CI runs no install), so a hand-rolled
   guard is the correct browser-safe validator — the same pattern already used
   by `isPopupMessage` in the firefox background script.

2. **`navigator as Navigator & { userAgentData?: ... }` cast** plus the
   discretionary comment that justified it (and made a time-bound "DOM types
   don't include it yet" claim). Replaced with a top-level
   `declare global { interface Navigator { userAgentData?: { platform: string } } }`
   augmentation so the field is typed directly, then read `navigator.userAgentData`
   with no cast and no comment.

## Why

CLAUDE.md "Avoid TypeScript Type Assertions (`as`)" bans casts outside the
allowed exceptions, prescribes a validator at system boundaries, and
"Comments Document Why, Not What" forbids comments that narrate a workaround
instead of encoding the why in code.

## Verification

`pnpm nx run chrome-extension:check` passes — tsc, biome, knip, and 100%
coverage (the file is a `*.browser.ts`, excluded from coverage, so no new
tests are required).

### DIFF
--- a/projects/extensions/chrome-extension/src/runtime/popup/popup.browser.ts
+++ b/projects/extensions/chrome-extension/src/runtime/popup/popup.browser.ts
@@ -3,6 +3,7 @@ import browser from "webextension-polyfill";
 import type {
 	ReadingListItem,
 	PopupMessage,
+	SavePhase,
 	GuardedResult,
 	SaveUrlResult,
 	RemoveUrlResult,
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #741
- **Commit**: 6c833523a4edd9e78e2f7c6479b977528a2a4042

PR #742, opened 2026-06-21
refactor(extensions): validate offscreen messages with type guards instead of `as`

FagnerMartinsBrack:
## What

Replace two untrusted-boundary `as` casts in `tinted-icon.browser.ts` with hand-rolled type guards.

- `getSavedIconData` cast the `browser.runtime.sendMessage` response to `Record<number, SerializedImageData>`.
- `getDefaultPathsForCurrentTheme` cast the response to `Theme`.

The result of `browser.runtime.sendMessage` is `unknown` — it crosses the extension message boundary from the offscreen document. Casting with `as` (instead of validating the shape) hides any malformed payload, which CLAUDE.md's "Avoid TypeScript Type Assertions" rule forbids at a boundary.

## How

Browser extension files must not depend on zod, so validation is done with plain type-guard functions:

- `isSerializedImageData(value): value is SerializedImageData`
- `isSerializedIconData(value): value is Record<number, SerializedImageData>`
- `isTheme(value): value is Theme`

Each call site now narrows the `unknown` response through a guard and throws a descriptive error on malformed input instead of asserting the type.

## Testing

`*.browser.ts` is excluded from coverage, so no new test is required. Verified `nx run chrome-extension:compile` and `nx run chrome-extension:lint` (tsc + biome + knip + unused-css) pass with the change isolated.

### DIFF
--- a/projects/extensions/chrome-extension/src/runtime/background/tinted-icon.browser.ts
+++ b/projects/extensions/chrome-extension/src/runtime/background/tinted-icon.browser.ts
@@ -9,6 +9,28 @@ interface SerializedImageData {
 
 type Theme = "dark" | "light";
 
+function isSerializedImageData(value: unknown): value is SerializedImageData {
+	if (typeof value !== "object" || value === null) return false;
+	const candidate: Record<string, unknown> = { ...value };
+	return (
+		typeof candidate.width === "number" &&
+		typeof candidate.height === "number" &&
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #742
- **Commit**: 138768f13fa5a8363bf07e09302cf0f3cd81c749

PR #743, opened 2026-06-21
refactor(hutch): serve home hero scripts as a same-origin bundle

FagnerMartinsBrack:
## What

Move the home page's two inline hero `<script>` IIFEs — the headline
word-rotator and the "keep scrolling" smooth-scroll — into a compiled,
same-origin browser bundle.

- New `home.client.ts` exports `initHeadlineRotator` and `initScrollHint`, both
  fully dependency-injected (document, timers, `requestAnimationFrame`,
  `matchMedia`, scroll, computed style) so they are unit-testable in Node.
- New `home.client.js` bundle registered in
  `projects/hutch/scripts/build-client-bundles.js`; its esbuild footer wires the
  real browser globals, exactly like the other client bundles.
- `home.component.ts` now emits `<script src="/client-dist/home.client.js"
  defer>` instead of two inlined IIFE strings.
- New `home.client.test.ts` covers both behaviours (no-op paths, reduced-motion,
  rotation tick + reschedule, visibility pause/resume, smooth-scroll stepping,
  and the absent-href / missing-anchor early returns) at 100% coverage.
- `home.route.test.ts` now asserts the same-origin bundle `<script src>` instead
  of incidental rotator strings.

## Why

The web skill's "Browser JS Is Bundled and Served Same-Origin" rule reserves
`*.client.ts` bundles for animation / non-htmx behaviour and forbids inlining
large hand-written IIFE strings into the page body. The headline rotator and
smooth-scroll are precisely that case.

## Notes

- No `node:assert` in the browser file (not resolvable in the esbuild browser
  target); invariants the inline code left implicit are now testable
  early-returns.
- No `as` / `!` casts: the rotator is narrowed with `querySelector<HTMLElement>`
  plus a closure-local const, mirroring `expiry-counter.client.ts`.
- This project does not exclude `*.client.ts` from coverage, so a sibling
  `home.client.test.ts` was required and reaches 100%.


FagnerMartinsBrack:

## Context
- **PR**: #743
- **Commit**: 48b9e834cc94278da83e417944f321e17b074adb

projects/hutch/src/runtime/web/pages/home/home.client.ts
```diff
@@ -0,0 +1,153 @@
...
+	"essays",
+	"longreads",
+];
+
+const ROTATE_INTERVAL_MS = 2500;
+const FADE_IN_DELAY_MS = 150;
+const SWAP_SETTLE_MS = 700;
+const SCROLL_DURATION_MS = 350;
+
+interface HeadlineRotatorDeps {
```
FagnerMartinsBrack, line 35: @claude Using interface in a client script? Is this ok? Does the client scripts go through TS? Report in detail how that works

PR #748, opened 2026-06-21
refactor(hutch): remove defaultValue escape hatch from requireEnv

FagnerMartinsBrack:
## Audit finding (high) — Never default missing environment variables

**Rule:** Never default missing environment variables (no `requireEnv` defaultValue)
**Source:** CLAUDE.md
**File:** `projects/hutch/src/runtime/domain/require-env.ts`

`requireEnv(name, { defaultValue })` let a missing variable silently fall back to a default — the exact pattern the rule forbids ("let the process fail if a variable is not set").

### Change
- Removed the `defaultValue` option; `requireEnv(name)` now always asserts the variable is set.
- **Callers** that relied on a default now require the value:
  - `app.ts`: `RATE_LIMIT_*` (×4) and `APP_ORIGIN` — already provided by the Lambda `environment` block in `infra/index.ts` for production.
  - `server.ts`: `PORT`. Because `server.ts` reads `PORT` at module load and is imported by the web Lambda (via `createHutchApp`), **`PORT` is now also added to the Lambda `environment`** so the cold-start read succeeds (serverless-http binds no port, but the read must not throw).
- `.envrc` provides all six for local dev and the test run (the pre-commit hook sources `.envrc`), so route/unit tests that construct the app keep passing.
- `require-env.test.ts` drops the two tests that exercised the removed option.

### Production safety
Only the web Lambda (`lambda.main.ts → app.ts → server.ts`) loads these reads; the other handlers don't. `RATE_LIMIT_*`/`APP_ORIGIN` were already on its env; `PORT` is added here.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -44,6 +44,16 @@ export TRIAL_SCHEDULER_ROLE_ARN='arn:aws:iam::000000000000:role/hutch-trial-sche
 export ANALYTICS_SALT='local-dev-salt'
 export EXPIRY_COUNTDOWN='enabled'
 export FOUNDING_MEMBER_LIMIT='50'
+# Local-dev server port + app origin. In production these are required and set
+# by the Lambda environment / deploy; requireEnv no longer defaults them.
+export PORT='3000'
+export APP_ORIGIN='http://localhost:3000'
+# Rate-limit rules (requests/seconds). Production sets these on the Lambda; the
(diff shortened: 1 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #748
- **Commit**: 62be68ce6b689802026c45dc310c2b1be21161d1

PR #764, opened 2026-06-21
test(email-verification): assert TTL is a future unix timestamp

FagnerMartinsBrack:
## What

Strengthen the `createVerificationToken` test in `dynamodb-email-verification.test.ts` from an existence-only check on `expiresAt` to a value assertion that the TTL is a **future** unix timestamp.

## Why

`expect(put?.input.Item).toHaveProperty("expiresAt")` only confirmed the field is present — it would still pass if the provider wrote a past or zero TTL, which would silently expire every verification link immediately. Asserting `expiresAt > now` actually exercises the TTL contract.

## Notes

The captured command `input` is typed `Record<string, unknown>`, so the suggested `put?.input.Item?.expiresAt` doesn't type-check. The boundary value is validated through a small `z.object({ expiresAt: z.number() })` parse to obtain a typed number, matching the repo's "no `as`" / zod-at-boundaries conventions.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TAED8yfCx5zuFXFxPe635t)_

### DIFF
--- a/projects/hutch/src/runtime/providers/email-verification/dynamodb-email-verification.test.ts
+++ b/projects/hutch/src/runtime/providers/email-verification/dynamodb-email-verification.test.ts
@@ -1,3 +1,4 @@
+import { z } from "zod";
 import { ConditionalCheckFailedException, type DynamoDBDocumentClient } from "@packages/hutch-storage-client";
 import type { UserId } from "@packages/domain/user";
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #764
- **Commit**: 6e6080df435196431f08b6e8e8a1a84af5736868

PR #766, opened 2026-06-21
fix(changelog-banner): keep readers on the page when dismissing

FagnerMartinsBrack:
## Problem

Closing the site-wide changelog banner (the sticky "NEW …" widget at the top) sent the reader to the **homepage** instead of leaving them where they were — most visibly while reading the very blog post the banner announced.

## Root cause

The banner's close control is a no-JS POST form to `/banner/changelog/dismiss`. The handler redirected the reader back using the `Referer` header (`safeBackPath`). But both deployables run `helmet()`, whose **default `Referrer-Policy: no-referrer` strips `Referer`** from the request — so the handler always saw `undefined` and fell back to `/`.

The existing route tests passed only because supertest sets `Referer` manually, which masked the real-browser behaviour.

## Fix

Carry the origin page **explicitly** rather than depending on a header the app's own security policy removes (and which is unreliable in general):

- The changelog banner shell renders the current page path into a hidden `returnTo` field, threaded through a new `BannerState.currentPath` (sourced from each site's request `originalUrl`).
- The dismiss route now reads `returnTo` and validates it as a safe **same-origin relative path** before redirecting. `safeBackPath` is replaced by `safeReturnPath`, which resolves the value against a throwaway origin and rejects absolute URLs, `javascript:`, protocol-relative (`//host`), backslash (`/\host`), and control-character authorities a browser would resolve off-site.

This is self-contained to the banner feature — no global security-header or analytics changes. On the blog (no htmx) the field always carries the exact current page, so dismissing keeps the reader on the post.

## Tests

- `safeReturnPath` unit tests cover every branch (non-string, off-origin, `//`-normalised path, unparseable, happy path).
- Dismiss route POST tests now drive the `returnTo` form field instead of `Referer`.
- `bannerStateFromRequest` / `initBuildBannerState` tests assert `currentPath` is populated from `originalUrl`.
- `renderChangelogBannerShell` tests assert the hidden field renders and is HTML-escaped.
- The blog post-page test asserts the dismiss form posts `returnTo=/blog/<slug>` — the direct regression test for the reported bug.

`pnpm check` passes across the affected projects (`web-shell`, `hutch`, `blog-site`) with coverage thresholds met; the changed files are at 100%.

https://claude.ai/code/session_01KMYLcKPVtCqGt9o9d4CxNt

---
_Generated by [Claude Code](https://claude.ai/code/session_01KMYLcKPVtCqGt9o9d4CxNt)_

### DIFF
--- a/projects/blog-site/src/runtime/web/pages/blog/blog.page.ts
+++ b/projects/blog-site/src/runtime/web/pages/blog/blog.page.ts
@@ -96,7 +96,11 @@ export function initBlogRoutes(deps: { blogPosts: BlogPosts; base: RenderBase })
 	router.get("/", (req: Request, res: Response) => {
 		const posts = blogPosts.getAllPosts();
 		const changelogBanner = hideIfDismissed(blogPosts.getLatestChangelogBanner(), req);
-		sendComponent(req, res, base(BlogIndexPage({ posts }), { ...GUEST_STATE, changelogBanner }));
+		sendComponent(
+			req,
+			res,
+			base(BlogIndexPage({ posts }), { ...GUEST_STATE, changelogBanner, currentPath: req.originalUrl }),
+		);
 	});
 
 	router.get("/:slug", (req: Request<{ slug: string }>, res: Response) => {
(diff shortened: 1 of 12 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #766
- **Commit**: f5bdb58286929efbdfb29ca1c158561166ec7059

PR #768, opened 2026-06-22
feat(stripe): tag subscription metadata with userId

FagnerMartinsBrack:
## What

Attach the Readplace `userId` to the Stripe subscription's `metadata` when we create a subscription on an existing customer.

```
metadata[userId]=<userId>
```

## Why

So a subscription in Stripe can be traced back to its Readplace account even when the customer paid under a **different email** than the one on file. Today nothing on the Stripe object carries our internal id, so a mismatched email leaves the subscription un-identifiable from the Stripe side.

## Where

`createSubscriptionOnExistingCustomer` is the call that creates the subscription producing **the first charge** (Stripe invoices and charges the saved card immediately). It runs in two places, both of which already have the `userId` in hand:

- **Trial-end conversion** — `subscription-start-request-handler` (the first charge when a trial converts to paid)
- **One-click resubscribe** — the `cancelled` branch of `POST /account/subscribe`

`userId` is now a required field on the `CreateSubscriptionOnExistingCustomer` contract, so every subscription created this way is tagged.

## Notes / scope

- The Stripe **Checkout** path (`createCheckoutSession`, used for existing-user-subscribe and the test-only email/Google signup) is **not** covered here — that subscription is created inside Stripe's hosted flow and would need `subscription_data[metadata][userId]`. Happy to follow up if you want every checkout-created subscription tagged too; the trial conversion was the "first charge" target.
- Test fixtures (`in-memory-stripe-subscriptions`) now record `userId`, and the real provider test asserts `metadata%5BuserId%5D=…` is sent to Stripe.

## Verification

`pnpm check` passes (typecheck, lint, tests, coverage, knip) across all projects; the pre-commit hook ran the full `nx run-many --target=check --all` green.

https://claude.ai/code/session_019jJ86aaqxo7kPVHBE5pNd5

---
_Generated by [Claude Code](https://claude.ai/code/session_019jJ86aaqxo7kPVHBE5pNd5)_

### DIFF
--- a/projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts
+++ b/projects/hutch/src/runtime/providers/stripe-subscriptions/stripe-subscriptions.test.ts
@@ -1,5 +1,8 @@
 import assert from "node:assert/strict";
+import { UserIdSchema } from "@packages/domain/user";
 import { initStripeSubscriptions } from "./stripe-subscriptions";
+
+const USER_ID = UserIdSchema.parse("usr_test_abc123");
 
 function jsonResponse(status: number, body: object): Response {
 	return new Response(JSON.stringify(body), {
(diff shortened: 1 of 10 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #768
- **Commit**: dd1c9b2888129dc1f8fd270f27425aaefdbfa386

PR #765, opened 2026-06-21
feat(mcp): add article read tools, keep status/delete app-only

FagnerMartinsBrack:
## What

Expands the Readplace MCP server to expose existing read functionality, while deliberately keeping the two destructive actions — marking read/unread and deleting — **out of the assistant** for safety.

### New read tools (delegate to the same store seams `/queue` uses)
- **`get_article`** — full metadata for one saved article by id.
- **`get_article_content`** — cleaned reader HTML on demand, with a "still being fetched" status instead of an error while the crawl runs.
- **`get_article_summary`** — the existing AI TL;DR, with `pending` / `ready` / `failed` / `skipped` status.
- **`list_queue` (enhanced)** — opaque-cursor pagination (`limit` + `nextCursor`), `sort` (saved/read date) + `order`, and a structured result that now carries each article's **`id`** so the by-id tools can target it. No-args calls stay **byte-for-byte backward compatible**.

Every lookup is owner-scoped (`findArticleById` is keyed by `userId`), so a cross-user or malformed id resolves to "not found" rather than another user's article — verified by an integration test.

### Status change & deletion stay app-only (the safety ask)
`set_article_status` and `delete_article` are **advertised but never mutate**. Calling either returns a clear message pointing the user to the Readplace app (`/queue`), plus `structuredContent.performed: false`. They're annotated `readOnlyHint: true` since nothing is written. This way, when a user asks the assistant to mark-read or delete, it gives a clear "do it in the app" answer instead of an empty "I can't" — without ever letting a hallucination commit a destructive write. An end-to-end test calls both and then proves the queue is unchanged.

### Cross-cutting polish
- Accurate tool **annotations** on every tool (incl. `save_link`/`list_queue`).
- **`structuredContent`** + serialized-text fallback on the data tools.
- Updated `/mcp` connect page, server card, and `initialize` instructions to describe the toolset and the app-only policy.

## Design notes
- `save_link` is unchanged (still gated by the save-access seam); the new write tools mirror the app, which does **not** gate `/:id/status` or `/:id/delete` — but here they don't write at all.
- The read logic lives in a unit-tested `article-operations.ts` module, keeping the composition root thin.
- Cursor pagination is satisfied at the MCP layer by wrapping the store's existing offset pagination in an opaque token — no schema/table change.

## Testing
[...]

PR #772, opened 2026-06-22
feat(auth): brand the request principal as AuthenticatedUserId (lock MCP to it)

FagnerMartinsBrack:
## What

Hardens the user-scoping guarantee behind the Readplace MCP server (added in #765) so that **using a caller-supplied value as the request principal becomes a compile error and a CI failure**, not something that depends on every call site being wired correctly by hand.

This is a follow-up to a security + type-safety review of #765. The review's verdict was: **isolation already holds at runtime** — no MCP tool accepts a `userId`, the principal is bound server-side from the OAuth bearer token, and the DynamoDB store scopes every read by a `{userId, url}` key. But the *type system* did not enforce any of it: a single `UserId` brand was shared between "the logged-in principal" and "a userId read off a row / an event payload," so a future refactor could silently pass the wrong one. This PR closes that gap and documents two genuine (non-security) issues the review surfaced.

> Follow-up to #765 (now merged). Targets `main`; contains only the hardening commit.

## The keystone: a mint-only `AuthenticatedUserId`

A second, narrower brand that is **assignable to `UserId` but not from it**:

```ts
// domain/user/user.schema.ts
const AuthenticatedUserIdSchema = UserIdSchema.brand<"AuthenticatedUserId">();
export type AuthenticatedUserId = z.infer<typeof AuthenticatedUserIdSchema>;
export function authenticatedUserIdFrom(userId: string): AuthenticatedUserId {
	return AuthenticatedUserIdSchema.parse(userId);
}
```

I compile-tested three encodings against the repo's exact zod (4.4.3) / TS (6.0.3) under `--strict`. A second *independent* `z.brand` is **not** assignable to `UserId`, so it would force an `as` cast at every store call (which `CLAUDE.md` forbids). The **chained** brand above is the only form that both widens silently at the store boundary *and* mints with zero casts:

| Encoding | Widens to `UserId` (no `as`) | Rejects a plain `UserId` as principal |
|---|---|---|
| second independent `z.brand` | ❌ (needs `as` at ~every store call) | ✅ |
| **chained `.brand().brand()`** (this PR) | ✅ | ✅ |

- **Minted only at the two request-auth boundaries:** `ValidateAccessToken` (OAuth bearer) and `GetSessionUserId` (session cookie) now return `AuthenticatedUserId`.
- **Required by the interactive entry points:** `McpRequestContext`, the five `McpServerDeps` ops, and `resolveOwned` now take `AuthenticatedUserId`. Passing a plain `UserId` (a `saver.userId`, an event `detail.userId`, a tool argument) is a compile error.
- **Strings/`UserId` kept at the DynamoDB layer** (the reviewer's carve-out): `FindArticleById` / `FindArticlesByUser` etc. are unchanged — `AuthenticatedUserId` widens into them automatically.
- **On-behalf-of paths intentionally NOT migrated:** reader-ready fan-out (`saver.userId`), and the async event handlers (`export-user-data`, trial-feedback, reader-ready-notify) keep plain `UserId` — they act on a subject who is not the live caller. Branding them would assert a false security property.

The blast radius is small precisely because the brand *widens*: only the mint sites and the MCP entry points changed; every page handler and store call that merely passes the principal through compiles untouched.

## Lock-in (defense in depth)

1. **Compile-time (the brand).** Caller/public input is a `string`/`UserId`, never an `AuthenticatedUserId` — so wiring it into a user-scoped read won't compile.
2. **Mint encapsulation (`authenticated-principal-mint-sites.test.ts`).** The only way to *forge* a principal is to call `authenticatedUserIdFrom` on attacker input. A new architectural test scans non-test source and fails CI if that identifier appears anywhere outside the allowlisted auth-boundary files.
3. **Tool-surface invariant (`tool-definitions.test.ts`).** A new test asserts no MCP tool declares a `userId`-like field and every tool keeps `additionalProperties: false`, and that each Zod validator strips an injected `userId`. A future tool that added a principal parameter fails CI immediately.

## Other review findings fixed

- **Summary `undefined → pending` is now documented, not accidental.** `findGeneratedSummary` returns `undefined` for legacy rows the provider says to "re-prime rather than render a stuck pending row." The reader UI itself maps `undefined → "pending"` (`article-reader.ts: summary?.status ?? "pending"`), so the MCP tool mirroring that is *consistent with the app*, not divergent. Added a `why` comment + test title so it stays an intentional mirror and isn't "fixed" into an MCP↔web inconsistency.
- **`cursor.ts` no longer overclaims.** The doc comment and the user-facing error claimed "stale/expired" and tamper protection that an unsigned, TTL-less, shape-only cursor doesn't provide. Trimmed to the truth: decode returns `null` for garbage/truncated/wrong-shape tokens; pagination safety comes from owner-scoping via `context.userId`, not the cursor. Error message: "invalid or expired" → "invalid".

## Tests added

- Cross-user **read** isolation now also covers `get_article_content` and `get_article_summary` (was `get_article` + `list_queue` only).
- Cross-user **write** no-op: user B's `delete_article` / `set_article_status` on A's id is `performed:false` and leaves A's queue unchanged.
- The "unchanged" assertion now compares the **whole article** before/after the write tools, not just status + count.
- `tool-definitions` and `authenticated-principal-mint-sites` guard tests (above).

## Testing

`pnpm check` (lint, types, knip, unused-CSS, unit/integration, Playwright E2E, 100% coverage).

---


FagnerMartinsBrack:

## Context
- **PR**: #772
- **Commit**: ba2847abc20d1d250a70d4dff6233835f486f5a3

projects/hutch/src/runtime/authenticated-principal-mint-sites.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const ALLOWED_MINT_SITES = [
+	"src/packages/domain/src/user/user.schema.ts", // defines the mint
+	"src/packages/domain/src/user/index.ts", // re-exports it
+	"src/packages/test-fixtures/src/providers/oauth/validate-access-token.ts", // bearer-token boundary
+	"src/packages/test-fixtures/src/providers/auth/in-memory-auth.ts", // session boundary (in-memory)
+	"projects/hutch/src/runtime/providers/auth/dynamodb-auth.ts", // session boundary (DynamoDB)
+];
+
+function* sourceFiles(dir: string): Generator<string> {
```
FagnerMartinsBrack: @claude instead of writing a custom function can you use eslint to define that the import can only be used by the files allowed to call the authenticatedUserIdFrom() and fail the lint for any call site beyond the allowed ones? Report on options. Check this for reference: https://mtsknn.fi/blog/eslint-import-restrictions/

FagnerMartinsBrack: @claude implement option A using biome

FagnerMartinsBrack:

## Context
- **PR**: #772
- **Commit**: db7fb72a9fef62c731d827b8ab9cad49f2895200

FagnerMartinsBrack:

## Context
- **PR**: #772
- **Branch**: claude/authenticated-userid-lockin

FagnerMartinsBrack:

## Context
- **PR**: #772
- **Commit**: f6e1ab9c6327ab9d27876a711897718e0773db06

PR #804, opened 2026-06-23
test(extract-links): clarify PSL differential block documents, not guards

FagnerMartinsBrack:
## Summary

The `describe` block at line 464 of `extract-links-from-page.test.ts` compares the `tldts` `getDomain` library against test-only `lastTwoLabels`/`lastThreeLabels` helpers. It never exercises the production same-site decision (`GET_DOMAIN_OPTIONS` / `isSameSite` / `harvestLinks`), so it **documents** why a Public Suffix List is needed rather than **guards** the behavior.

This renames the describe block so its name makes that intent explicit and points to the three behavioral tests that already pin the production decision through `initExtractLinksFromPageUrl`:

- `bbc.co.uk` — keeps a different registrable domain that shares a multi-part public suffix
- `ycombinator` — drops parent/sibling subdomains sharing the registrable domain
- `github.io` — keeps sibling tenants of a PSL private-section platform

### Why rename instead of fold into a comment

The alternative — folding the contrast into a comment on `GET_DOMAIN_OPTIONS` — was rejected because:

- Project guidelines treat discretionary comments as a last resort requiring explicit human approval.
- `GET_DOMAIN_OPTIONS` and `isSameSite` already carry comments explaining the PSL rationale.
- Keeping the differential tests preserves them as executable documentation that can't silently rot, while the rename removes the ambiguity about what they guard.

## Verification

`pnpm nx run extract-links-from-page:check` passes — 29 tests green, 100% coverage across statements/branches/functions/lines.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MmwN2GhNXPTpfeaQ6pqkx2)_

### DIFF
--- a/src/packages/extract-links-from-page/src/extract-links-from-page.test.ts
+++ b/src/packages/extract-links-from-page/src/extract-links-from-page.test.ts
@@ -461,7 +461,7 @@ describe("initExtractLinksFromPageUrl", () => {
 	});
 });
 
-describe("registrable-domain rule: getDomain({ allowPrivateDomains: true }) over a hand-crafted label rule", () => {
+describe("documents why a Public Suffix List, not label-slicing, decides the registrable domain (production guarded by the bbc.co.uk, ycombinator, and github.io behavioral tests above)", () => {
 	const lastTwoLabels = (host: string) => host.split(".").slice(-2).join(".");
 	const lastThreeLabels = (host: string) => host.split(".").slice(-3).join(".");
 


FagnerMartinsBrack:

## Context
- **PR**: #804
- **Commit**: ad62a62e98d80c8d00d134a3dd38a495f942c5fd

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The new describe name relocates rot risk rather than eliminating it

The PR's stated rationale for renaming instead of commenting is that comments rot. That reasoning is sound, but the chosen name reintroduces the same rot vector in test-name form:

```
"documents why a Public Suffix List, not label-slicing, decides the registrable domain
 (production guarded by the bbc.co.uk, ycombinator, and github.io behavioral tests above)"
```

Two parts can drift with no mechanism keeping them in sync:

- **"above"** is positional. If this block is moved above the `initExtractLinksFromPageUrl` describe (or the referenced tests are relocated), the word becomes wrong.
- **"bbc.co.uk, ycombinator, and github.io"** is a curated enumeration. It's a 1:1 mapping to the three differential `it`s, but the production same-site path is *also* guarded by the substack (line 165) and IP-fallback (line 131) tests, which aren't named. If a sibling behavioral test is renamed, reordered, or added, the enumeration silently goes stale — the precise failure mode the PR cites to avoid a comment.

This is genuinely LOW: the block sits directly below the tests it references, the fixture domains are stable, and the worst case is a slightly-inaccurate test name (no test or production-behavior impact). Worth naming because it engages the PR's own justification — the rename moves the rot surface rather than closing it.

A shorter name carries the same intent with less to drift, since the three differential `it` descriptions already make the contrast concrete — e.g. `"documents why a Public Suffix List, not label-slicing, decides the registrable domain (production behavior pinned by the same-site tests above)"`, or drop the parenthetical entirely. Author's call; not blocking. [Fix this →](https://claude.ai/code?q=In%20src%2Fpackages%2Fextract-links-from-page%2Fsrc%2Fextract-links-from-page.test.ts%20(branch%20claude%2Fkeen-goldberg-lgst2i%2C%20line%20464)%2C%20shorten%20the%20describe%20block%20name%20so%20it%20drops%20the%20brittle%20enumerated%2Fpositional%20cross-reference%20(%22the%20bbc.co.uk%2C%20ycombinator%2C%20and%20github.io%20behavioral%20tests%20above%22)%20that%20can%20drift%20if%20those%20sibling%20tests%20are%20renamed%2C%20reordered%2C%20or%20added%20to%2C%20while%20keeping%20the%20%22documents%20why%20a%20Public%20Suffix%20List%2C%20not%20label-slicing%2C%20decides%20the%20registrable%20domain%22%20intent.&repo=Readplace/readplace.com)

---

**Verification performed:**
- Confirmed the differential block (lines 464–488) exercises only `tldts` `getDomain` plus the test-local `lastTwoLabels`/`lastThreeLabels` helpers — it never imports or calls `initExtractLinksFromPageUrl`, `harvestLinks`, `isSameSite`, or `GET_DOMAIN_OPTIONS`. Notably, line 482's test hardcodes `{ allowPrivateDomains: true }` rather than reading `GET_DOMAIN_OPTIONS`, so it would still pass even if production flipped that flag — corroborating "documents, not guards."
- Confirmed the three named behavioral tests (ycombinator line 95, bbc.co.uk line 114, github.io line 148) all drive the real path `initExtractLinksFromPageUrl → harvestLinks → isSameSite → getDomain(GET_DOMAIN_OPTIONS)`, and all sit above this block.
- Test count is 29, matching the PR's reported `check` result. Coverage is unaffected by construction (the change is a single `describe` string literal — no branch/statement/function delta), so I did not re-run the suite.
- No redundancy concern with the existing `GET_DOMAIN_OPTIONS`/`isSameSite` comments: those explain *why production* uses the PSL; the differential tests are executable documentation of the *consequence* that a comment can't provide (they'd fail if `tldts` changed). Not triplication.

PR #805, opened 2026-06-23
test(hutch): derive oauth test access token from userId

FagnerMartinsBrack:
## What

In `projects/hutch/src/runtime/web/test-helpers/oauth-token.ts`, the minted `accessToken` string is now derived from the `userId` (`` `test-access-token-${userId}` ``) instead of the constant `"test-access-token"`.

## Why

The in-memory OAuth model's token map is keyed by `accessToken` (`deps.tokens.set` keys by `accessToken`). With a constant token string, a future test that mints tokens for two different users within a single harness would collide on the same key, with the second mint silently overwriting the first. Deriving the key from the `userId` keeps each minted token distinct.

`UserId` is a branded string (`z.string().brand<"UserId">()`), so the value interpolates directly — the `.value` in the task description was illustrative; there is no wrapper object.

## Verification

`pnpm check` passes (full nx check across all projects). The route tests that use this helper (`api.routes.test`, `auth.route.test`, `queue.save-content`, `queue.save-html`) pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XK3oExsiKpwjueFZe8Nhvy)_

### DIFF
--- a/projects/hutch/src/runtime/web/test-helpers/oauth-token.ts
+++ b/projects/hutch/src/runtime/web/test-helpers/oauth-token.ts
@@ -10,7 +10,7 @@ function createTestToken(userId: UserId): Token {
 	return {
 		accessToken: `test-access-token-${userId}`,
 		accessTokenExpiresAt: new Date(Date.now() + 3600000),
-		refreshToken: "test-refresh-token",
+		refreshToken: `test-refresh-token-${userId}`,
 		refreshTokenExpiresAt: new Date(Date.now() + 30 * 24 * 3600000),
 		client: {
 			id: EXTENSION_CLIENT_ID,
--- a/projects/hutch/src/runtime/web/test-helpers/oauth-token.ts
+++ b/projects/hutch/src/runtime/web/test-helpers/oauth-token.ts
@@ -8,7 +8,7 @@ const TEST_USER_ID = UserIdSchema.parse("test-user-123");
 
 function createTestToken(userId: UserId): Token {
 	return {
-		accessToken: "test-access-token",
+		accessToken: `test-access-token-${userId}`,
 		accessTokenExpiresAt: new Date(Date.now() + 3600000),
 		refreshToken: "test-refresh-token",
 		refreshTokenExpiresAt: new Date(Date.now() + 30 * 24 * 3600000),


FagnerMartinsBrack:

## Context
- **PR**: #805
- **Commit**: fc5c33585d87597d2b2199dbf13e8e3c7ef6d086

FagnerMartinsBrack:

None found

## Medium Priority Issues

### 1. The fix only covers half the collision — `refreshToken` is still a constant map key

The PR's rationale is that the in-memory model keys its token store by the minted string, so two users minted in one harness collide. That's correct for the access-token map:

```ts
// src/packages/test-fixtures/src/providers/oauth/oauth-model.ts:198
deps.tokens.set(token.accessToken, stored);              // now distinct per user ✅
// :200
if (refreshToken) {
  deps.refreshTokenIndex.set(refreshToken, token.accessToken);  // still keyed by the CONSTANT ❌
}
```

`createTestToken` still mints `refreshToken: "test-refresh-token"` (`oauth-token.ts:13`). So for the *exact* two-users-in-one-harness scenario this PR sets out to fix, the second mint silently overwrites the first user's entry in `refreshTokenIndex`, and a `getRefreshToken("test-refresh-token")` would resolve to the wrong user's access token. The PR's own wording — "keeps each minted token distinct" — isn't fully true: the refresh token isn't distinct.

This is **not a regression and nothing breaks today** (the constant `refreshToken` predates this PR, and no current caller mints two users *and* exercises the refresh grant through this helper — the route tests only send the bearer access token). But it's worth landing in the same change, because:

1. It directly completes the stated intent — otherwise a future author trusts this helper for multi-user minting (the reason `saveAccessTokenForUser` takes a `userId`) and gets bitten on the refresh side.
2. It matches the repo's own convention. Sibling route tests that need distinct tokens already suffix **both**: `queue.save-intent.route.test.ts:18,20` and `queue.save-content.route.test.ts:23,25`. This helper now suffixes only the access token.

Suggested one-liner, mirroring the change already made:

```ts
refreshToken: `test-refresh-token-${userId}`,
```

`RefreshTokenSchema` is an unconstrained `z.string().brand<"RefreshToken">()` (`oauth.schema.ts:9`) and no test asserts the literal refresh value, so this is safe.

## Low Priority Suggestions

### 1. The `queue.*` route tests duplicate this helper's token-minting logic

`queue.save-html.route.test.ts:16-38`, `queue.save-content.route.test.ts`, and `queue.save-intent.route.test.ts` each define their own local `createTestToken`/`createAccessToken` rather than importing the shared `oauth-token.ts` helper that `api.routes.test.ts` and `auth.route.test.ts` use. Pre-existing and out of scope for a one-line change, but it's why the distinctness fix must be reasoned about in four places. Worth a follow-up to migrate those onto the shared helper. (`queue.save-html` also still uses `"test-access-token"` and `as UserId`/`as Client` casts that the shared helper already avoids.)

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #805
- **Branch**: claude/sleepy-archimedes-6glo4d

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #805
- **Commit**: 8d1e4ef5e533e2bea27dfcdcf1f9b9d5c7026ce8

PR #806, opened 2026-06-23
refactor(ios): drop POC naming and prep app for deployment

FagnerMartinsBrack:
## What & why

Removes every "POC" reference from the iOS app to get it deployment-ready, tidies comments in favour of self-documenting code, and applies a review against the repo's Swift skills (CLAUDE.md).

The app's **App Store identity was already POC-free and is unchanged**: bundle id `com.readplace`, display name `Readplace`, module name `Readplace`. The "POC" only lived in internal Xcode/build/CI names and comments.

## Renames
- **Xcode**: target/scheme/product `ReadplacePOC` → `Readplace`, `@main` struct `ReadplacePOCApp` → `ReadplaceApp` (file renamed), test target `ReadplacePOCTests` → `ReadplaceTests`, `ReadplacePOC.xcodeproj` → `Readplace.xcodeproj`. The now-redundant `PRODUCT_MODULE_NAME` override is dropped (module name derives from the target, so `@testable import Readplace` still resolves and `TEST_HOST` is auto-regenerated by XcodeGen).
- **nx project + directory**: `projects/ios-readplace-poc` → `projects/ios-readplace` (`ci.yml` path filters/working-dirs and `built-in-clients.ts` cross-reference updated).
- **TestFlight tag namespace**: `ios-readplace-poc@v*` → `ios-readplace@v*`.

## Deployment continuity (important)
The publish workflow derives the next build number from git tags, and there are live `ios-readplace-poc@v1`…`@v9` tags driving it. To avoid resetting the counter (CFBundleVersion must only ever increase), the version step now reads **both** namespaces and takes the numeric max, so the **next build stays `v10`** and all future builds tag `ios-readplace@vN`. Verified locally: legacy-only → 10, mixed → 11, empty → 1.

## Skills review fixes (CLAUDE.md Swift guidance)
- **`TokenStore`**: replaced the silent `UserDefaults(suiteName:) ?? .standard` fallback (the exact anti-pattern named under *No Default In-Memory Implementations*) with a `preconditionFailure`, matching the existing `makeWebAuthFlow` precedent. *(Behaviour change: a build with a genuinely missing App Group entitlement now fails fast instead of silently degrading. A correctly-provisioned/signed build always has it.)*
- Removed debug `NSLog` scaffolding from the share extension — one logged a saved URL (user PII), which the logging guidance prohibits.
- Dropped POC framing and redundant "what" comments throughout; kept the cross-system "why" docs (OAuth redirect-URI exact-match, App Group invariants, HATEOAS/transport rationale).

### Reviewed, flagged for a follow-up (not changed here)
- `OAuthService` uses the `Service` suffix (vs. *No Design Pattern Names*); the rename target isn't obvious and it touches a test file, so left as-is for a deliberate decision.
- Branded types for `Article`/`id` strings; force-unwraps of compile-time-constant URLs are acceptable under the documented literal exception. Storing tokens in the App Group `UserDefaults` rather than the Keychain remains a known limitation.

## Verification
- Workflow YAML validated; build-counter shell logic unit-checked locally.
- Swift can't be compiled in this Linux environment — it's covered by CI's macOS `ios-tests` job post-merge, which gates the TestFlight publish (a broken build can't ship).
- The only Node change is a one-line comment in `built-in-clients.ts`, exercised by the PR's `web-tests` (`pnpm check`).

https://claude.ai/code/session_013irbChJYdgMbR8Lnv3rgtk

---
_Generated by [Claude Code](https://claude.ai/code/session_013irbChJYdgMbR8Lnv3rgtk)_


FagnerMartinsBrack:

## Context
- **PR**: #806
- **Commit**: 6663b482821530678b1dd79db7ca85b6e4f4e433

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. PR description's build-number example is now stale (`v10` → will actually be `v11`)
The description says the next build "stays `v10`", but the repo now also has `ios-readplace-poc@v10`, so the dual-namespace `max + 1` logic produces **11** next. Not a code defect — `CFBundleVersion` stays monotonic (10 → 11) and all three claimed cases check out. Only the description's literal numbers drifted.

### 2. `TokenStore` precondition guards suite-name misuse, not the missing-entitlement case the description names
Removing `UserDefaults(suiteName:) ?? .standard` for a `preconditionFailure` is exactly the CLAUDE.md-prescribed fix and a net improvement. Nuance: `UserDefaults(suiteName:)` returns `nil` only for the bundle-id/global-domain suite names, not for a missing App Group entitlement (which returns a non-shared instance), so the guard won't fire in the scenario the description names. Code change is correct regardless.

### 3. (Optional) `OAuthService` `Service` suffix — correctly deferred to a follow-up.

PR #783, opened 2026-06-23
feat(inbox): email forwarding addresses behind ?feature=email (M1)

FagnerMartinsBrack:
## Email Forwarding — Milestone 1

Lays the foundation for the email-forwarding feature: every user gets opaque forwarding addresses `in-<token>@<domain>`. **M1 builds only the address itself** — minting, listing, disabling, and displaying it. Nothing receives or parses mail yet (that's M2+).

Everything is **hidden by default** behind the existing per-request `?feature=email` toggle, so production traffic sees no new surface until the flag is flipped on a request. The flag gates *visibility*, never *quality* — the table, store, error handling, and tests are all production-grade.

### What a flagged user can do
Open the new **Inbox** nav entry → land on **"Your forwarding addresses"** → see their addresses, **copy** each, **create** more (one per newsletter), and **disable** ones they no longer want.

### Deliverable 1 — token + mapping (internal)
- **Domain** (`@packages/domain/inbox`): branded `InboxToken` / `InboxAddress` zod schemas, a 6-char lowercase base36 token generator (`randomInt`, unbiased), and a pure `buildInboxAddress({ token, domain })`.
- **Store** (`inbox-addresses` DynamoDB table): PK `address` (globally unique; answers the future receive-path lookup), GSI `userId-index` (lists a user's addresses — **1:N**). `createAddress` does a conditional put `attribute_not_exists(address)` with bounded retry, **throwing on exhaustion** so the failure surfaces to alerting. Addresses are **never deleted** — a freed hash could be re-minted for another user and leak their mail — so `disableAddress` stamps a nullable `disabledAt` and the row stays.
- **Provisioning at signup**: one address is auto-created at account creation, wired once at the `createApp` composition seam so **all** signup paths (password, trial, checkout, Google) are covered. Best-effort: a failure is logged for alerting but never blocks signup.

### Deliverable 2 — gated page + nav
- `emailFeatureEnabled` threads `BannerState → NavProps → buildNavGroups`; the shell re-derives it from the request query (no cross-project import — `EMAIL_FEATURE` is the single shared constant).
- `/inbox` route (GET list, POST create, POST disable) 404s without the flag. Manage-addresses page is HTML-only and `noindex`.
- **Copy** uses a zero-JS baseline (selectable read-only field) plus a progressive-enhancement button revealed only when `navigator.clipboard` exists.

### Decisions reflected (J1–J6)
| | |
|---|---|
[...]

PR #808, opened 2026-06-24
feat(inbox): surface a graceful error when create fails

FagnerMartinsBrack:
## Graceful UI error for failed `POST /inbox/create`

Follow-up to #783. The create catch in `inbox.page.ts` previously logged the failure and silently redirected to `/inbox?feature=email`, so a failed create looked like a no-op to the user — contradicting the PR #783 decisions table, which promises a **"graceful UI error"** on collision/creation failure.

### What changed
- **`inbox.page.ts`** — the create catch now redirects to `/inbox?feature=email&error=create` (instead of the plain inbox path). The GET route reads `req.query.error === "create"` and threads `createFailed` into the page. This mirrors the established `/account` error-redirect pattern (`?error=payment_method`), so it's zero-JS and survives a full page load.
- **`inbox.component.ts`** — `InboxPage` accepts an optional `createFailed` flag.
- **`inbox.template.html`** — renders an `role="alert"` banner (`data-test-inbox-error`) above the list when `createFailed` is set: *"We couldn't create a forwarding address just now. Please try again in a moment."*
- **`inbox.styles.css`** — `.inbox__error` styling using the shared `--error` token.

### Tests
- Updated the existing failed-create test to assert the new redirect location (`…&error=create`).
- Added a route test that follows the redirect and asserts `[data-test-inbox-error]` renders.
- Added a route test asserting the indicator is absent on a normal visit.

`pnpm check` green across all 34 projects (type-check, lint, tests, 100% coverage, knip, unused-css).

---
_Generated by [Claude Code](https://claude.ai/code/session_011kHoTsoqpXg9NGf3nJFTAp)_

### DIFF
--- a/projects/hutch/src/runtime/web/pages/inbox/inbox.component.ts
+++ b/projects/hutch/src/runtime/web/pages/inbox/inbox.component.ts
@@ -13,8 +13,12 @@ const INBOX_SCRIPT = `<script src="/client-dist/inbox.client.js" defer></script>
  * the gated POST routes stay reachable when submitted from the flagged page. */
 const INBOX_QUERY = `?feature=${EMAIL_FEATURE}`;
 
-export function InboxPage(params: { addresses: InboxAddressEntry[] }): PageBody {
+export function InboxPage(params: {
+	addresses: InboxAddressEntry[];
+	createFailed?: boolean;
+}): PageBody {
 	const content = render(INBOX_TEMPLATE, {
+		createFailed: params.createFailed === true,
 		hasAddresses: params.addresses.length > 0,
 		addresses: params.addresses.map((entry) => ({
 			address: entry.address,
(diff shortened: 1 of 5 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #808
- **Commit**: b581c055ef05228ca7f90d6e57f0bc799f51dcc9

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Error-banner background diverges from the established `--error-bg` convention
`inbox.styles.css:28` sets `background: var(--card)` for `.inbox__error`, but the two existing error banners in the codebase — `.import__upload-error` (`import.styles.css:284`) and the auth error banner (`auth.styles.css:139`) — both use `background: var(--error-bg)`. The PR body says it uses "the shared `--error` token", which holds for the text/border but not the background. The result is a card-coloured box with a red border rather than the tinted error fill used elsewhere. Readable either way, so this is purely a visual-consistency nit — switch to `var(--error-bg)` unless `--card` was a deliberate choice.

### 2. Optional: query parsing is inlined rather than extracted like `/account`
The cited `/account` pattern factors its query read into a tested `parseAccountQuery` helper plus an `ACCOUNT_ERROR_PAYMENT_METHOD_URL` constant (`account.view-model.ts:65`, `account.url.ts:11`). Here both the read (`req.query.error === "create"`, `inbox.page.ts:41`) and the URL (`inboxCreateFailedPath`, `inbox.page.ts:26`) are inlined. For a single error flag this is genuinely simpler and I would **not** extract it — noting only because the PR frames itself as mirroring that pattern. No change needed.

PR #809, opened 2026-06-24
fix(inbox): mirror DynamoDB ownership guard in in-memory inbox address fixture

FagnerMartinsBrack:
## Problem

The in-memory `InboxAddress` fixture and the production DynamoDB adapter disagreed on what `disableAddress` does for a **missing** or **non-owned** address:

- `in-memory-inbox-address.ts` silently **no-op'd** — returned without writing.
- `dynamodb-inbox-address.ts` uses `ConditionExpression: "userId = :uid"`, which **throws `ConditionalCheckFailedException`** in those cases.

That ownership guard is deliberate defense-in-depth — `inbox.page.ts` refers to it as "the (also ownership-guarded) store write" so a forged address for someone else's row can never disable it. Because the fixture diverged from production, a test exercising a forged/missing address would pass against the fixture while production threw, violating the test-driven-design rule that in-memory fixtures must mirror production behaviour.

## Change

Picked the **throw** contract (keeping production's security guard intact, untouched) and aligned the fixture to it:

- `disableAddress` in the in-memory fixture now throws `ConditionalCheckFailedException` when the address is missing or owned by another user, matching the DynamoDB adapter exactly.
- Updated the two affected fixture tests (`rejects a disable for an address that does not exist`, `rejects a disable requested by a non-owner`) to assert the throw rather than a silent no-op.
- Added `@packages/hutch-storage-client` as a workspace dependency of `@packages/test-fixtures` (source of the shared exception type).
- Documented the shared ownership-guard contract on the `InboxAddressStore.disableAddress` interface in `@packages/domain`.

## Verification

- `pnpm nx run @packages/test-fixtures:check` — passes, 100% coverage on the fixture.
- Full `pnpm check` (pre-commit hook) — passes across all 34 projects.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L98gpRMRpyub3ZKUWNS9mW)_

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1240,6 +1240,9 @@ importers:
       '@packages/hutch-logger':
         specifier: workspace:*
         version: link:../hutch-logger
+      '@packages/hutch-storage-client':
+        specifier: workspace:*
+        version: link:../hutch-storage-client
       '@packages/provider-contracts':
         specifier: workspace:*
         version: link:../provider-contracts
--- a/src/packages/domain/src/inbox/inbox-address.types.ts
+++ b/src/packages/domain/src/inbox/inbox-address.types.ts
@@ -23,5 +23,9 @@ export interface InboxAddressStore {
 	 * be absent from the result — so callers must not assume read-your-write. The
 	 * in-memory fixture is strongly consistent and so will not surface the lag. */
 	listAddressesByUserId: (userId: UserId) => Promise<InboxAddressEntry[]>;
+	/** Stamps `disabledAt` on an address the user owns. The write is
+	 * ownership-guarded: disabling an address that does not exist or belongs to
+	 * another user throws `ConditionalCheckFailedException` rather than silently
+	 * succeeding, so a forged address can never reach a foreign row. */
 	disableAddress: (input: { userId: UserId; address: InboxAddress }) => Promise<void>;
 }
--- a/src/packages/test-fixtures/package.json
+++ b/src/packages/test-fixtures/package.json
@@ -166,6 +166,7 @@
     "@packages/domain": "workspace:*",
     "@packages/extract-links-from-page": "workspace:*",
     "@packages/hutch-logger": "workspace:*",
+    "@packages/hutch-storage-client": "workspace:*",
     "@packages/provider-contracts": "workspace:*",
     "@packages/web-test-harness": "workspace:*",
     "zod": "4.4.3"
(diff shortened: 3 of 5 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #809
- **Commit**: 6130d36d154753a93c1e4e543fca157d73aad46a

FagnerMartinsBrack:
### Review summary

This is a clean, well-reasoned fix. It aligns the in-memory fixture to the **real** production contract (fake follows real, never the reverse) and keeps the production security guard untouched. I verified the behaviour, the exception identity, coverage, and the blast radius — all sound.

**What I verified:**
- **Faithful mirror.** The fixture's `if (row === undefined || row.userId !== userId) throw …` exactly reproduces the DynamoDB adapter's `ConditionExpression: "userId = :uid"` semantics: a `UpdateItem` against a missing item *or* a row whose `userId` differs both fail the condition and throw `ConditionalCheckFailedException`. The owned-and-present happy path still stamps `disabledAt`. ✔
- **Same exception identity.** Both the fixture and the test import `ConditionalCheckFailedException` from `@packages/hutch-storage-client` (which re-exports the single `@aws-sdk/client-dynamodb` copy), matching `dynamodb-inbox-address.ts`. So `rejects.toThrow(ConditionalCheckFailedException)` checks the *real* class and `instanceof` holds. ✔
- **Throw is safely absorbed, never reached in normal flow.** `inbox.page.ts` `/disable` pre-checks `owned.some(...)` before calling the store, so the store-level throw is genuine defense-in-depth. And Express 5 (5.2.1) auto-forwards async rejections to the error handler, so even a TOCTOU race can't hang the request. ✔
- **Blast radius clean.** The only non-test callers are `app.ts:449` (dev/e2e composition) and `fixture.ts:385`, both routed through the `/disable` ownership guard — nothing relied on the old silent no-op, consistent with full `pnpm check` passing. ✔
- **Coverage.** All three arms of the compound `||` are exercised (missing, non-owner, owned). The two tests were correctly flipped from no-op assertions to throw assertions, and additionally assert no partial mutation. ✔
- **Conventions.** No `as`, no `process.env`, no raw `console`, no noop logger. The `throw` (rather than `assert`) is correct — the point is to emit the *specific* `ConditionalCheckFailedException` type, which `assert` cannot do. ✔

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Domain interface contract names an AWS-SDK-specific exception type (optional / defensible)
`src/packages/domain/src/inbox/inbox-address.types.ts:26-30` — the new `disableAddress` JSDoc documents the contract in terms of `ConditionalCheckFailedException`, an `@aws-sdk/client-dynamodb` type, inside the otherwise persistence-agnostic `InboxAddressStore` interface. It's prose-only (no import, so no real coupling), but a domain-layer reader now has to know a DynamoDB exception class. This is defensible and arguably correct — naming the concrete observable type is what binds *both* implementations to throw the same thing, which is the faithfulness this PR establishes; an abstract wording would let a future fixture throw a plain `Error` and silently diverge again. Author's call.

PR #788, opened 2026-06-23
feat(hutch): bridge bearer auth to a session cookie for in-app webviews (server — deploy first)

FagnerMartinsBrack:
## iOS EPIC 3 — Server half (deploy this first)

> [!IMPORTANT]
> **Deploy ordering: this server PR must ship and deploy *before* the iOS client PR.**
> Both changes are additive, so this is safe to deploy on its own — older shipped
> clients ignore the new action and endpoint. The iOS TestFlight build depends on
> the `update-status` action and `POST /auth/session` existing in production.
> **Client PR (deploy second): #789** (`claude/sleepy-archimedes-6glo4d-ios`).

### Context
TestFlight feedback asked for *swipe-to-mark-read* and an *in-app authenticated reader* (Readplace reader + AI summary) instead of opening the original source site. The reader page (`/queue/:id/view`) and its htmx poll/mutation XHRs are **cookie-session** authenticated (`hutch_sid`), but the iOS app only holds an **OAuth bearer** token. This PR adds the two additive server surfaces that bridge that gap; the iOS client consumes them in #789.

### Changes
- **`update-status` Siren action** on each saved-article sub-entity (`article-siren.ts`). Entity-level, posts `status` as `application/x-www-form-urlencoded` to the existing `POST /queue/:id/status` route (no route change needed — it already validates `req.body.status` via `ArticleStatusSchema` and 303-redirects). The browser extension's `delete` action is **retained** alongside it.
  - Named `update-status` (capability), not `mark-read` (domain state), per the extension-api-design skill.
- **`POST /auth/session`** (`auth.page.ts`): reads `Authorization: Bearer`, validates it with `validateAccessToken`, mints a session via `createSession`, sets the `hutch_sid` cookie (reusing the existing `sessionCookieOptions`), and returns **204** (no body, no redirect). 401 on missing/invalid bearer.
  - Grants **no new privilege** — the bearer already authorizes the full API. CSRF-safe (browsers never auto-attach bearer tokens). Cookie is httpOnly, host-scoped, `SameSite=Lax`, and `Secure` in HTTPS deployments. No `:id` coupling and no redirect → no open-redirect surface, reusable for any future authenticated webview.
  - Wires `validateAccessToken` into `initAuthRoutes` (`server.ts`); the value already existed at the composition root.

No new env vars, no Lambda/EventBridge changes.

### Tests (all green via `pnpm nx run hutch:check`)
[...]

PR #811, opened 2026-06-24
feat(extension): open the private reader from saved-article links

FagnerMartinsBrack:
## Why

Clicking a saved-article link in the browser extension opened the **public** `/view/<url>` page instead of the user's **private reader**, even though they're logged in to the extension.

Root cause is an auth-channel mismatch, not the link itself:

- The popup link is already correct: `item.readUrl ?? item.url`, where `readUrl` is `…/queue/:id/view` (the private reader).
- `/queue/:id/view` resolves the owner from the **`hutch_sid` session cookie** (`req.userId`), never from a bearer. It's also declared **before** `dualAuth`, so an anonymous request is silently redirected to the public `/view` page rather than to `/login`.
- The extension only holds an **OAuth bearer**. Opening a reader link spawns a normal browser tab that carries no `hutch_sid` cookie → anonymous → public view.

PR #788 shipped `POST /auth/session` (bearer → session cookie) "server — deploy first" for exactly this, but no client used it yet, and the endpoint wasn't wired for the extension's cross-origin credentialed call.

## What

**Client (`browser-extension-core`, inherited by both Chrome and Firefox — no popup/manifest change):**
- New `Auth.ensureWebSession()` mints the `hutch_sid` cookie from the stored bearer via `POST /auth/session` with `credentials:"include"`.
- `core` fires it whenever the popup is about to surface reader links — the reading-list load (`fetch("reading-list")`) and the "not-saveable" drop-into-list path. The existing `readUrl ?? url` anchor then opens the authenticated private reader.

**Server (`hutch`):**
- `/auth/session` is the first extension endpoint to use a **credentialed** cross-origin POST, so it gets scoped CORS — `OPTIONS` preflight + `Access-Control-Allow-Credentials: true` — mirroring the existing `extensionCors` pattern used by `/`, `/queue`, `/oauth/token`. Credentials are **not** loosened for the bearer-only endpoints.
- The `hutch_sid` cookie stays `SameSite=Lax` + `HttpOnly` + `Secure`. Under `host_permissions`, Chromium treats the extension fetch as first-party to `readplace.com`, so a Lax cookie both stores and is sent on the subsequent top-level navigation.

## Verification

- `pnpm check` green for `hutch`, `browser-extension-core`, `chrome-extension`, `firefox-extension` (100% coverage held; the pre-commit hook ran the full workspace check).
- New unit tests: `ensureWebSession` posts the bearer with `credentials:"include"` (and no-ops logged-out); `core` mints on list-fetch and on the not-saveable save, swallows mint failures.
- New integration tests: `OPTIONS /auth/session` preflight returns the reflected origin + `Allow-Credentials: true`; the credentialed POST returns those headers and still mints the cookie.
- Validated by a 4-lens adversarial review (runtime cookie deposit, server reachability, diff/coverage, failure modes) — the two CORS blockers it found are fixed here.

## Known limitations / follow-ups (from the review)

These degrade gracefully to the **public** reader (today's behaviour), so they don't regress anything, but are worth a follow-up:

1. **First-open race.** The mint is fire-and-forget, so on the very first popup open of a browser session (cookie cleared on browser close — no `maxAge`), a click *before* the `POST /auth/session` round-trip lands has no cookie yet. Steady-state is fine (re-minted on every list open). The fully race-free option is to intercept the anchor click → await the mint → open the tab; not done here to preserve native anchor affordances (middle/cmd-click, copy-link).
2. **Real-browser confirmation (esp. Firefox).** The cookie deposit + top-level-nav send is verified by unit/integration tests, not an E2E in a real extension build. Worth an E2E asserting a reader link lands on `/queue/:id/view`, not `/view` — particularly on Firefox, where the cross-origin credentialed handling is stricter.
3. **Session-row churn.** Each mint calls `createSession`, writing a fresh (TTL'd) `hutch_sid` row per popup-list-open; could be short-circuited when a valid cookie already exists.
4. **Local/dev is a no-op** (`HUTCH_SERVER_URL=http://127.0.0.1:3000` ≠ the `host_permissions` host), so this is only exercisable against an https `readplace.com` build.

## Test plan

- [ ] Load the built extension against `readplace.com`, log in, open the popup, click a saved article → lands on the private `/queue/:id/view` reader (Mark-as-read button + progress), not `/view`.
- [ ] Repeat in Firefox.

### DIFF
--- a/projects/browser-extension-core/src/auth/auth.types.ts
+++ b/projects/browser-extension-core/src/auth/auth.types.ts
@@ -23,6 +23,12 @@ export type RefreshTokens = () => Promise<RefreshResult>;
 
 export type GetAccessToken = () => Promise<string | null>;
 
+/** Mints a browser session cookie from the stored bearer. The reader at
+ * /queue/:id/view authorizes the owner from the hutch_sid cookie, never a
+ * bearer, so a list link followed into a normal tab needs that cookie to land
+ * on the private reader instead of the public /view fallback. */
+export type EnsureWebSession = () => Promise<void>;
+
 export type WhenLoggedIn = <T>(fn: () => T) => GuardedResult<T>;
 
 export interface Auth {
@@ -30,6 +36,7 @@ export interface Auth {
 	logout: Logout;
 	refreshTokens: RefreshTokens;
 	getAccessToken: GetAccessToken;
+	ensureWebSession: EnsureWebSession;
 	whenLoggedIn: WhenLoggedIn;
 }
 
@@ -45,7 +52,7 @@ export interface OAuthAuthDeps {
 	openTab(url: string): Promise<number>;
 	waitForRedirect(params: { tabId: number; urlPrefix: string }): Promise<string>;
 	closeTab(tabId: number): Promise<void>;
-	fetchFn(url: string, init: { method: string; headers: Record<string, string>; body: string }): Promise<{ ok: boolean; status: number; json(): Promise<Record<string, string>> }>;
+	fetchFn(url: string, init: { method: string; headers: Record<string, string>; body?: string; credentials?: "include" }): Promise<{ ok: boolean; status: number; json(): Promise<Record<string, string>> }>;
 	tokenStorage: TokenStorage;
 	logger: HutchLogger;
 }
--- a/projects/browser-extension-core/src/auth/in-memory-auth.test.ts
+++ b/projects/browser-extension-core/src/auth/in-memory-auth.test.ts
@@ -88,6 +88,14 @@ describe("initInMemoryAuth", () => {
 		});
 	});
 
+	describe("ensureWebSession", () => {
+		it("should resolve without error", async () => {
+			const auth = initInMemoryAuth();
+
+			await expect(auth.ensureWebSession()).resolves.toBeUndefined();
+		});
+	});
+
 	describe("logout", () => {
 		it("should return not-logged-in after logout", async () => {
 			const auth = initInMemoryAuth();
--- a/projects/browser-extension-core/src/auth/in-memory-auth.ts
+++ b/projects/browser-extension-core/src/auth/in-memory-auth.ts
@@ -1,4 +1,4 @@
-import type { Auth, GetAccessToken, Login, Logout, RefreshTokens, WhenLoggedIn } from "./auth.types";
+import type { Auth, EnsureWebSession, GetAccessToken, Login, Logout, RefreshTokens, WhenLoggedIn } from "./auth.types";
 
 export function initInMemoryAuth(): Auth {
 	let loggedIn = false;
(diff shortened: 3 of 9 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #811
- **Commit**: 37e9e9b35b303d96fa7bc707750eb91a1710c92b

FagnerMartinsBrack:
### Review: open the private reader from saved-article links

Solid, well-scoped change. The auth-channel diagnosis is correct, the CORS is *narrowly* scoped to the one route that needs credentials (the comment at `server.ts:599` explains the trade-off well), the cookie stays `HttpOnly`/`Secure`/`Lax`, and the mint is correctly gated on `whenLoggedIn`. The `isAllowedExtensionOrigin` extraction is a clean behaviour-preserving refactor (`server.ts:586`). Unit + integration coverage is good and the security posture is sound — minting requires a valid bearer, the cookie is `HttpOnly`, and bearers are never auto-attached so there's no CSRF surface (`auth.page.ts:450`).

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The load-bearing behaviour has no automated coverage — verify in a real browser (esp. Firefox)
The two new integration tests (`auth.route.test.ts:1065`, `:1080`) assert *server* CORS headers + `Set-Cookie` presence via supertest. Supertest is a Node HTTP client, not a browser — it does **not** model the two things this feature actually depends on: storing a `SameSite=Lax` cookie from a cross-site, credentialed `fetch` response, and the extension fetch being treated as first-party under `host_permissions`. A green CI run does not prove a reader link lands on `/queue/:id/view`. Acknowledged by the author (follow-up #2 + unchecked test-plan boxes); flagging so reviewers don't read green checks as proof.

### 2. Fire-and-forget mint silently drops non-ok HTTP responses
`ensureWebSession` (`oauth-auth.ts:126-134`) awaits `deps.fetchFn(...)` but never inspects `.ok`/`.status`, and `establishWebSession`'s `.catch` (`core.ts:111-115`) only catches network rejections. A `401` from an expired bearer is dropped with zero signal, undercutting the verification in #1. `deps.logger` is already in scope.

### 3. `establishWebSession` re-mints on every popup list-open (session-row churn)
`fetch("reading-list")` (`core.ts:203`) and each not-saveable save (`core.ts:181`) call `createSession` every time. Author flagged it (follow-up #3). The naive "skip if a cookie exists" short-circuit is non-trivial: the mint POST is cross-site, so a `SameSite=Lax` `hutch_sid` generally won't be attached to it — the server can't see the existing cookie. Coupled to the same SameSite question as #1.

### 4. The `core.ts` comment narrows the actual trigger set
The comment at `core.ts:106-110` says the mint fires on "the list load and the not-saveable drop-into-list," but the save path fires on every `!result.ok` (`core.ts:181`) — also `already-saved` and `messages`. That's correct behaviour; only the comment's enumeration is narrow.

PR #812, opened 2026-06-24
feat(inbox): render disabled forwarding addresses as disabled inputs

FagnerMartinsBrack:
## What

On `/inbox?feature=email`, a **disabled** forwarding address used to render exactly like an active one: a selectable `readonly` field plus a (JS-revealed) **Copy** button. That's misleading — a disabled address no longer receives mail, so there's nothing useful to copy.

This change makes a disabled row clearly read as inactive:

- **Drops the Copy button** for disabled addresses (only the active address keeps a copy affordance).
- **Swaps the selectable `readonly` field for a native `disabled` `<input>`** — the browser blocks focus/selection, and a new `.inbox__address-field--disabled` modifier greys it out (`--muted` background, `--muted-foreground` text, `cursor: not-allowed`, `user-select: none`).

Active rows are unchanged.

## How

The `{{#each addresses}}` row in `inbox.template.html` now renders two complete shapes under the existing `{{#if this.enabled}}` branch:

| State | Field | Copy button | Action |
|-------|-------|-------------|--------|
| Active | `readonly`, selectable | shown (JS-revealed) | Disable form |
| Disabled | `disabled`, greyed out, no selection | removed | "Disabled" marker |

No component/route logic changed — `enabled` was already passed to the template. The copy client JS is untouched: it only wires `[data-inbox-copy]` buttons, which disabled rows no longer emit.

## Tests

- Added an `InboxPage` component test rendering a mixed `[active, disabled]` list and asserting (positively) that: the disabled field has `disabled` + the `--disabled` modifier and is not `readonly`; the active field stays `readonly`; and exactly one copy button remains, belonging to the active address.
- Extended the existing active-row test to assert the field is not `disabled`.
- `pnpm nx run hutch:check` and the full `pnpm check` (pre-commit hook, 34 projects) pass — 100% coverage thresholds met, lint/knip/unused-css green.

https://claude.ai/code/session_01KyRJm2jGGdNvTbkS9wFJDE

---
_Generated by [Claude Code](https://claude.ai/code/session_01KyRJm2jGGdNvTbkS9wFJDE)_

### DIFF
--- a/projects/hutch/src/runtime/web/pages/inbox/inbox.component.test.ts
+++ b/projects/hutch/src/runtime/web/pages/inbox/inbox.component.test.ts
@@ -43,9 +43,51 @@ describe("InboxPage", () => {
 		const field = doc.querySelector(".inbox__address-field");
 		assert.equal(field?.getAttribute("value"), "in-3f9a2c@read.place");
 		assert.equal(field?.getAttribute("readonly"), "");
+		assert.equal(field?.hasAttribute("disabled"), false);
 		const copy = doc.querySelector("[data-inbox-copy]");
 		assert.equal(copy?.getAttribute("data-inbox-address"), "in-3f9a2c@read.place");
 		assert.equal(copy?.hasAttribute("hidden"), true);
+	});
+
+	it("renders a disabled address as a disabled field and drops its copy button, leaving the copy affordance only on the active one", () => {
+		const doc = parse(
+			InboxPage({
+				addresses: [
+					entry(),
+					entry({
+						address: InboxAddressSchema.parse("in-abc123@read.place"),
+						token: InboxTokenSchema.parse("abc123"),
+						disabledAt: "2026-06-22T00:00:00.000Z",
+					}),
+				],
+			}).content.html,
+		);
+
+		const fields = Array.from(doc.querySelectorAll(".inbox__address-field"));
+		assert.equal(fields.length, 2, "both addresses still render a field");
+
+		const [active, disabled] = fields;
+		assert.equal(active.getAttribute("value"), "in-3f9a2c@read.place");
+		assert.equal(active.hasAttribute("readonly"), true);
+		assert.equal(active.hasAttribute("disabled"), false);
+
+		assert.equal(disabled.getAttribute("value"), "in-abc123@read.place");
+		assert.equal(disabled.hasAttribute("disabled"), true);
+		assert.equal(disabled.hasAttribute("readonly"), false);
+		assert.equal(
+			disabled.classList.contains("inbox__address-field--disabled"),
+			true,
+			"disabled field carries the disabled modifier",
+		);
+
+		const copyableAddresses = Array.from(doc.querySelectorAll("[data-inbox-copy]")).map((el) =>
+			el.getAttribute("data-inbox-address"),
+		);
+		assert.deepEqual(
+			copyableAddresses,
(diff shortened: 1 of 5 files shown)


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28078658005
- **PR**: #812
- **Branch**: claude/focused-dijkstra-y5qw9k

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #812
- **Commit**: 6878ea9514fb94a2b54d1f86195430a0c3be4e5d

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Unrelated test-infra change bundled into a UI PR (scope)
Commit `6878ea95` (CI-failure-fixer, attempt #1) added a one-retry cushion to the Playwright **browser-install** step in `src/packages/test-phase-runner/src/run-test-phases.ts`. The change itself is sound — it mirrors the existing `runCommandPhase` e2e retry (`run-test-phases.ts:183-194`), is idempotent (`playwright install` is re-runnable), and both new paths (retry-then-succeed, fail-twice-rethrow) are covered by the added tests. But it is unrelated to the inbox feature this PR is titled for, so the PR now carries two independent changes. This is an expected side effect of the CI-fixer committing to the PR branch; just flagging it so the human reviewer evaluates the two changes separately. Minor: the commit message `fix: resolve CI failure (attempt #1)` omits a Conventional-Commits scope (e.g. `fix(test-phase-runner):`).

### 2. Address `<input>` markup is now duplicated across both `{{#if this.enabled}}` arms
In `inbox.template.html` the `<input class="inbox__address-field" …>` is rendered in full in each branch, differing only by `readonly` + `data-inbox-address` (active) vs `disabled` + the `--disabled` modifier (disabled). The web skill ("Iterate Lists, Don't Branch in Templates") prefers building the per-row shape in TypeScript so the template renders one shape once. This is genuinely awkward here because Handlebars can't emit a conditional bare attribute (`readonly` vs `disabled`) without unescaped `{{{ }}}`, so some branching is hard to avoid — hence low priority. If this row grows more states later, consider mapping a richer per-row view object (e.g. `copyable: boolean`, `fieldState`) in `inbox.component.ts:19-22` and trimming the template branch. Non-blocking.

### 3. Negative attribute assertions are slightly redundant (optional)
The tests add `assert.equal(field?.hasAttribute("disabled"), false)` (active row) and `assert.equal(disabled.hasAttribute("readonly"), false)` (disabled row). test-driven-design's "Avoid Negative Test Assertions" prefers positive assertions. These are defensible — they pin the *mutual exclusivity* of the two states, which is the whole point of the change — but they overlap with the positive `readonly` / `disabled` assertions already made on the same elements. Keep them if you value the explicit contract; otherwise they can be trimmed.

### 4. `disabled` input and screen-reader announcement (optional, minor)
A `disabled` `<input>` is removed from the tab order and some screen readers will skip announcing its value. That's acceptable here: the adjacent `data-test-inbox-status="disabled"` label conveys the state and there's nothing actionable on a disabled address. If you ever want the value reliably announced, `readonly` + a selection-blocking class (`user-select: none; pointer-events: none`) would keep the field in the accessibility tree — but `disabled` matches the stated intent ("the browser blocks focus/selection") and is fine as-is.

PR #815, opened 2026-06-24
docs(skills): hypermedia-api-design — generalize to all clients + cross-client conformance + MCP

FagnerMartinsBrack:
## Summary

Generalize the `extension-api-design` skill into **`hypermedia-api-design`** so the Siren/HATEOAS contract doctrine surfaces for **every** client — chrome & firefox extensions, the iOS app, the MCP server, and future clients — not just the browser extension, and encode the findings from a cross-client hypermedia audit (plus fresh MCP research) as durable conformance guidelines.

Two commits, **skill-only** (plus the references that name the skill). No client code behaviour changes — the audit findings are applied as *guidelines*, not fixes.

- **Commit 1 — `docs(skills): generalize extension-api-design into hypermedia-api-design`**
  Faithful generalization: rename the skill + broaden its `description` to trigger for all client surfaces (incl. MCP) and server Siren emission; replace "the extension" wording with client-agnostic wording; relocate the extension-only mechanics (the walker pattern, `init*Understanding`/`groupOf`/`httpCacheable`) into a clearly-labelled **Per-Client Implementations** section. No doctrine added or removed. Updates every reference to the old name: the iOS README link, two `popup.browser.ts` comments, two `browser-extension-core` comments, and the four `.github/workflows/claude-*.md` skill-routing entries (also broadened to all clients).

- **Commit 2 — `docs(skills): add cross-client conformance rules and MCP to hypermedia-api-design`**
  Adds a **Client Conformance** section (8 rules from the audit), two anti-pattern rows, and an **MCP** subsection (Siren↔MCP mapping + divergences + in-repo audit). Extends the capability checklist to cover MCP tool definitions.

---

## Research part 1 — the generic hypermedia client (the yardstick)

From the linked essays + Fielding + the Siren spec, and verified against the in-repo **e2e HATEOAS harness** (the team's existing generic-client exemplar — "a runner that discovers and executes available actions… identify pages by body class, not URL… `isAvailable` must use element-based detection, not URL path checks"):

- Hardcode **exactly one URL** (the entry point); every other URL comes from a response `link.href`/`action.href` and is used **verbatim**.
- Read **method + fields** from the action; the client never decides the verb or field set.
- Render/enable a control **iff** the current response advertises the matching action/link (HATEOAS). Absence ⇒ no control.
- Code against the **media type**, not any endpoint's response shape; negotiate with `Accept` **and** verify the response `Content-Type`.
- Couple only to the **vocabulary** (action `name`s, link `rel`s) — never to URLs, methods, field names, or response shapes.

**Headline finding:** both shipping clients are hypermedia-faithful at the *transport* layer (they follow server hrefs/methods and discover by name/rel), so there are **zero "high"-severity break-on-routine-change gaps**. The divergence is concentrated in the **view layer** (controls rendered regardless of advertised affordances) and in **client-side reimplementation of server navigation** — exactly the "app is a browser, Siren is HTML" test failing. The doctrine that cures it lived in a skill scoped to the extension, invisible to iOS/MCP work; that mis-scoping is itself a contributing cause, and is what commit 1 fixes.

---

## Research part 2 — cross-client audit (browser extension + iOS)

23 verified deviations (17 iOS, 6 extension; 8 medium, 15 low after adversarial verification). Grouped by the theme each became a **Client Conformance** rule for:

| Theme → conformance rule | iOS evidence | Extension evidence |
|---|---|---|
| **Render affordances, don't assume them** (HATEOAS at the view layer) | Delete swipe always shown (`ReadingListView.swift:98-104`, silent no-op when `deleteHref==nil`); "+save" always shown (`:33-40`); `update-status` action never surfaced (`SirenModels.swift:229-231`) | Per-item delete button always rendered (`popup.browser.ts:231-257`) |
| **Navigate by the server link, not a domain property** | Tap opens raw `article.url`, ignoring the parsed `read` link (`ReadingListView.swift:94-97`; `SirenModels.swift:230`) | (correct here: `item.readUrl ?? item.url`) |
| **One scheme-aware href resolver** | `absolute()` keys on `hasPrefix("http")`, blindly prepends base otherwise (`ReadplaceAPI.swift:186-190`) | 8 action sites do `${serverUrl}${href}` with no check (`siren-reading-list.ts:195-403`); only the self-link is scheme-aware (`:583-585`) |
| **Verify the response media type** | any 200 blind-decoded as Siren (`ReadplaceAPI.swift:158-170,192-195`) | same (`siren-reading-list.ts:575-578`) |
| **Drive search/filter/pagination from server fields + links** | (correct: follows `nextHref`) | client-side pagination ignores `next`/`prev` (`siren-reading-list.ts:731-735`, `popup.browser.ts:105-265`); `search` hardcodes `url`/`status`, ignores `action.fields` (`:400-418`) |
| **Don't duplicate server policy** | `SaveSharedPage` hardcodes the server's 10 MiB cap and branches on it (`SaveSharedPage.swift:37,50`) | — |
| **Bind actions through one generic path** | per-operation bespoke decode; cherry-picks `delete`/`read` into typed fields (`ReadplaceAPI.swift:80-201`, `SirenModels.swift:229-231`) | (has the generic `understanding` map — closest to the yardstick) |
| **Parse leniently / comment + test hygiene** | one malformed control fails the whole decode (`SirenModels.swift:5-23`); comments name `/queue/*` routes; a test pins `/queue/a1/delete` | required-`href` zod → whole-collection parse throws |

**Directive-premise corrections (verified):**
- **Siren mandates `href`** on links/actions — a "no-href affordance" can't validly occur; the real issue behind point 4 is the whole-decode blast radius (one malformed control blanks the page).
- **`readplace://` is the iOS OAuth callback deep-link**, not a reading-list href scheme — moot for the data surface.
- **Point 6 (tests asserting server responses) is largely moot** — the tests assert *local fixtures*, not live responses, and production code follows actions verbatim; only cosmetic redundancy remains.
- **Point 3 (skeletons over spinners)** has thin grounding — the web codebase itself has almost no skeletons (`queue-processing-pulse`, `article-body__reader-loading` only); kept out of the conformance rules as it's not a hypermedia coupling.

---

## Research part 3 — MCP parity ("it should work similarly")

**Verdict: partially — yes in stance, partially in mechanics.** MCP is built on the same "server owns the protocol, client discovers at runtime" doctrine: a generic MCP client hardcodes only the connection (one endpoint URL or stdio command) and discovers tools/resources/prompts by descriptor over JSON-RPC — exactly as the Siren walker hardcodes only `/` and follows server-minted actions/links. So the **Client Conformance rules apply to an MCP client too.** It diverges enough to be "partially": a stateful session with a one-time `initialize` capability+version handshake (vs per-request content negotiation), server-initiated async pushes (`list_changed`/`updated` vs a synchronous `303`-back), a fixed invocation verb (`tools/call`) instead of a per-affordance `href`/`method`, and three flat global registries instead of one nested document with per-entity actions.

| Siren / HATEOAS | MCP |
|---|---|
| Single entry point `/` | Server connection — one Streamable HTTP endpoint or stdio command |
| `Accept` content negotiation, per request | `initialize` capability + protocol-version handshake, once per session |
| `actions[]` (`name`/`href`/`method`/`fields`); `name` is the contract | Tools (`tools/list` → `name`, `inputSchema`; `tools/call`); `name` is the contract |
| `action.fields` | `inputSchema` (JSON Schema) + optional `outputSchema` to validate the result |
| `links[]` by `rel`/`href`, followed not built | Resources (`resources/list` → `uri`; `resources/read`); RFC 6570 `uriTemplate` |
| Opaque `next` link; never build `?page=` | Opaque `cursor`/`nextCursor`; echo unchanged, never parse/persist |
| `303` back to the collection after a mutation | `notifications/*/list_changed`, `resources/updated` → re-list/re-read |
| Generic `messages[]` channel | JSON-RPC error (call failed) vs `isError:true` result (operation refused); typed content blocks |
| (no analog) | Prompts (`prompts/list`/`prompts/get`) — user-initiated; MCP-only |

**In-repo MCP audit:** hutch is the MCP **server** and adheres to the doctrine — the tool set is a single source of truth in `tool-definitions.ts` (`TOOL_DEFINITIONS`, 8 tools each with a JSON Schema paired to a Zod validator, kept in lock-step by `tool-definitions.test.ts`), advertised via `tools/list` in `mcp-server.ts`, with **no client copy that could drift**. There is **no in-repo MCP-consuming client** (the real clients are external assistants against `https://readplace.com/mcp`). The WebMCP surface (`webmcp.client.ts`) is a **provider** that registers one local `save_link` tool — a provider declares its own shape, so discover-don't-hardcode does not apply to it. No substantiated server-protocol deviations. (The one doc nit the audit surfaced — the skill front-matter still reading `name: extension-api-design` — is resolved by commit 1.)

Spec: https://modelcontextprotocol.io

---

## Scope & notes

- **Skill-only** change. The cross-client gaps are encoded as *guidelines*; bringing the extension/iOS clients into conformance is separate code work (kept out per the diagnosis-only steer).
- Generalizing the doc is **necessary but not sufficient**: e.g. the conformance rule "resolve hrefs by scheme" exists because both clients string-concatenate hrefs today, against the contract's own promise that changing an `href` is non-breaking.
- A single hypermedia skill (rather than a parallel MCP skill) avoids duplicating ~85% of the doctrine and the drift that would follow — the same "don't run parallel evolutions of one concept" rule the skill itself states.

Based off `main` (not PR #811).

🤖 Research via multi-agent workflows; authored with Claude Code.

### DIFF
--- a/.claude/skills/hypermedia-api-design/SKILL.md
+++ b/.claude/skills/hypermedia-api-design/SKILL.md
@@ -140,6 +140,23 @@ A client binds both levels: collection-level actions (e.g. `save-article`, `sear
 | Synthesising state after a mutation (`allItems.filter(i => i.id !== deletedId)`) | Server is the source of truth; follow the 303 and read the new collection |
 | Exporting an `/api` SDK that knows resource URLs | Becomes another versioned surface; expose only the walker and the entry point |
 | Interpolating unescaped user data into a `messages[].content.body` | The client injects it via `innerHTML`; unescaped server output is markup injection (see "Server-Driven Messages Are Trusted HTML") |
+| Naming a concrete server route or method in client code or comments (`POST /queue/save-html`, "303s to `/queue`") | The client reads `href`/`method` from the response; a route baked into a comment rots and reintroduces the URL coupling the contract removes |
+| Asserting a specific server response body/URL/shape in a client test | A client test should exercise generic protocol handling — follow whatever action/link the response carries — not pin the server's current URLs or shapes |
+
+## Client Conformance
+
+These rules make any client (extension, iOS, MCP, future) behave like a browser over the contract above — "the app is a browser, Siren is HTML." Each prevents a real cross-client failure mode.
+
+| Rule | Why |
+|---|---|
+| Render a control (button, swipe, menu item, tap target) **only if the current response advertised** the matching action/link; hide or disable it otherwise. | A browser shows a Delete button only when the HTML has one. A control for an absent action is a phantom affordance that fails silently — the worst HATEOAS failure mode. |
+| Navigate by the server-supplied link (`rel`), not by a domain property. | Opening a raw `url` property instead of following the `read` link discards the server's chosen destination; if the server re-points the link, the client never follows it. |
+| Resolve every href through one helper: use an `http(s)://` href verbatim, resolve a scheme-less href against the base, treat any other scheme as no href (unactionable). | Per-call-site `base + href` concatenation corrupts an absolute href and mis-resolves other schemes — yet the Evolvability table promises changing an `href` is non-breaking. |
+| Verify the response `Content-Type` is the negotiated media type before parsing; render a standard "unsupported media type" view for anything else. | Negotiating with `Accept` but blind-decoding any 200 body turns a proxy HTML page or a future media type into an opaque "couldn't read the response" instead of an honest "I don't understand this type." The message-content gate already does this — apply it to the response envelope too. |
+| Parse leniently: one malformed link/action must degrade to unactionable, never fail the whole response decode. | An atomic decode that rejects a single odd control blanks the entire page — extend the same tolerance the entity `properties` already get. (Note: Siren requires `href` on links/actions, so a hrefless control is a malformed control, not a valid one.) |
+| Drive search, filter, sort, and pagination from the server's `fields` and links — never client-built params or client-side re-paging. | Hardcoding `?page=` or re-paginating one fetched page client-side hides items past the first page and breaks when the server changes paging; hardcoding a filter field name breaks on a rename. |
+| Don't duplicate server policy (size caps, validation thresholds) in the client; attempt the action and follow the server's fallback/refusal. | A client copy of a server constant goes stale on a server change and mis-routes; the server already advertises the fallback action to follow. |
+| Bind a response's actions through one generic path; don't cherry-pick named affordances into per-operation code with hardcoded shapes. | Per-operation bespoke handling means each new server capability needs new client code and a redeploy; a generic action map exposes whatever the server offers. |
 
 ## Per-Client Implementations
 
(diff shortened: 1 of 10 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #815
- **Commit**: 2cba7dafde4fb73ac852e11d13c988aa6095331e

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Conformance rules encode *known-current* deviations — confirm the code follow-ups are tracked
The 8 **Client Conformance** rules are written as durable doctrine, but several describe non-conformance that exists in shipping clients **today** per the PR's own audit (iOS `absolute()` keying on `hasPrefix("http")`; ~8 extension call-sites doing `${serverUrl}${href}`; delete/`+save` controls rendered even when the action is absent; blind-decoding any 200 as Siren). The PR intentionally keeps the code fixes out of scope ("diagnosis-only"), which is reasonable — but a guideline that documents a real current gap quietly becomes aspirational if nothing tracks the fix. Non-blocking: consider opening/linking tracking issues for the code-side follow-ups so the rules don't drift from reality. (Process note, not a skill defect.)

### 2. `State Lives in the Network` still names extension-only symbols inline
Lines 82–85 reference `httpCacheable(understanding)`, `knownItems`, and `initSirenReadingList` inside what is now a client-agnostic section. Commit 1 relocated the walker mechanics into *Per-Client Implementations › Browser extension* for exactly this reason; these two bullets are the remaining extension-specific references outside that section. They're now qualified with "in the browser extension's…", so it's not wrong — just slightly inconsistent with the relocation. Optional.

PR #821, opened 2026-06-24
fix(inbox): derive limitReached from live address count

FagnerMartinsBrack:
## What

`GET /inbox` now derives `limitReached` from the live forwarding-address count rather than the `&error=limit` query flag:

```ts
const limitReached =
  addresses.filter((a) => a.disabledAt === undefined).length >= INBOX_ADDRESS_MAX_PER_USER;
```

(`INBOX_ADDRESS_MAX_PER_USER` imported from `@packages/domain/inbox`.)

## Why

Previously the per-user limit banner only appeared right after a rejected create (via the `&error=limit` redirect). Now it shows proactively whenever a user is genuinely at the cap — e.g. when they land on `/inbox` already holding the maximum number of live addresses.

The create handler still redirects to `&error=limit` on a rejected create, but that flag is no longer the sole trigger for the banner.

## Tests

- Added a route test asserting the banner shows at the cap on a plain `GET /inbox?feature=email` (no query param).
- Reworked the existing limit test to create real addresses up to the cap (the in-memory store faithfully throws `InboxAddressLimitReachedError` at the limit), keeping the `&error=limit` redirect and no-error-logging assertions while making the banner assertion consistent with the new live-count derivation.

`pnpm nx run hutch:check` passes — 2334 tests, 100% function coverage.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VUy6uXZY7W6aWZpnqRz4c8)_

### DIFF
--- a/projects/hutch/src/runtime/providers/inbox-address/dynamodb-inbox-address.ts
+++ b/projects/hutch/src/runtime/providers/inbox-address/dynamodb-inbox-address.ts
@@ -8,6 +8,7 @@ import { z } from "zod";
 import { UserIdSchema } from "@packages/domain/user";
 import {
 	buildInboxAddress,
+	countLiveAddresses,
 	generateInboxToken,
 	INBOX_ADDRESS_MAX_CREATE_ATTEMPTS,
 	INBOX_ADDRESS_MAX_PER_USER,
@@ -61,10 +62,8 @@ export function initDynamoDbInboxAddress(deps: {
 			// soft guardrail (a racing pair of creates may briefly land one over the
 			// cap) — fine, since the cap exists to stop an unbounded create loop, not
 			// to enforce an exact-to-the-row ceiling.
-			const live = (await listAddressesByUserId(userId)).filter(
-				(entry) => entry.disabledAt === undefined,
-			);
-			if (live.length >= INBOX_ADDRESS_MAX_PER_USER) {
+			const liveCount = countLiveAddresses(await listAddressesByUserId(userId));
+			if (liveCount >= INBOX_ADDRESS_MAX_PER_USER) {
 				throw new InboxAddressLimitReachedError(INBOX_ADDRESS_MAX_PER_USER);
 			}
 			const createdAt = deps.now().toISOString();
--- a/projects/hutch/src/runtime/web/pages/inbox/inbox.page.ts
+++ b/projects/hutch/src/runtime/web/pages/inbox/inbox.page.ts
@@ -3,7 +3,11 @@ import type { Request, RequestHandler, Response, Router } from "express";
 import express from "express";
 import { z } from "zod";
 import { EMAIL_FEATURE, sendComponent } from "@packages/web-shell";
-import { InboxAddressLimitReachedError, InboxAddressSchema } from "@packages/domain/inbox";
+import {
+	INBOX_ADDRESS_MAX_PER_USER,
+	InboxAddressLimitReachedError,
+	InboxAddressSchema,
+} from "@packages/domain/inbox";
 import type { InboxAddressStore } from "@packages/domain/inbox";
 import { Base } from "../../base.component";
 import type { BuildBannerState } from "../../banner-state";
@@ -47,7 +51,11 @@ export function initInboxRoutes(deps: InboxDependencies): Router {
 		assert(req.userId, "userId required - route must be protected by requireAuth");
 		const addresses = await deps.inboxAddressStore.listAddressesByUserId(req.userId);
 		const createFailed = req.query.error === "create";
-		const limitReached = req.query.error === "limit";
+		// Derive from the live count, not the &error=limit flag, so the banner
+		// shows whenever the cap is genuinely reached — not only right after a
+		// rejected create.
+		const limitReached =
+			addresses.filter((a) => a.disabledAt === undefined).length >= INBOX_ADDRESS_MAX_PER_USER;
 		sendComponent(
 			req,
 			res,
--- a/projects/hutch/src/runtime/web/pages/inbox/inbox.page.ts
+++ b/projects/hutch/src/runtime/web/pages/inbox/inbox.page.ts
@@ -4,6 +4,7 @@ import express from "express";
 import { z } from "zod";
 import { EMAIL_FEATURE, sendComponent } from "@packages/web-shell";
 import {
+	countLiveAddresses,
 	INBOX_ADDRESS_MAX_PER_USER,
 	InboxAddressLimitReachedError,
 	InboxAddressSchema,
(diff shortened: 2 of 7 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #821
- **Commit**: 3d5153f739036781554e341546d571d3c0ddf1f0

FagnerMartinsBrack:

None found

The core change is sound. The new derivation in `inbox.page.ts:57-58` exactly mirrors how both store adapters enforce the cap — same `disabledAt === undefined` filter, same `>= INBOX_ADDRESS_MAX_PER_USER` comparison (`in-memory-inbox-address.ts:16-21` and `dynamodb-inbox-address.ts:64-69`). So the banner now appears precisely when a create would be rejected. No off-by-one in the seeding loops, and swapping the hand-stubbed `createAddress` throw for real store-driven seeding is a genuine improvement.

## Medium Priority Issues

### 1. `&error=limit` is now emitted but read by nothing — code diverges from the PR description

The create handler still redirects with `&error=limit` (`inbox.page.ts:77`), but the GET handler no longer reads it — `limitReached` is derived purely from the live count. Contrast `&error=create`, which is still consumed at `inbox.page.ts:53`. The PR description says the flag is "no longer the **sole** trigger" (implying it remains one trigger), but in code it triggers nothing.

- **Dead coupling.** A redirect emits a signal that no reader consumes.
- **Narrow eventual-consistency edge.** The production GSI read is eventually consistent (`dynamodb-inbox-address.ts:39-54`); the live-count read on the redirect target can undercount, so the post-rejection banner is no longer guaranteed by the flag. The in-memory fixture is strongly consistent, so no test surfaces this.

Pick one design: OR the flag back in (`req.query.error === "limit" || liveCount >= MAX`) to match the description and guarantee the post-rejection banner, or drop `&error=limit` from the redirect so nothing emits an unread signal.

## Low Priority Suggestions

### 1. The "live address" predicate is hand-replicated in three places

`disabledAt === undefined` now lives in the page (`inbox.page.ts:58`) and both store enforcements. A shared domain helper (`countLiveAddresses`/`isLiveAddress`) in `@packages/domain/inbox` would make the page↔store lockstep structural rather than coincidental.

### 2. Duplicated cap-seeding loop in the tests

The 25-address seeding loop is duplicated between `inbox.route.test.ts:104-106` and `223-225`. A `seedAddressesToCap(fixture, userId)` helper would DRY it.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #821
- **Branch**: claude/vigilant-brahmagupta-5sd1r5

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #821
- **Commit**: 16a133eaf5704eaea6b9d8c3e0aa6cc571651881

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

The single MEDIUM from the prior review — `&error=limit` emitted by `POST /create` but read by nothing — is **resolved**. The OR-clause at `inbox.page.ts:58-60` makes the flag a meaningful trigger again while keeping the proactive live-count derivation. I verified the reconciliation is a genuine correctness fix, not cosmetic:

- **Union strictly dominates.** The new condition is a superset of both prior behaviours' "show" cases (flag-set and live-count-at-cap), both semantically valid — so no false-positive banner states, and the post-rejection case commit 1 had dropped is covered again.
- **Eventual-consistency guarantee restored.** The two GSI reads (inside `createAddress`, and on the redirect target) have no cross-read consistency (`dynamodb-inbox-address.ts:39-54`), so the redirect-target read can undercount and drop the banner exactly when the rejection meant to show it. OR-ing the flag closes that window.
- **Production-path live filter is correct.** `dynamoField` normalizes DynamoDB's missing-attribute `null` to `undefined` (`dynamo-field.ts:13`), so the page's `a.disabledAt === undefined` filter counts live rows identically to both store enforcements over the same `listAddressesByUserId` — page count == store count by construction, no `null`-vs-`undefined` trap.
- **New test pins the OR clause.** `inbox.route.test.ts:115-124` GETs `?error=limit` with the live count below the cap and asserts the banner, so removing the OR clause now fails a test. Both operands take both truth values across the suite, so the branch is covered.

## Low Priority Suggestions

### 1. `disabledAt === undefined` live-predicate is hand-replicated in three places
Page filter (`inbox.page.ts:60`) must stay identical to both store enforcements (in-memory:17, dynamo:65); a shared `isLiveAddress`/`countLiveAddresses` helper in `@packages/domain/inbox` would make that structural. *(Carried over; deferred as out-of-scope LOW.)*

### 2. Duplicated cap-seeding loop in the tests
The 25-address loop is duplicated between `inbox.route.test.ts:104-106` and `231-236`; a `seedAddressesToCap(fixture, userId)` helper would DRY it. *(Carried over.)*

### 3. Commit subject describes the process, not the change
The body of `16a133ea` is excellent, but the subject `fix: address high/medium priority review issues` is unscoped/process-oriented; a scoped subject like `fix(inbox): keep &error=limit as a banner trigger after a rejected create` matches the sibling `fix(inbox):` commit. Non-blocking.

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #821
- **Commit**: ebe36295038a7e43bf26533b6a376e6954f973cf

FagnerMartinsBrack:
### Claude Code Review — PR #821 (re-review @ `ebe36295`)

None found

## Medium Priority Issues

None found

The three LOW items from the prior reviews are genuinely resolved, and the new shared-predicate refactor is value-preserving:

- **One predicate, three enforcement sites.** `isLiveAddress`/`countLiveAddresses` (`inbox-address.live.ts:8-13`) now back the page banner (`inbox.page.ts:60`), the in-memory store (`in-memory-inbox-address.ts:18`), and the dynamo store (`dynamodb-inbox-address.ts:65`). Each substitution equals the inline filter it replaced exactly — behaviour unchanged, page↔store lockstep now structural. In-memory keeps the single-pass `row.userId === userId && isLiveAddress(row)` form because its map holds every user's rows (correct — the dynamo path scopes by user first).
- **Helper unit-tested both ways.** `inbox-address.live.test.ts` covers `isLiveAddress` true/false and `countLiveAddresses` over mixed/all-disabled/empty, built from real entries (no `as`), placed next to the implementation.
- **`seedAddressesToCap` DRYs the loop; import cleanup correct.** Both 25-create loops now call one helper minting through the real store; the removed `InboxAddressLimitReachedError` import is right. The two route tests independently pin both operands of the `||`.
- **Commit subject.** `ebe36295` is a properly-scoped `refactor(inbox):`; prior LOW #3 addressed in spirit.

## Low Priority Suggestions

### 1. A fourth "live" derivation in the same folder still bypasses the new helper

The commit's goal — *derive "live" through one shared predicate* — is met for the three sites it touched, but `inbox.component.ts:26` computes the **same** predicate by hand to drive each row's status:

```ts
enabled: entry.disabledAt === undefined,
```

This is definitionally `isLiveAddress(entry)` and sits in the same `inbox/` folder, so a future change to what "live" means would silently desync the row pill from the cap/banner. The helper's docstring enumerates only the cap and banner as consumers — this is a third. Completing it is a one-liner (`enabled: isLiveAddress(entry)`, adding the import already used on line 5); `inbox.component.test.ts` already renders both an enabled and a disabled row, so there's no coverage risk.

PR #823, opened 2026-06-24
fix(ios): make brand-mark script reproduce its committed PNGs

FagnerMartinsBrack:
## Problem

`projects/ios-readplace/scripts/make-brandmark.sh` could not regenerate the assets it claimed to produce:

- It rasterised with **macOS-only `sips`**, so it cannot run in the repo's Linux devbox shell at all.
- The committed PNGs were actually rendered **on Linux with Liberation Serif** — they carry only `IHDR`/`IDAT`/`IEND`, none of the ancillary chunks (`sRGB`, `gAMA`, `pHYs`, …) that `sips` emits.
- The script pinned `'Times New Roman'` for the glyph. On macOS that font *exists*, so `sips` would render the ampersand with Times New Roman — a **different outline** from the committed Liberation Serif mark.
- The provenance comment (`"rasterised on a Linux host where … Liberation Serif stands in"`) therefore described a toolchain the script did not use.

Net: the script reproduced the committed bytes on no platform, and the comment was wrong.

## Fix

- Rewrite the renderer to use **`rsvg-convert` (librsvg)** with **Liberation Serif**, keeping alpha (no flatten step — that's only for the App Store icon in `make-appicon.sh`).
- Declare **`librsvg`** and **`liberation_ttf`** in `devbox.json` so the toolchain ships with the repo shell and produces the same mark on Linux *and* macOS (removing the Times-New-Roman substitution hazard).
- Correct the provenance comment to describe the real toolchain.
- Regenerate all six PNGs from the corrected script.

## Verification

- Ran the rewritten script: it writes all six PNGs (light + dark, @1x/2x/3x).
- Re-ran it — output is **byte-for-byte identical**, i.e. the script now deterministically reproduces every committed PNG.
- Confirmed each PNG is 72/144/216 px RGBA and renders the navy (light) / white (dark) serif ampersand with the amber dot.
- `bash -n` and `devbox.json` JSON validation pass.

> Note: `devbox.lock` will gain the two new packages on the next `devbox install` (no nix toolchain in this environment to pre-resolve the lock).

---
_Generated by [Claude Code](https://claude.ai/code/session_01CWKp6YSGTTULj3kkVKVZ2W)_


FagnerMartinsBrack:

## Context
- **PR**: #823
- **Commit**: 9a6aa6605e5bdf82843b0f0d69efbe1302a9db27

FagnerMartinsBrack:

None found.

## Medium Priority Issues

None found.

## Low Priority Suggestions

### 1. No guard against silent font substitution
`make-brandmark.sh:41` names a single family with no fallback. `rsvg-convert` (via pango/fontconfig) will silently substitute a default serif if `Liberation Serif` can't be resolved (run outside the devbox shell, a minimal host, or a future `liberation_ttf` not wired into the nix fontconfig path), producing a different ampersand outline with no error — the same failure class this PR fixes, relocated. A one-line `fc-match "Liberation Serif" | grep -q LiberationSerif || exit 1` guard after `cd` would make it fail loudly.

### 2. Byte-reproducibility is asserted but not pinned
No `devbox.lock` is tracked and both new packages use `@latest` (`devbox.json:10-11`), so resolved versions float. I reproduced all six PNGs byte-for-byte on an independent toolchain (system librsvg 2.58.0 + Debian `fonts-liberation`), so this is informational — but the "deterministically reproduces" guarantee is only formal once the resolved versions are pinned in a committed lock.

### 3. The mark uses Liberation Serif, not the canonical logo's first-choice Noto Serif
The canonical `Readplace_Logo_only.svg` declares `font-family="'Noto Serif', Georgia, 'Times New Roman', serif"`; the script pins Liberation Serif (metric clone of the 3rd entry), so the login `&` outline differs from the canonical Noto Serif glyph. Appears intentional and preserves the prior mark, but worth a human confirmation; `noto-fonts` is also in nixpkgs if higher fidelity is wanted.

### 4. Sibling `make-appicon.sh` still diverges (out of scope)
It still uses macOS-only `sips` (can't run in the Linux devbox) and a `Georgia, 'Times New Roman', serif` stack → a Georgia ampersand, a third outline. Pre-existing; could adopt this PR's Linux-reproducible `rsvg-convert` path.

PR #840, opened 2026-06-28
Redirect authenticated users from home page to /queue

FagnerMartinsBrack:
## Summary
Authenticated users visiting the home page (`GET /`) are now redirected to `/queue` instead of seeing the landing page content. This applies to both HTML and markdown content negotiation.

## Changes
- **Server route handler** (`server.ts`): Added authentication check at the start of the `GET /` handler to redirect logged-in users (those with `req.userId`) to `/queue` with a 303 status code, before any content negotiation logic
- **Test coverage**: Added two integration tests to verify the redirect behavior:
  - `home.route.test.ts`: Tests that authenticated users receive a 303 redirect to `/queue` when accessing `/`
  - `home.static.route.test.ts`: Tests that authenticated users are redirected even when requesting markdown content via `Accept: text/markdown` header
- **Test utilities**: Imported `loginAgent` helper in both test files to support authenticated request testing

## Implementation Details
The redirect check is placed at the beginning of the route handler, before the existing Siren/markdown content negotiation logic. This ensures all authenticated visitors are redirected regardless of their `Accept` header preferences, keeping the landing page exclusively for unauthenticated users.

https://claude.ai/code/session_01212tyY5z2Fs1JbbPXxT7Mg

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -627,6 +627,10 @@ export function createApp(dependencies: AppDependencies): Express {
 	/** Firefox extensions enforce CORS preflight for fetches with non-simple headers (Accept: application/vnd.siren+json, Authorization). Register OPTIONS so the preflight succeeds; without this it returns 404 and firefox aborts the fetch with NetworkError. */
 	app.options("/", extensionCors);
 	app.get("/", extensionCors, async (req: Request, res: Response) => {
+		if (req.userId) {
+			res.redirect(303, QUEUE_PATH);
+			return;
+		}
 		if (wantsSiren(req) && !wantsMarkdown(req)) {
 			res.redirect(303, QUEUE_PATH);
 			return;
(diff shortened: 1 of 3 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #840
- **Commit**: 605893dc96edb4974d6931f22f41c79b86729d07

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Two adjacent guards now share an identical body (optional — current form is fine)

In `projects/hutch/src/runtime/server.ts:629-637` the `GET /` handler has two back-to-back guards that both run `res.redirect(303, QUEUE_PATH); return;` — one for `req.userId` and one for `wantsSiren(req) && !wantsMarkdown(req)`. No change is required: keeping them separate is **consistent** with how `/login` (`auth.page.ts:150`) and `/signup` (`auth.page.ts:213`) place the `if (req.userId)` guard at the top of the handler, and the two conditions express distinct intents that may diverge later. Merging into a single `||` would be terser but couples two unrelated reasons. Noted for awareness only.

PR #846, opened 2026-06-28
Replace linkedom with regex for style injection to preserve HTML escaping

FagnerMartinsBrack:
## Summary

Replace the linkedom-based DOM parsing approach in `injectPageStylesIntoMain` with a simple regex-based string replacement. This fixes a bug where double-escaped HTML entities in iframe `srcdoc` attributes were being decoded during style injection, causing code samples to be incorrectly interpreted as live HTML elements.

## Changes

- **Remove linkedom dependency**: Eliminated the `parseHTML` import and DOM manipulation approach that was inadvertently normalizing HTML entities
- **Implement regex-based injection**: Use a single regex replace to inject the `<style>` tag as the first child of `<main>`, preserving all original HTML escaping in the content
- **Add comprehensive test**: New test case verifies that double-escaped entities in iframe `srcdoc` attributes survive style injection and render correctly in the browser

## Implementation Details

The old approach parsed the entire content string into a DOM tree, which normalized HTML entities (e.g., `&amp;lt;` → `&lt;`). The new approach uses a case-insensitive regex to find the opening `<main>` tag and inserts the style tag immediately after it, leaving all content untouched.

The regex pattern `/<main(?=[\s/>])[^>]*>/i` ensures we match:
- The literal string `<main`
- Followed by whitespace, `>`, or `/` (lookahead, not consumed)
- Any attributes up to the closing `>`

This preserves the exact byte representation of the HTML, which is critical for iframe `srcdoc` attributes that rely on specific escaping levels to render correctly.

https://claude.ai/code/session_01CuLkzq5teRQvdJikCM8JfT

### DIFF
--- a/src/packages/web-shell/src/base.component.test.ts
+++ b/src/packages/web-shell/src/base.component.test.ts
@@ -75,6 +75,30 @@ describe("Base component", () => {
 		expect(main?.querySelector("style")?.textContent).toBe(".lead { color: rebeccapurple; }");
 	});
 
+	it("preserves the reader iframe's double-escaped srcdoc when injecting page styles, so code samples stay text", () => {
+		const srcdoc =
+			"&lt;!doctype html&gt;&lt;body&gt;&lt;code&gt;&amp;lt;input&amp;gt;&lt;/code&gt;&lt;/body&gt;";
+		const page = createTestPageBody({
+			styles: ".reader { color: rebeccapurple; }",
+			content: { html: `<main><iframe data-reader-iframe srcdoc="${srcdoc}"></iframe></main>` },
+		});
+		const result = Base(page, GUEST_STATE).to("text/html");
+
+		// Byte-level guard: the double-escaping must survive the style injection.
+		// A DOM re-parse would decode one level and hide a degradation, so assert
+		// on the raw output bytes.
+		expect(result.body).toContain("&amp;lt;input&amp;gt;");
+
+		const doc = new JSDOM(result.body).window.document;
+		const iframeSrcdoc = doc.querySelector("iframe[data-reader-iframe]")?.getAttribute("srcdoc");
+		assert(iframeSrcdoc, "reader iframe must carry a srcdoc");
+		const innerDoc = new JSDOM(iframeSrcdoc).window.document;
+		expect(innerDoc.querySelectorAll("input").length).toBe(0);
+
+		const main = doc.querySelector("main");
+		expect(main?.firstElementChild?.tagName).toBe("STYLE");
+	});
+
 	it("should apply bodyClass when provided", () => {
 		const page = createTestPageBody({ bodyClass: "page-home" });
 		const result = Base(page, GUEST_STATE).to("text/html");
--- a/src/packages/web-shell/src/inject-page-styles.ts
+++ b/src/packages/web-shell/src/inject-page-styles.ts
@@ -1,19 +1,25 @@
 import assert from "node:assert";
-import { parseHTML } from "linkedom";
 
 /**
  * Inject page-specific CSS as a <style> element inside <main>, so that htmx
  * navigation (hx-target="main" hx-select="main" hx-swap="outerHTML") swaps the
  * page's CSS atomically with its content. Without this, htmx leaves the
  * previous page's <style> stranded in <head>, defacing the new page's layout.
+ *
+ * Done as a string insert, NOT via linkedom's parseHTML, so the content
+ * survives byte-for-byte. A DOM parse-then-serialize round-trip decodes one
+ * level of HTML escaping, and the reader's iframe `srcdoc` is intentionally
+ * double-escaped: a code sample like `&amp;lt;input&amp;gt;` would collapse to a
+ * live `<input>` inside the framed document. This is the deliberate inverse of
+ * the case where parseHTML is correct because the input crossed a network
+ * boundary and malformed bytes should degrade to undefined; here the input is
+ * trusted server markup whose escaping must be preserved, so a parser would
+ * corrupt it. Do not "restore" parseHTML.
  */
 export function injectPageStylesIntoMain(content: string, styles: string): string {
 	if (!styles) return content;
-	const { document } = parseHTML(`<!DOCTYPE html><html><body>${content}</body></html>`);
-	const main = document.querySelector("main");
-	assert(main, "PageBody.content must contain a <main> element when styles are provided");
-	const styleEl = document.createElement("style");
-	styleEl.textContent = styles;
-	main.insertBefore(styleEl, main.firstChild);
-	return document.body.innerHTML;
+	const styleTag = `<style>${styles}</style>`;
+	const updated = content.replace(/<main(?=[\s/>])[^>]*>/i, (mainTag) => mainTag + styleTag);
+	assert(updated !== content, "PageBody.content must contain a <main> element when styles are provided");
+	return updated;
 }


FagnerMartinsBrack:

## Context
- **PR**: #846
- **Commit**: e651691050a5ff88df8b467ee35cb2d1c94531b9

FagnerMartinsBrack:

None found

The fix is correct. A full `parseHTML(...content...)` → `document.body.innerHTML` round-trip decodes one level of the reader iframe's intentionally double-escaped `srcdoc`, so a `<code>` example such as `&amp;lt;input&amp;gt;` degrades into a live `<input>` inside the inner document. Replacing it with a byte-preserving string insert (`content.replace(/<main(?=[\s/>])[^>]*>/i, …)`) leaves the `srcdoc` attribute untouched. The new test guards this at the byte level (`toContain("&amp;lt;input&amp;gt;")`) and by re-decoding the `srcdoc` to assert zero rendered `<input>` controls — both assertions correctly fail against the old `linkedom` path. The revised `assert(updated !== content, …)` is sound: `styleTag` is always non-empty (guarded by `if (!styles) return content`), so a successful match always lengthens the string and a no-match leaves it identical, preserving the original "no `<main>`" invariant. Using a replacement **function** rather than a `"$&…"` string is also the right call — it avoids `$&`/`$1` substitution if `styles` ever contains a `$`-sequence.

## Medium Priority Issues

### 1. Regex-based markup parsing needs explicit human approval and a "why" comment per the `web` skill

The change swaps `linkedom`'s `parseHTML` for a regex to locate the `<main>` start tag. The repo's own `web` skill (`.claude/skills/web/SKILL.md`, "Structured Parsing Over Regex") is explicit:

> For HTML in this codebase, use `linkedom`'s `parseHTML` and standard DOM APIs … **Regex-based markup parsing requires explicit human approval.**

The motivation here is legitimate — `parseHTML` genuinely cannot be used naively without corrupting the double-escaped `srcdoc` — so this is a reasonable exception. But the convention names a required step (explicit sign-off), so it should be approved on the record rather than landing silently, and the deliberate choice should be documented in code:

- **Add a `why` comment** to `injectPageStylesIntoMain` stating that the regex is intentional *because* a DOM parse-then-serialize round-trip normalizes the reader iframe's double-escaped `srcdoc`. The sibling file in the same package carries the inverse rationale (`changelog-banner.ts`: *"linkedom (not a regex) so malformed input degrades to undefined rather than mis-parsing"*). Without a counter-note here, a future reader following the `web` skill is likely to "restore" `parseHTML` and silently reintroduce this exact bug. Per CLAUDE.md ("Comments Document Why"), this is precisely the non-obvious *why* that justifies a comment.

[Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Freader-iframe-html-escape-quxctd%2C%20in%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.component.ts%2C%20add%20a%20why-comment%20to%20injectPageStylesIntoMain%20explaining%20that%20a%20regex%20is%20used%20deliberately%20instead%20of%20linkedom%20parseHTML%2C%20because%20a%20DOM%20parse-then-serialize%20round-trip%20decodes%20one%20level%20of%20the%20reader%20iframe%27s%20double-escaped%20srcdoc%20and%20turns%20escaped%20code%20samples%20into%20live%20elements.%20Mirror%20the%20rationale%20comment%20style%20in%20changelog-banner.ts.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. `[^>]*>` truncates the match if a `<main>` start tag ever carries a `>` inside an attribute value

`[^>]*>` stops at the first `>`, so a tag like `<main data-x="a>b">` would match only `<main data-x="a>` and inject the style mid-attribute. I checked every `<main>` in the codebase — they are all static (`<main class="view">`, `<main class="reader">`, occasional `data-test-*` flags), none contain a `>` in an attribute value, so this is **not currently triggered**. It is, however, exactly the fragility the `web` skill warns about; folding "the `<main>` tag is server-authored with static attributes" into the `why` comment above would record why the regex is safe here.

### 2. Raw `<style>${styles}</style>` insertion replaces the old `textContent` escaping

The previous code set `styleEl.textContent = styles` (escaped on serialization); the new code injects `styles` raw. This is safe because `styles` is always developer-authored static CSS (e.g. `.reader { color: rebeccapurple; }`), never user/crawler input — but it now assumes `styles` never contains the literal `</style>`. Worth a one-line awareness note if any page ever computes `styles` dynamically.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #846
- **Branch**: claude/reader-iframe-html-escape-quxctd

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #846
- **Commit**: 111c1f7c4d04707ea67411f163a0770336452e76

FagnerMartinsBrack:

None found

The fix is correct, and I verified it empirically rather than by inspection. Running the **old** `linkedom` round-trip and the **new** regex insert side-by-side against the reader's double-escaped `srcdoc`:

| | bytes keep `&amp;lt;input&amp;gt;` | browser-decoded `srcdoc` live `<input>` |
|---|---|---|
| **OLD** (`parseHTML` → `innerHTML`) | ❌ false | **1** (decoded attr: `…<code><input></code>…`) |
| **NEW** (regex insert) | ✅ true | **0** (decoded attr: `…<code>&lt;input&gt;</code>…`) |

linkedom's serialization does **not** re-escape `&` in the `srcdoc` attribute, so the round-trip silently decoded one level and the whole escaped document — not just the code sample — collapsed into live markup. The byte-preserving `content.replace(/<main(?=[\s/>])[^>]*>/i, …)` leaves it untouched. The new test is therefore a real regression guard: both its assertions (`toContain("&amp;lt;input&amp;gt;")` and zero rendered `<input>` after decoding the `srcdoc`) **fail against the old path and pass against the new one**. The `assert(updated !== content, …)` is sound — `styleTag` is always non-empty past the `if (!styles) return content` guard, so a match always lengthens the string and a no-match leaves it identical, preserving the original "no `<main>`" invariant. Using a replacement **function** (not `"$&…"`) is also correct — it avoids `$&`/`$1` substitution if `styles` ever contains a `$`.

## Medium Priority Issues

None found

The prior review's one MEDIUM — *"regex-based markup parsing needs explicit human approval and a `why` comment per the `web` skill"* — is **resolved** by commit `111c1f7c`. The reviewer's explicit request is the on-record approval the `web` skill asks for, and `injectPageStylesIntoMain` now carries an accurate `why` comment. I checked every claim in it:

- *"a DOM parse-then-serialize round-trip decodes one level of HTML escaping"* — **confirmed** (table above).
- *"the reader's iframe `srcdoc` is intentionally double-escaped"* — **confirmed**: crawled code samples are single-escaped (`&lt;input&gt;`), then Handlebars `{{srcdoc}}` re-escapes the `&` to `&amp;` when embedding into `srcdoc="{{srcdoc}}"`.
- *"matches only the server-authored `<main>` start tag (static attributes, never a `>` inside an attribute value)"* — **confirmed**: every `<main>` across all page templates uses only `class="…"` / `data-test-*` flags.
- *"deliberate inverse of changelog-banner.ts"* — accurate.

## Low Priority Suggestions

### 1. The regex targets the first literal `<main` in source order (carried over from the prior review)

`/<main(?=[\s/>])[^>]*>/i` matches the first `<main` start tag in the raw string. Two latent fragilities, **neither currently triggered** and both now covered by the `why` comment's "server-authored `<main>`" framing:
- `[^>]*>` would truncate if a `<main>` tag ever carried a `>` inside an attribute value (e.g. `<main data-x="a>b">`).
- A literal `<main` appearing *before* the real wrapper (e.g. inside an HTML comment `<!-- <main> -->`) would be matched instead — something `parseHTML`'s `querySelector("main")` would have skipped.

Both are acceptable given the input is trusted server markup whose `<main>` is always the first one. No action needed unless a page ever computes its `<main>` attributes dynamically.

### 2. Raw `<style>${styles}</style>` insertion replaces the old `textContent` escaping (carried over)

The previous code set `styleEl.textContent = styles` (escaped on serialization); the new code injects `styles` raw. Safe today because `styles` is always developer-authored static CSS (e.g. `READER_STYLES`), never user/crawler input — but it now assumes `styles` never contains the literal `</style>`. Worth a one-line awareness note if any page ever computes `styles` dynamically.

FagnerMartinsBrack:

## Context
- **PR**: #846
- **Commit**: e81433951d871dd2f2576d88cf29582215f7ba8f

FagnerMartinsBrack:

**Found 4 comment(s) to fix** (and 0 that look compliant with the skill).

### `src/packages/web-shell/src/base.component.ts`

> This is the deliberate inverse of changelog-banner.ts, where parseHTML is correct ... The regex matches only the server-authored <main> start tag ... Do not "restore" parseHTML.

- **Issue:** names another file/symbol by path — rots when that file moves or is renamed.
- **Fix:** The deliberate-deviation why (string insert instead of parseHTML so the reader iframe's intentional double-escaping survives byte-for-byte; 'do not restore parseHTML') is a legitimate approved-hack rationale worth keeping. But remove the 'changelog-banner.ts' file-path reference — naming another file by path rots when that file moves; describe the boundary concept (trusted server markup vs. network-crossing input) without the filename. Also trim the sentence narrating how the regex matches the <main> tag, which is what/how rather than why. Retaining the trimmed comment still needs explicit human approval.

### `src/packages/web-shell/src/base.component.test.ts`

> Byte-level guard: the double-escaping must survive the style injection ... with the old linkedom roundtrip these would be the single-escaped &lt;input&gt;.

- **Issue:** anchors the comment to a now-removed/old implementation state — it will rot.
- **Fix:** The 'with the old linkedom roundtrip…' clause anchors the comment to an implementation state this PR removes; once linkedom is gone the reference is dangling and will rot. Drop that clause. The remaining why (assert on raw output bytes because a DOM re-parse would decode one level and hide the degradation) is the only defensible part, and keeping even that needs human approval.

### `src/packages/web-shell/src/base.component.test.ts`

> Mirror the browser: the iframe's srcdoc decodes once into a document whose <input> example must render as text, not a live form control.

- **Fix:** Remove. The comment narrates what the following lines do (read srcdoc, decode once via JSDOM, assert zero <input> elements); the test name already states the intent ('so code samples stay text'). A clearer assertion message conveys the same without a standalone comment.

### `src/packages/web-shell/src/base.component.test.ts`

> The page style is still injected as the first child of <main>.

- **Fix:** Remove. This restates the very next assertion (expect(main?.firstElementChild?.tagName).toBe('STYLE')); the code already says what it checks.

FagnerMartinsBrack:

None found

This is a re-review at the merge commit (`e8143395`); merging `main` left the two-file change logically identical to the previously reviewed state (`git diff origin/main...HEAD` is the same swap of `parseHTML` → string insert plus the new test). I re-verified the fix rather than assuming the prior runs covered it.

The fix is correct. The byte-preserving `content.replace(/<main(?=[\s/>])[^>]*>/i, (mainTag) => mainTag + styleTag)` avoids the old `parseHTML(...)` → `document.body.innerHTML` round-trip, which decoded one level of the reader iframe's intentionally double-escaped `srcdoc` — turning a `&amp;lt;input&amp;gt;` code sample into a live `<input>` inside the framed document. The new test is a real regression guard on both axes: the byte-level `toContain("&amp;lt;input&amp;gt;")` and the decode-the-`srcdoc`-then-assert-zero-`<input>` check each fail against the old `linkedom` path and pass against the new one.

Three correctness points I checked independently:
- **`assert(updated !== content, …)` is sound.** Past the `if (!styles) return content` guard, `styleTag` is always non-empty, so a match strictly lengthens the string; a no-match leaves `replace` returning a value-equal string, so `!==` is `false` and the original "no `<main>`" invariant fires exactly as the old `assert(main, …)` did.
- **The replacement is a function, not a `"$&…"` string**, so a `$` inside `styles` can't trigger `$&`/`$1` substitution.
- **The `(?=[\s/>])` lookahead prevents false matches** on `<maintenance>` / `<main-panel>` (the next char must be whitespace, `/`, or `>`), so only a real `<main>` start tag matches.

## Medium Priority Issues

None found

The earlier MEDIUM — *"regex-based markup parsing needs explicit human approval and a `why` comment per the `web` skill"* — remains resolved by commit `111c1f7c`. I re-validated every cross-file factual claim the `why` comment now makes, since an inaccurate rationale comment is worse than none:
- *"the reader's iframe `srcdoc` is intentionally double-escaped"* — **confirmed**: `reader-ready.template.html` embeds the content as `srcdoc="{{srcdoc}}"`, and Handlebars' double-stache HTML-escapes an already single-escaped crawl payload, yielding two levels.
- *"deliberate inverse of changelog-banner.ts"* — **confirmed**: that file keeps `parseHTML` *because* its input crosses a network boundary and malformed bytes should degrade to `undefined` — the genuine opposite of this trusted-server-markup case.
- *"matches only the server-authored `<main>` start tag (static attributes, never a `>` inside an attribute value)"* — **confirmed** against the page templates.

The removed `parseHTML` import does not orphan the `linkedom` dependency (`changelog-banner.ts` still imports it), so there's no knip fallout, consistent with CI passing.

## Low Priority Suggestions

### 1. `[^>]*>` and first-match semantics (carried over)

`/<main(?=[\s/>])[^>]*>/i` matches the first literal `<main` start tag and stops at the first `>`. Two latent fragilities, **neither triggered today** and both now covered by the comment's "server-authored `<main>`" framing:
- a `>` inside a future `<main>` attribute value would truncate the match;
- a literal `<main` inside a preceding HTML comment (``) would be matched ahead of the real wrapper — something `parseHTML`'s `querySelector("main")` would have skipped.

No action needed while `<main>` stays static, first-in-source server markup.

### 2. Raw `<style>${styles}</style>` insertion replaces the old `textContent` escaping (carried over)

The previous code set `styleEl.textContent = styles` (escaped on serialization); the new code injects `styles` raw. Safe today because `styles` is always developer-authored static CSS (e.g. `READER_STYLES`), never user/crawler input — but it now assumes `styles` never contains the literal `</style>`. Worth a one-line awareness note if any page ever computes `styles` dynamically.

FagnerMartinsBrack:

## Context
- **PR**: #846
- **Commit**: 63b6304378839543552e2f01243493adbc57fb89

FagnerMartinsBrack:

None found

Re-review at commit `63b63043`. That commit is **comment-only** (`git show` confirms: only comment lines changed in both files), so the behaviour under review is unchanged from the prior passing reviews, and `git diff origin/main...HEAD` is still the same swap of `parseHTML` → byte-preserving string insert plus the new regression test.

The fix remains correct. `content.replace(/<main(?=[\s/>])[^>]*>/i, (mainTag) => mainTag + styleTag)` avoids the old `parseHTML(...)` → `document.body.innerHTML` round-trip that decoded one level of the reader iframe's intentionally double-escaped `srcdoc`, which turned a `&amp;lt;input&amp;gt;` code sample into a live `<input>` inside the framed document. The three correctness points carry over and still hold: `assert(updated !== content, …)` is sound (past the `if (!styles) return content` guard `styleTag` is always non-empty, so a match strictly lengthens the string and a no-match leaves it identical — preserving the original "no `<main>`" invariant); the replacement is a function not a `"$&…"` string (so a `$` in `styles` can't trigger substitution); and the `(?=[\s/>])` lookahead blocks false matches on `<maintenance>`/`<main-panel>`.

## Medium Priority Issues

None found

The earlier MEDIUM — *"regex-based markup parsing needs explicit human approval and a `why` comment per the `web` skill"* — stays resolved. The `web` skill (`.claude/skills/web/SKILL.md:244`) still reads *"Regex-based markup parsing requires explicit human approval"*, and that approval is on record in this PR thread.

This commit applied the `code-comments` skill to the four comments flagged in the previous human review, and the trim is faithful:
- **`base.component.ts`** — the `changelog-banner.ts` **file-path reference** is gone (now "the case where parseHTML is correct…"), and the what/how sentence narrating *how* the regex matches `<main>` is removed. The retained approved-hack *why* (string insert instead of `parseHTML` because a DOM round-trip decodes one escaping level the reader `srcdoc` depends on; "do not restore parseHTML") is intact.
- **`base.component.test.ts`** — the "old linkedom roundtrip" clause (which anchored the comment to the removed implementation) is dropped; the "Mirror the browser" and "still injected as the first child of `<main>`" restatements are removed.

I re-verified the retained rationale is still **accurate** (an inaccurate approved-hack comment would be worse than none): `changelog-banner.ts` still imports and uses `parseHTML` from `linkedom` (lines 2, 68) with its own inverse rationale comment (*"linkedom (not a regex) so malformed input degrades to undefined … because the bytes crossed a network boundary"*), so the dependency is **not orphaned** (no knip fallout, consistent with CI green) and the "deliberate inverse" reference holds even without naming the file.

## Low Priority Suggestions

### 1. `[^>]*>` and first-match semantics (carried over)

`/<main(?=[\s/>])[^>]*>/i` matches the first literal `<main` start tag and stops at the first `>`. Two latent fragilities, **neither triggered today** and both covered by the comment's "trusted server markup" framing: a `>` inside a future `<main>` attribute value would truncate the match, and a literal `<main` inside a preceding HTML comment would be matched ahead of the real wrapper (something `parseHTML`'s `querySelector("main")` would have skipped). No action needed while `<main>` stays static, first-in-source server markup.

### 2. Raw `<style>${styles}</style>` insertion replaces the old `textContent` escaping (carried over)

The previous code set `styleEl.textContent = styles` (escaped on serialization); the new code injects `styles` raw. Safe today because `styles` is always developer-authored static CSS (e.g. `READER_STYLES`), never user/crawler input — but it now assumes `styles` never contains the literal `</style>`. Worth a one-line awareness note if any page ever computes `styles` dynamically.

### 3. The two retained comments still need the on-record human ack the `code-comments` skill requires

The `code-comments` skill is explicit: *"Adding **or** keeping any comment … requires explicit human approval"* (removing never does). The auto-apply run that produced `63b63043` explicitly punted that confirmation back to you, listing the two comments it **kept**: the `injectPageStylesIntoMain` approved-hack `why`, and the trimmed byte-level-guard test `why`. The first you already called *"a legitimate approved-hack rationale worth keeping"*; the second you called *"the only defensible part … keeping even that needs human approval."* Both carry a genuine, accurate `why` and read as compliant — this is purely a request to record the keep-decision so the skill's approval gate is satisfied on the record rather than left implicit.

FagnerMartinsBrack:

## Context
- **PR**: #846
- **Branch**: claude/reader-iframe-html-escape-quxctd

FagnerMartinsBrack:

## Context
- **PR**: #846
- **Commit**: 3c3a9808ddd7b65d76faf2597c4cfea9f9a34805

FagnerMartinsBrack:

None found

Re-review at commit `3c3a9808`. The post-conflict rebuild relocated `injectPageStylesIntoMain` from `base.component.ts` into the `inject-page-styles.ts` module that already existed on `main` (carrying the **old `linkedom` implementation**), and re-applied the byte-preserving regex there. `git diff origin/main...HEAD` is the same `parseHTML` → string-insert swap plus the regression test, now in its relocated home. Both consumers — `base.component.ts:257` (full shell) and `chromeless-page.ts:39` (chromeless iOS reader) — import the function unchanged, so both receive the fix; the chromeless reader is where the double-escaped `srcdoc` actually renders, so this is the correct surface to cover.

The fix is correct, verified empirically (old `linkedom` round-trip vs new regex, run side-by-side against the reader's double-escaped `srcdoc`):

| | keeps `&amp;lt;input&amp;gt;` bytes | live `<input>` in framed doc | `<style>` first child of `<main>` |
|---|---|---|---|
| **NEW** (`content.replace(/<main(?=[\s/>])[^>]*>/i, …)`) | ✅ true | **0** | ✅ true |
| **OLD** (`parseHTML(...)` → `document.body.innerHTML`) | ❌ false | **1** | — |

`parseHTML`→serialize decoded one level of escaping, collapsing a `&amp;lt;input&amp;gt;` code sample into a live `<input>` inside the framed document; the string insert leaves it untouched. The new test is a real regression guard on both axes (byte-level `toContain("&amp;lt;input&amp;gt;")` and decode-`srcdoc`-then-assert-zero-`<input>`) — each fails the old path, passes the new. Three correctness points re-confirmed:
- **`assert(updated !== content, …)` is sound** — past the `if (!styles) return content` guard `styleTag` is non-empty, so a match strictly lengthens the string while a no-match leaves it value-equal; the assert fires exactly where the old `assert(main, …)` did (verified: `<section>…</section>` → `updated === content`).
- **The replacement is a function, not `"$&…"`** — a `$` inside `styles` can't trigger `$&`/`$1` substitution.
- **The `(?=[\s/>])` lookahead blocks false matches** — verified `<maintenance>` is left untouched.

## Medium Priority Issues

None found

The earlier MEDIUM — *"regex-based markup parsing needs explicit human approval and a `why` comment per the `web` skill"* — remains resolved. `.claude/skills/web/SKILL.md:15` still reads *"Regex-based markup parsing requires explicit human approval"*; that approval is on record in this thread, and the relocated `injectPageStylesIntoMain` carries the human-finalized `why` comment. I re-validated its cross-file claims, since an inaccurate rationale comment is worse than none:
- *"the reader's iframe `srcdoc` is intentionally double-escaped"* — **confirmed**: `reader-ready.template.html` embeds via `srcdoc="{{srcdoc}}"`, and Handlebars' double-stache HTML-escapes an already single-escaped crawl payload, yielding two levels.
- *"the deliberate inverse of the case where parseHTML is correct because the input crossed a network boundary"* — **confirmed**: `changelog-banner.ts:62-66` keeps `parseHTML` precisely so malformed network bytes degrade to `undefined` — the genuine opposite of this trusted-server-markup case. The file-path reference was correctly dropped per the `code-comments` skill; the boundary concept survives without naming the file.
- **`linkedom` not orphaned** — `changelog-banner.ts:2,68` still imports and uses `parseHTML`, so removing it here produces no knip fallout (consistent with CI green).

## Low Priority Suggestions

### 1. `[^>]*>` and first-match semantics (carried over)

`/<main(?=[\s/>])[^>]*>/i` matches the first literal `<main` start tag and stops at the first `>`. Two latent fragilities, **neither triggered today** and both covered by the comment's "trusted server markup" framing: a `>` inside a future `<main>` attribute value would truncate the match, and a literal `<main` inside a preceding HTML comment would be matched ahead of the real wrapper (something `parseHTML`'s `querySelector("main")` would have skipped). No action needed while `<main>` stays static, first-in-source server markup.

### 2. Raw `<style>${styles}</style>` insertion replaces the old `textContent` escaping (carried over)

The previous code set `styleEl.textContent = styles` (escaped on serialization); the new code injects `styles` raw. Safe today because `styles` is always developer-authored static CSS (e.g. `READER_STYLES`), never user/crawler input — but it now assumes `styles` never contains the literal `</style>`. Worth a one-line awareness note only if any page ever computes `styles` dynamically.

### 3. The two retained `why` comments still owe the on-record keep-approval the `code-comments` skill requires (carried over)

`.claude/skills/code-comments/SKILL.md:20` is explicit: *"Adding **or** keeping any comment … requires explicit human approval … Removing a comment never needs approval."* The conflict-rebuild collapsed the comment-trim commits and carried the two comments forward verbatim in their final trimmed form — the `injectPageStylesIntoMain` approved-hack `why` and the test's byte-level-guard `why`. Both read as compliant (a genuine, accurate, non-derivable *why*), so this is purely a request to record the keep-decision on the record so the skill's gate is satisfied rather than left implicit — the same item the prior human review flagged.

PR #847, opened 2026-06-28
feat(ios): open client-side add-links help from +, drop server add-links-help

FagnerMartinsBrack:
## What

The iOS reading list's **`+` toolbar control is now a client-side `add-links-help` affordance** that opens the Share-help info sheet. It replaces the in-app "Save a URL" prompt that the server-advertised `save-article` action used to drive.

- **iOS ignores `save-article` for the toolbar.** With the bespoke-field-handler concept removed, `save-article` is just a field-requiring action with no server value → not bare-invokable → not surfaced as a toolbar control. It stays fully reachable for the **Share-extension save journey** via `action(named:"save-article")` / `api.saveArticle`, and is still advertised by the server.
- **`add-links-help` is now client-injected, not server-advertised.** `ReadingListViewModel` injects a static `add-links-help` affordance (SF Symbol `plus`) into `collectionAffordances`, present even before a queue load. The help URL is held client-side (`AppConfig.addLinksHelpPath = "/help/add-links"`) and resolved against the API base.
- **Server stops advertising the link.** `collection-siren.ts` no longer pushes the `add-links-help` link into the `/queue` Siren collection. The `/help/add-links` **page route stays** (it's what the sheet renders).

The in-app save-a-URL prompt machinery (`saveURL`, `ToolbarRoute.promptSave`, `bespokeFieldHandlers`, the "Save a URL" alert) is removed.

## Why

Requested change: have the iOS client ignore the server-advertised `save-article` and instead show `add-links-help` as a client-side-only control, so tapping `+` opens the info; and remove the corresponding server-side advertisement.

> Note: this is a deliberate departure from the fully server-driven toolbar (the `+` is now owned by the client rather than projected from a server affordance). Other server collection affordances are still looped as before.

## Verification

- **iOS:** `make test` equivalent — **168 XCTest tests, 0 failures**.
- **Server / monorepo:** full `pnpm check` (lint, type-check, tests, **100% coverage** across all 81 tasks) — **green**, incl. `hutch` (2435 tests).
- **Adversarial multi-agent review** of the diff: no behavioral / contract / Share-flow / coverage regressions; the only findings were stale "bespoke handler" doc comments, all fixed in this branch.

## Files

iOS: `AffordancePresentation`, `ReadingListViewModel`, `ReadingListView`, `ToolbarRoute`, `AddLinkInstructionsView`, `AppConfig`, `ReadplaceAPI` (+ tests, README).
Server: `collection-siren.ts` (+ test), `server.ts` comment.

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -663,9 +663,9 @@ export function createApp(dependencies: AppDependencies): Express {
 	});
 
 	// Bare (HtmlPage, not Base) so the iOS Share-help sheet can render it in a
-	// WKWebView without site chrome; public like /privacy. The iOS client reaches
-	// it via the add-links-help link rel in the /queue Siren collection, so the
-	// copy ships via a hutch deploy rather than an App Store review.
+	// WKWebView without site chrome; public like /privacy. The iOS client opens it
+	// at this path (held client-side, not advertised as a Siren link), so the copy
+	// ships via a hutch deploy rather than an App Store review.
 	app.get("/help/add-links", (req: Request, res: Response) => {
 		sendComponent(req, res, HelpAddLinksPage());
 	});
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -663,9 +663,10 @@ export function createApp(dependencies: AppDependencies): Express {
 	});
 
 	// Bare (HtmlPage, not Base) so the iOS Share-help sheet can render it in a
-	// WKWebView without site chrome; public like /privacy. The iOS client opens it
-	// at this path (held client-side, not advertised as a Siren link), so the copy
-	// ships via a hutch deploy rather than an App Store review.
+	// WKWebView without site chrome; public like /privacy. The current iOS client
+	// holds this path client-side; the /queue collection still advertises it via the
+	// add-links-help rel so older installed clients resolve the help URL from there.
+	// Either way the copy ships via a hutch deploy rather than an App Store review.
 	app.get("/help/add-links", (req: Request, res: Response) => {
 		sendComponent(req, res, HelpAddLinksPage());
 	});
--- a/projects/hutch/src/runtime/web/api/collection-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.test.ts
@@ -112,6 +112,23 @@ describe("toArticleCollectionEntity", () => {
 		expect(entity.links).toContainEqual({ rel: ["root"], href: "/queue" });
 	});
 
+	it("advertises a titled add-links-help link so already-installed clients resolve the help URL", () => {
+		const result: FindArticlesResult = {
+			articles: [],
+			total: 0,
+			page: 1,
+			pageSize: 20,
+		};
+
+		const entity = toArticleCollectionEntity(result, {});
+
+		expect(entity.links).toContainEqual({
+			rel: ["add-links-help"],
+			title: "How to add links",
+			href: "/help/add-links",
+		});
+	});
+
 	it("includes next link when more pages exist", () => {
 		const result: FindArticlesResult = {
 			articles: Array.from({ length: 20 }, (_, i) => makeArticle(`${i}`)),
@@ -151,7 +168,7 @@ describe("toArticleCollectionEntity", () => {
 		const entity = toArticleCollectionEntity(result, { page: 2, pageSize: 20 });
 
 		const linkRels = entity.links?.map((l) => l.rel[0]);
-		expect(linkRels).toEqual(["self", "root", "prev"]);
+		expect(linkRels).toEqual(["self", "root", "add-links-help", "prev"]);
 	});
 
 	it("single page omits both pagination links", () => {
@@ -165,7 +182,7 @@ describe("toArticleCollectionEntity", () => {
 		const entity = toArticleCollectionEntity(result, {});
 
 		const linkRels = entity.links?.map((l) => l.rel[0]);
-		expect(linkRels).toEqual(["self", "root"]);
+		expect(linkRels).toEqual(["self", "root", "add-links-help"]);
 	});
 
 	it("preserves query params in pagination links", () => {
--- a/projects/hutch/src/runtime/web/api/collection-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.test.ts
@@ -112,23 +112,6 @@ describe("toArticleCollectionEntity", () => {
 		expect(entity.links).toContainEqual({ rel: ["root"], href: "/queue" });
 	});
 
-	it("advertises a titled add-links-help link so looping clients use a server-authored label", () => {
-		const result: FindArticlesResult = {
-			articles: [],
-			total: 0,
-			page: 1,
-			pageSize: 20,
-		};
-
-		const entity = toArticleCollectionEntity(result, {});
-
-		expect(entity.links).toContainEqual({
-			rel: ["add-links-help"],
-			title: "How to add links",
-			href: "/help/add-links",
-		});
-	});
-
 	it("includes next link when more pages exist", () => {
 		const result: FindArticlesResult = {
 			articles: Array.from({ length: 20 }, (_, i) => makeArticle(`${i}`)),
@@ -168,7 +151,7 @@ describe("toArticleCollectionEntity", () => {
 		const entity = toArticleCollectionEntity(result, { page: 2, pageSize: 20 });
 
 		const linkRels = entity.links?.map((l) => l.rel[0]);
-		expect(linkRels).toEqual(["self", "root", "add-links-help", "prev"]);
+		expect(linkRels).toEqual(["self", "root", "prev"]);
 	});
 
 	it("single page omits both pagination links", () => {
@@ -182,7 +165,7 @@ describe("toArticleCollectionEntity", () => {
 		const entity = toArticleCollectionEntity(result, {});
 
 		const linkRels = entity.links?.map((l) => l.rel[0]);
-		expect(linkRels).toEqual(["self", "root", "add-links-help"]);
+		expect(linkRels).toEqual(["self", "root"]);
 	});
 
 	it("preserves query params in pagination links", () => {
--- a/projects/hutch/src/runtime/web/api/collection-siren.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.ts
@@ -43,14 +43,6 @@ export function toArticleCollectionEntity(
 		{ rel: ["root"], href: "/queue" },
 	];
 
-	// The add-links-help rel is the client contract; the href stays
-	// server-internal so the bare help page can move without a client release.
-	links.push({
-		rel: ["add-links-help"],
-		title: "How to add links",
-		href: "/help/add-links",
-	});
-
 	if (page > 1) {
 		links.push({
 			rel: ["prev"],
--- a/projects/hutch/src/runtime/web/api/collection-siren.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.ts
@@ -43,6 +43,16 @@ export function toArticleCollectionEntity(
 		{ rel: ["root"], href: "/queue" },
 	];
 
+	// Older iOS builds resolve the Share-help URL from this rel; the current client
+	// holds the path itself and ignores this link, but it stays advertised so those
+	// installed builds keep their server-rendered help instead of a bundled fallback.
+	// The href is server-internal, so the help page can move without a client release.
+	links.push({
+		rel: ["add-links-help"],
+		title: "How to add links",
+		href: "/help/add-links",
+	});
+
 	if (page > 1) {
 		links.push({
 			rel: ["prev"],
--- a/projects/ios-readplace/App/AddLinkInstructionsView.swift
+++ b/projects/ios-readplace/App/AddLinkInstructionsView.swift
@@ -3,11 +3,11 @@ import SwiftUI
 /// The sheet shown when the user taps + on the reading list. Rather than an
 /// in-app paste box (removed), it teaches adding links through the iOS Share
 /// menu by rendering the server's help page in a webview, so the copy ships via
-/// a hutch deploy rather than an App Store release. The help URL is discovered
-/// from the queue's `add-links-help` link `rel` (never a hard-coded route); when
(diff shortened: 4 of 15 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #847
- **Commit**: 1bb8d133c41bac49672e12595aa69390238eb78a

FagnerMartinsBrack:

None found.

## Medium Priority Issues

### 1. Client appends its injected `add-links-help` unconditionally — a server-advertised one would render a duplicate `+`

In `ReadingListViewModel.apply` (`projects/ios-readplace/App/ReadingListViewModel.swift:192`):

```swift
collectionAffordances = page.affordances.filter(\.isToolbarControl) + [Self.addLinksHelp]
```

A server-advertised `add-links-help` **link** passes `.isToolbarControl`: its presentation maps `add-links-help → isToolbarControl = true` (`AffordancePresentation.swift:68-73`) and a link is always `isInvokableByBareControl` (`AffordancePresentation.swift:121-126`). So if the server ever advertises `add-links-help` again — a rollback of this PR's server change, or an unrelated server dev re-adding it for another surface — the toolbar renders **two identical `+` controls**.

This is precisely the change the doctrine promises is safe ("a server affordance added later renders with zero client change") — here it renders as a visible duplicate. Now that the `+` is client-owned, the client should make its injected affordance canonical and drop any same-token server affordance before appending:

```swift
let serverControls = page.affordances.filter { $0.isToolbarControl && !Affordance.isAddLinksHelp($0.token) }
collectionAffordances = serverControls + [Self.addLinksHelp]
```

There is also **no test** for the "server advertises `add-links-help` **and** the client injects it" case — `testToolbarSurfacesOnlyTheClientAddControlForADefaultCollection` only covers the server omitting it. Add a case asserting exactly one `add-links-help` affordance when the queue advertises that link.

## Low Priority Suggestions

### 1. The help route `/help/add-links` is now hardcoded in the client, reintroducing URL coupling the previous design deliberately avoided

`AppConfig.addLinksHelpPath = "/help/add-links"` (`AppConfig.swift:50`) pins a concrete server route in the client. The code this PR removes kept the href *server-internal* on purpose — the old `collection-siren.ts` comment read *"the href stays server-internal so the bare help page can move without a client release."* The `hypermedia-api-design` doctrine lists this as its #1 anti-pattern. The practical cost is small (the help *content* still ships via a hutch deploy, and the route is stable), but relocating it now needs an App Store release. This is the explicit "deliberate departure" in the PR description — flagging to confirm the trade-off is accepted, not drift.

### 2. `AppConfig.addLinksHelpPath` is resolved twice from the same constant

Because `addLinksHelp` is a `static let`, it can't see `api.baseURL`, so the URL the sheet actually opens is resolved separately as `addLinksHelpURL` (`ReadingListViewModel.swift:51`). The tap handler routes on the affordance's `token` then opens `addLinksHelpURL`, so the injected affordance's own `href` is vestigial (it exists only because `Affordance(link:)` requires a non-nil href). Harmless, but it's two sources for one path — optional consolidation.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #847
- **Branch**: ios-plus-opens-add-links-help

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #847
- **Commit**: 9402c1a1f674ed0733295ad7c967e9cb41fc4795

FagnerMartinsBrack:
### Re-review of commit `9402c1a1`

The prior **MEDIUM** (duplicate `+`) is **fixed correctly and well-tested.** `apply` now drops any server affordance whose token is `add-links-help` via the single `Affordance.isAddLinksHelp` source before appending the canonical client control (`ReadingListViewModel.swift:194-199`), and `testToolbarKeepsExactlyOneAddControlWhenTheServerAlsoAdvertisesAddLinksHelp` (`ReadingListViewModelTests.swift:118`) proves the *survivor is the client's* (asserts `href == AppConfig.addLinksHelpPath`), not merely that the count is 1. Good fix. That fix, however, changes the calculus on the server-side removal — see Medium #1.

None found

## Medium Priority Issues

### 1. The server-side removal of `add-links-help` is now unnecessary, and it degrades already-installed iOS clients to the bundled fallback

Now that the client **dedupes** a server-advertised `add-links-help` (`ReadingListViewModel.swift:195-198`), the new client is immune to the server advertising that link — which removes the only reason this PR also had to drop it from `collection-siren.ts` in lockstep.

Dropping it **now** has a cost the dedupe fix doesn't cover: every **already-installed** iOS build still sources its help URL from the `add-links-help` rel on the queue (old `QueuePage.addLinksHelpHref`). With the link gone, those apps resolve `nil` and fall back to the bundled native page (`AddLinkInstructionsView.swift:47-49`). It's graceful (still teaches Share) but they lose the **server-rendered, deploy-updatable** help page until the user updates the app.

The `hypermedia-api-design` doctrine's deprecation order is *add-new → wait for clients to migrate → remove-old*; for an App-Store-gated client "migrate" means old versions aging out, which this PR doesn't wait for. Since the new client already dedupes, **keeping the server link costs the new client nothing** while preserving old clients' richer help.

Recommendation: keep `collection-siren.ts` advertising `add-links-help` (restore the `links.push` and its `collection-siren.test.ts` case) and defer the removal until old iOS versions age out. A repo-wide search confirms only iOS ever consumed this rel — the browser extensions and MCP never read it — so nothing else is affected either way.

This is **not blocking** and is acceptable as a deliberate trade-off if the team is comfortable with not-yet-updated apps showing the bundled fallback; flagging it for a conscious decision because the dedupe fix is new information that reframes the prior review's "old apps degrade gracefully" as a *reason to defer*, not a reason to proceed.

## Low Priority Suggestions

### 1. `/help/add-links` is hardcoded in the client (carry-over — confirm the trade-off)

`AppConfig.addLinksHelpPath = "/help/add-links"` (`AppConfig.swift:50`) pins a concrete server route in the client — the doctrine's #1 anti-pattern ("Naming a concrete server route … in client code"). This is the explicit "deliberate departure" in the PR description; the practical cost is small (the help *content* still ships via a hutch deploy; the route is stable), but relocating the route now needs an App Store release. Flagging to confirm the evolvability cost is accepted, consistent with the prior review.

### 2. The help path is held in two places that must stay in sync (carry-over — optional)

`addLinksHelpURL` is resolved in `init` (`ReadingListViewModel.swift:51`) and the injected affordance separately carries an `href` it never opens (the tap routes on `token`, then opens `addLinksHelpURL`), so that `href` is vestigial — it exists only because `Affordance(link:)` requires a non-nil href. Both derive from the one `AppConfig.addLinksHelpPath` constant, so there's no drift risk; optional consolidation only.

### 3. `testToolbarTracksTheCurrentResponsesAffordances` doesn't verify the `+` persists alongside the server action

The second assertion uses `compactMap(\.action).map(\.name) == ["purge-all"]` (`ReadingListViewModelTests.swift:262-265`), which filters out the `add-links-help` **link** — so the comment's claim "renders alongside the client + control" isn't actually checked here (a regression that dropped the injected `+` on a non-first load would slip past this test; its presence is covered only on the *first* load by other tests). Asserting `map(\.token) == ["purge-all", "add-links-help"]` would prove both render together and match the comment.

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **PR**: #847
- **Branch**: ios-plus-opens-add-links-help

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #847
- **Commit**: 6844225770953c6df9b798270feb92d89edb91ce

FagnerMartinsBrack:
### Re-review of commit `68442257`

The prior **MEDIUM** (the premature server-side removal degrading already-installed iOS clients) is **correctly resolved.** `collection-siren.ts:50-54` restores the `add-links-help` link, its `collection-siren.test.ts` cases are back (the dedicated "advertises a titled link" test plus the two `linkRels` ordering assertions), and the `server.ts:665-669` route comment now accurately says the collection "still advertises it via the add-links-help rel." The client dedupe from attempt-1 means the new client is immune to it, so old clients keep their server-rendered help at no cost to the new one. Good fix.

That restore, however, left the **iOS-side** comments asserting the opposite — see Medium #1.

None found

## Medium Priority Issues

### 1. Three client doc comments + the README still claim the server doesn't advertise `add-links-help` — now false after this commit restored it

Commit `68442257` re-added the server advertisement and updated the **two server-side** comments (`server.ts`, `collection-siren.ts`) to match — but the **four client-side claims** that the server does *not* advertise `add-links-help` were left untouched, so they now contradict the server code in this same PR:

- `ReadingListViewModel.swift:18-20` — *"the server no longer advertises it"*
- `ReadingListViewModel.swift:26` — *"never advertised by the server"*
- `ReadingListViewModel.swift:38` — *"the server advertises no such link"*
- `README.md:97-99` — *"an `add-links-help` affordance the app injects itself, not one advertised by the server"*

The server **does** advertise it again (`collection-siren.ts:50-54`). These comments are not merely suboptimal — `ReadingListViewModel.swift:18-20`/`:38` are flatly false, and they also contradict the **sibling comment in the same file** at `ReadingListViewModel.swift:186-189`, which exists *precisely because* the server can advertise it: *"a server that re-advertises add-links-help never renders a duplicate +."* One comment says "the server never advertises this"; another 160 lines down says "when the server advertises this, we dedupe it." Both can't be right.

This is the exact "comment rots when the surrounding code changes" failure CLAUDE.md warns against — here the surrounding code (the server half) changed in the *same PR*, and it's the same class of stale-comment hygiene the earlier review rounds kept fixing. The accurate framing is the one the code already implements: the client **ignores** the server-advertised `add-links-help` link and treats its own injected `+` as canonical (deduping any server-advertised one), rather than the server not advertising it at all. Reword the three comments and the README bullet to say that.

## Low Priority Suggestions

### 1. The restored `collection-siren.ts` comment "the help page can move without a client release" is now only true for old clients

`collection-siren.ts:49` says *"The href is server-internal, so the help page can move without a client release."* That held when the link was the only source of the URL. The **new** client now pins the route in `AppConfig.addLinksHelpPath = "/help/add-links"`, so relocating `/help/add-links` would break the shipping client (it resolves a 404) even though old clients would follow the moved rel. The comment is true *for the clients that read the link*, but a maintainer reading it could conclude the route is freely movable, which it no longer is. This is the same hardcoded-path trade-off already acknowledged in the PR description — just flagging that the server comment now slightly over-promises evolvability.

### 2. `testToolbarTracksTheCurrentResponsesAffordances` still doesn't verify the `+` persists alongside the server action (carry-over)

`ReadingListViewModelTests.swift:262-265` asserts `compactMap(\.action).map(\.name) == ["purge-all"]`, which filters out the `add-links-help` **link**, so the comment's claim *"renders alongside the client + control"* isn't actually checked on this non-first load. Asserting `map(\.token) == ["purge-all", "add-links-help"]` would prove both render together and match the comment.

### 3. `addLinksHelpURL` and the injected affordance's `href` are two derivations of one constant (carry-over — optional)

`addLinksHelpURL` is resolved in `init` (`ReadingListViewModel.swift:51`) while the injected affordance carries an `href` the tap path never opens (it routes on `token`, then opens `addLinksHelpURL`). Both derive from the one `AppConfig.addLinksHelpPath`, so there's no drift risk; optional consolidation only.

FagnerMartinsBrack:

## Context
- **Attempt**: 3 of 5
- **PR**: #847
- **Branch**: ios-plus-opens-add-links-help

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #847
- **Commit**: 21490470dadd705223af0ef801d2f4d2ea05ee7a

FagnerMartinsBrack:
### Re-review of commit `21490470`

The prior **MEDIUM** (three iOS doc comments + the README claiming the server doesn't advertise `add-links-help`) is **correctly resolved.** All four now describe what the code actually does — the client *injects its own* `+`, treats it as **canonical**, and **dedupes** any `add-links-help` the server also advertises:

- `ReadingListViewModel.swift:17-20` — "the client holds it itself rather than reading it from the server's add-links-help link" ✓
- `ReadingListViewModel.swift:23-27` — "injected by the client and kept canonical, so any add-links-help the server also advertises is deduped" ✓
- `ReadingListViewModel.swift:36-41` — "the client ignores any add-links-help the server advertises and treats this one as canonical" ✓
- `README.md:99-107` — "injects itself and treats as canonical, ignoring (deduping) any the server also advertises" ✓

These now agree with the restored server advertisement (`collection-siren.ts:50-54`) and the sibling dedupe comment at `ReadingListViewModel.swift:186-195`. The same-PR contradiction is gone.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `/help/add-links` is hardcoded in the client (carry-over — confirm the trade-off)

`AppConfig.addLinksHelpPath = "/help/add-links"` (`AppConfig.swift:50`) pins a concrete server route in the client — the `hypermedia-api-design` doctrine's #1 anti-pattern. This is the explicit "deliberate departure" in the PR description; the cost is small (the help *content* still ships via a hutch deploy; the route is stable). Flagging to confirm the evolvability cost is accepted, consistent with the prior reviews.

### 2. The restored `collection-siren.ts:49` comment now over-promises evolvability (carry-over)

`collection-siren.ts:49` still says *"The href is server-internal, so the help page can move without a client release."* That holds for the **old** clients that read this rel, but the **new** client pins the path in `AppConfig` (Low #1), so relocating `/help/add-links` would 404 the shipping client. Worth noting that the native fallback only triggers on a *load failure* (`AddLinkInstructionsView.swift:38-41`); a relocated route returning a 404 *page* would render in the WKWebView rather than the fallback. The comment is true for the clients that follow the link, just incomplete now — same hardcoded-path trade-off as Low #1.

### 3. `testToolbarTracksTheCurrentResponsesAffordances` doesn't verify the `+` persists alongside the server action (carry-over)

`ReadingListViewModelTests.swift:262-265` asserts `compactMap(\.action).map(\.name) == ["purge-all"]`, which filters out the `add-links-help` **link**, so the comment's claim *"renders alongside the client + control"* isn't actually checked on this non-first load. Asserting `map(\.token) == ["purge-all", "add-links-help"]` would prove both render together.

### 4. `addLinksHelpURL` and the injected affordance's `href` are two derivations of one constant (carry-over — optional)

`addLinksHelpURL` is resolved in `init` (`ReadingListViewModel.swift:53`) while the injected affordance carries an `href` the tap path never opens (it routes on `token`, then opens `addLinksHelpURL`). Both derive from the one `AppConfig.addLinksHelpPath`, so there's no drift risk; optional consolidation only.

PR #851, opened 2026-06-28
feat(extension): cap bulk-save page content client-side

FagnerMartinsBrack:
Drop the captured content of any BulkSavePage over a client-side cap
mirroring the server's MAX_PAGE_CONTENT_BYTES before batching, so the page
is sent URL-only and surfaced in the popup's too-big line via the result's
tooBig list. This avoids uploading content the server would only reject,
keeps each 20-page batch under MAX_BULK_CONTENT_REQUEST_BYTES, and lets a
batch of several oversize pages degrade per-page instead of failing
wholesale.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01X6yhuSxPFkm7VSfpNijD16

### DIFF
--- a/projects/browser-extension-core/src/core.test.ts
+++ b/projects/browser-extension-core/src/core.test.ts
@@ -2,7 +2,7 @@ import assert from "node:assert/strict";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
 import { initInMemoryAuth } from "./auth/in-memory-auth";
 import { UnauthorizedError } from "./auth/unauthorized-error";
-import { BrowserExtensionCore, BULK_SAVE_BATCH_SIZE } from "./core";
+import { BrowserExtensionCore, BULK_SAVE_BATCH_SIZE, MAX_BULK_SAVE_PAGE_CONTENT_BYTES } from "./core";
 import type { CoreError, ReadingList } from "./core";
 import { initInMemoryReadingList } from "./reading-list/in-memory-reading-list";
 import type { BulkSaveResult, BulkSavePage, SaveUrl, SavePages, SaveUrlResult } from "./reading-list/reading-list.types";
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #851
- **Commit**: 21efb0661e19d0354b0482cd30c4c50b28429c36

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The client↔server cap is a load-bearing invariant enforced only by a comment

The correctness of this change rests on `MAX_BULK_SAVE_PAGE_CONTENT_BYTES` (client, `core.ts:38`) staying `≤` the server's `MAX_PAGE_CONTENT_BYTES` (`src/packages/domain/src/article/article.schema.ts:17`), plus the matching `bytesToMb` rounding (`core.ts:42` vs `queue.page.ts:150`) and the identical strict-`>` boundary. All three match today (both `20 * 1024 * 1024`, identical rounding, `bytesToMb(20MiB+1)=20`). The risk is *drift*: if the server cap is lowered, the client silently resumes uploading content the server rejects — reintroducing the wholesale-batch failure this PR fixes, with no test going red. `browser-extension-core` has no dependency edge to `@packages/domain` (deps are only `@packages/hutch-logger` + `zod`), so the "can't import directly" rationale is accurate and follows the `BULK_SAVE_BATCH_SIZE` precedent. Consider extracting the constant + `bytesToMb` into a tiny shared workspace package (the domain article schema only needs `zod`, which the extension already has) so the mirror is compile-time-enforced.

### 2. The capping `.map` mixes transformation with a side effect

`core.ts:148-155` uses `pages.map(...)` both to return the url-only page and to `summary.tooBig.push(...)`. Correct, but a mapper that mutates external state reads awkwardly; a `for...of` or partition would separate "what's sent" from "what's reported." Purely stylistic.

### 3. An oversize page on an unsaveable host would appear in both `tooBig` and `skippedUrls`

The client caps content before the server validates the URL, whereas the server validates first (`queue.page.ts:653`) and returns early for unsaveable hosts — so server-side an unsaveable URL is only ever `skipped`, never `tooBig`. A page that is both >20 MiB and unsaveable would land in two summary lines. Extreme edge (unsaveable hosts are usually unscriptable, so `content` is absent); the link is still handled correctly.

### 4. Test coverage gap: no mixed batch (some over, some under cap)

The new tests cover all-oversize, exactly-at-cap kept, and single-oversize — good branch coverage including the boundary. Missing is a single `saveAll` call with both an oversize and an under-cap page, asserting the oversize one is stripped to URL-only while its sibling keeps its `content` in the same `savePages` call.

PR #852, opened 2026-06-28
feat(web-shell): make changelog NEW badge mean "unseen" per browser

FagnerMartinsBrack:
The site-wide changelog banner always rendered its NEW chip whenever the
banner was visible, so for anyone who never closed it NEW stayed lit on
every visit and stopped signalling novelty — and the next post shipped
already "NEW".

Make NEW mean unseen in this browser instead of merely undismissed. The
visible banner now carries its version as a data attribute and a trailing
synchronous inline script that records the version in
localStorage["readplace.changelog-seen"] on first sight and adds
.changelog-banner--seen (hiding the chip via CSS) on later loads of the
same version. A newer post re-lights NEW until it too has been seen once.
Running before paint avoids a flash of NEW on already-seen loads; storage
access is guarded so no-JS and private-mode browsers keep today's
always-NEW behaviour. Server-side visibility/dismissal is unchanged.

Recolour the chip to navy (--color-secondary) so it reads as "unseen"
against the amber chrome that links and CTAs already own.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_019yWKHn9NWQcbAR2QBTT2h9


FagnerMartinsBrack:

## Context
- **PR**: #852
- **Commit**: 7997952804d4f46cb12bc5808c7ff03603dc4afc

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #852
- **Branch**: claude/changelog-new-unseen-badge-5xlpny

FagnerMartinsBrack:

## Context
- **PR**: #852
- **Commit**: e3b692dd207eeb309d645737f18042bdee20accb

FagnerMartinsBrack:
### Review: changelog NEW badge means "unseen" per browser

The mechanic is sound. The version rides on `data-changelog-version`, the inline script reads it via `getAttribute` (never interpolated into the JS — the only interpolation is the hardcoded `CHANGELOG_SEEN_STORAGE_KEY`, so there's no injection surface), and storage access is wrapped in `try/catch` so no-JS and private-mode browsers keep today's always-NEW baseline. Running synchronously as the banner's last child, with the `--seen` rule already in `<head>`, means the chip is hidden before paint on already-seen loads — no flash. CSP is a non-issue: every deployable runs `helmet({ contentSecurityPolicy: false })`, so the inline `<script>` isn't blocked.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Chip foreground token (`--color-on-brand`) is named for a different background than it now sits on
`base.styles.ts:616-617` sets `background: var(--color-secondary)` (navy `#2B3A55`) while keeping `color: var(--color-on-brand)` (white `#FFFFFF`). The rendered result is correct and accessible — white-on-navy is ≈ 11.4:1 (WCAG AAA), and `--color-on-brand` is not overridden in the dark block (lines 56-75) so it stays white in both themes. The nit is purely semantic coupling: the foreground token is the one paired with the **brand** (amber) fill, so if `--color-on-brand` is ever retuned to a non-white value the navy chip's text would follow it incorrectly. There's no `--color-on-secondary` token today; introducing one (white in both themes) would make the pairing self-describing. Leaving it is defensible since white is the universal on-dark colour here.

### 2. The banner lives in the chrome (`.banner-area`), outside htmx's `<main>` swap target
Confirmed by reading `base.template.ts:96-103` (banner) vs `:117` (`{{{content}}}` → `<main>`). On `hx-boost` navigations htmx swaps only `<main>`, so the seen-script runs **once per full page load**, not per boosted navigation. Net effect: NEW persists across all in-session SPA navigations and only flips to hidden on the next *full* load (F5 / new tab / direct URL). That's reasonable — arguably nicer than per-view — and the SSR baseline is correct. Just worth a conscious note that "seen" is recorded at full-load granularity, not per route change.

### 3. "Seen" is recorded after a single render (threshold confirmation)
`changelog-banner.ts:105` writes the version on first sight, so a reader who never glances at the banner on their first load loses the chip on the next full load (the hook + "Read more" stay until server-side dismissal; only the chip goes). This matches the PR description, so it's intentional — flagging only to confirm the threshold is what you want.

### 4. Private-mode test forces `getItem` to throw; the realistic first-sight failure is `setItem`
`changelog-banner.test.ts:331` overrides `localStorage.getItem` to throw. The common real degradation (Safari private mode / quota) is `setItem` throwing on the **first sight**, after `getItem` already returned. Both live in the same `try`, so behaviour is identical and coverage is already 100% — a sibling test that throws from `setItem` on first sight would just document the actual failure path.

### 5. `CHANGELOG_SEEN_STORAGE_KEY` re-export has no out-of-package consumer
`index.ts:49` re-exports it, but only the in-package test imports it (via `./changelog-banner`). knip stays green because `src/index.ts` is a declared entry, so this is **not** a violation — just noting the public re-export is currently unused externally. Keeping it next to `CHANGELOG_SEEN_SCRIPT` for API symmetry is a fine reason to leave it.

FagnerMartinsBrack:

## Context
- **PR**: #852
- **Commit**: 296a6a78e4eca0317f634e9f1c5c4bc4c14b83f2

FagnerMartinsBrack:

**Found 2 comment(s) to fix** (and 2 that look compliant with the skill).

### `src/packages/web-shell/src/changelog-banner.ts`

> toggles `--seen` (which hides the chip via CSS) ... The first sight of a version records it, so reloads of the same announcement drop NEW while a newer post re-lights it. ... storage-blocked browsers 

- **Fix:** Trim to only the whys the minified one-liner genuinely cannot carry (reads from the data attribute so the script is never interpolated = no injection surface; runs before paint so an already-seen banner never flashes NEW; storage access guarded for private-mode throws). Remove the behaviour narration ('toggles --seen ... first sight records it ... reloads drop NEW ... newer post re-lights it') and the time-bound phrase 'today's always-NEW behaviour'. Whatever remains still needs explicit human approval to keep.

### `src/packages/web-shell/src/changelog-banner.ts`

> The visible branch also carries the version as a data attribute and a trailing inline script that turns the always-rendered NEW chip into an unseen-only signal per browser (CHANGELOG_SEEN_SCRIPT).

- **Fix:** Drop this appended sentence. It narrates what the template now emits (the version data attribute and the trailing inline script) which the markup already shows, and names the CHANGELOG_SEEN_SCRIPT symbol in prose (rots on rename). The original returnTo / no-referrer why in this comment is fine and unchanged.

FagnerMartinsBrack:
### Review: changelog NEW badge means "unseen" per browser

The mechanic is sound and I verified each load-bearing claim against the source:

- **No flash on already-seen loads.** `CHANGELOG_BANNER_STYLES` (incl. the new `.changelog-banner--seen .changelog-banner__chip { display: none }`) ships in the `<head>` `<style>` block (`base.template.ts:79-93`), and the seen-script is the banner's **last child** (`changelog-banner.ts:118`), so the class is added synchronously *before first paint* — the chip is never painted on a seen load, and dropping a `flex:0 0 auto` chip pre-paint means no visible layout shift either.
- **No injection surface.** The version is read at runtime via `getAttribute('data-changelog-version')`; the only value interpolated into the script string is the hardcoded `CHANGELOG_SEEN_STORAGE_KEY`. The `{{version}}` that lands in the attribute is a `ChangelogVersion` brand (`^[0-9a-f]{8}$`) and Handlebars-escaped, so the attribute is safe regardless.
- **Graceful degradation.** Storage access is wrapped in `try/catch`, so no-JS and private-mode browsers keep today's always-NEW baseline. A corrupted/garbage stored value self-heals (`!== version` → overwrite + show NEW). CSP is a non-issue — every deployable runs `helmet({ contentSecurityPolicy: false })`.
- **Selector hygiene.** The script targets the semantic class + a real data attribute (`.changelog-banner[data-changelog-version]`), **not** the `data-test-*` attribute — compliant with the web skill. The hidden banner omits the attribute, so the `if(!banner)return` guard handles the no-banner case; the parse-time fragment (`CHANGELOG_FRAGMENT_TEMPLATE`) lacks the `.changelog-banner` class so it can't be matched.
- **Inline is the correct call here.** A `defer`red external bundle can't run before paint, and the web skill's bundling rule targets `Function.toString()` helper leakage / CDN drift — neither applies to this hand-authored static string. There's direct precedent in the same template (`base.template.ts:104-116` set CSS vars pre-paint) and in `UPLOAD_AUTO_SUBMIT_SCRIPT`.
- **Contrast.** `--color-secondary` (navy `#2B3A55`, identical in both themes) with `--color-on-brand` (white, not overridden in dark) → ≈ 11.4:1, WCAG AAA.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Chip foreground token is named for a different background than it now sits on
`base.styles.ts:616-617` pairs `background: var(--color-secondary)` (navy) with `color: var(--color-on-brand)` (white). The render is correct and AAA-accessible, and `--color-on-brand` isn't overridden in the dark block so it stays white in both themes. The nit is purely semantic: the foreground token describes the **brand** (amber) fill, so if `--color-on-brand` is ever retuned to a non-white value the navy chip's text would follow it incorrectly. A `--color-on-secondary` token (white in both themes) would make the pairing self-describing. Defensible to leave, since white is the universal on-dark colour here. [Fix this →](https://claude.ai/code?q=In%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.styles.ts%20on%20branch%20claude%2Fchangelog-new-unseen-badge-5xlpny%20add%20a%20--color-on-secondary%20token%20(white%20in%20both%20light%20and%20dark%20themes)%20and%20use%20it%20for%20.changelog-banner__chip%20foreground%20instead%20of%20--color-on-brand%20so%20the%20token%20describes%20the%20navy%20background%20it%20sits%20on&repo=Readplace/readplace.com)

### 2. "Seen" is recorded at full-load granularity, not per htmx navigation
Confirmed the banner lives in `.banner-area` (`base.template.ts:96-103`), outside `{{{content}}}`/`<main>` (`:117`). On `hx-boost` navigations htmx swaps only `<main>`, so the seen-script runs **once per full page load**. Net effect: NEW persists across in-session SPA navigations and flips to hidden on the next *full* load (F5 / new tab / direct URL). Reasonable — arguably nicer than per-view — and the SSR baseline is correct. Flagging only so the granularity is a conscious choice.

### 3. "Seen" is recorded after a single render (threshold confirmation)
`changelog-banner.ts:105` writes the version on first sight, so a reader who never glances at the banner on their first load loses the chip on the next full load (the hook + "Read more" stay until server-side dismissal; only the chip goes). This matches the PR description, so it's intentional — flagging only to confirm the threshold is what you want.

### 4. Private-mode test throws from `getItem`; the realistic first-sight failure is `setItem`
The "swallows a throwing storage" test overrides `localStorage.getItem` to throw. The common real degradation (Safari private mode / quota) is `setItem` throwing on the **first sight**, after `getItem` already returned. Both live in the same `try`, so behaviour is identical and coverage is already 100% — a sibling test that throws from `setItem` on first sight would just document the actual failure path. [Fix this →](https://claude.ai/code?q=In%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fchangelog-banner.test.ts%20on%20branch%20claude%2Fchangelog-new-unseen-badge-5xlpny%20add%20a%20test%20where%20localStorage.setItem%20throws%20on%20first%20sight%20and%20assert%20CHANGELOG_SEEN_SCRIPT%20does%20not%20add%20changelog-banner--seen%20and%20does%20not%20throw&repo=Readplace/readplace.com)

### 5. `CHANGELOG_SEEN_STORAGE_KEY` re-export has no out-of-package consumer
`index.ts:49` re-exports it, but only the in-package test imports it (via `./changelog-banner`). knip stays green because `src/index.ts` is a declared entry, so this is **not** a violation — just noting the public re-export is currently unused externally. Keeping it next to `CHANGELOG_SEEN_SCRIPT` for API symmetry is a fine reason to leave it. [Fix this →](https://claude.ai/code?q=In%20src%2Fpackages%2Fweb-shell%2Fsrc%2Findex.ts%20on%20branch%20claude%2Fchangelog-new-unseen-badge-5xlpny%20remove%20the%20CHANGELOG_SEEN_STORAGE_KEY%20re-export%20since%20nothing%20outside%20the%20package%20imports%20it%2C%20or%20confirm%20it%20should%20be%20public%20API&repo=Readplace/readplace.com)

FagnerMartinsBrack:

## Context
- **PR**: #852
- **Commit**: 0d064dd55c4a697ba14369a0d66b46f46be1e2d6

FagnerMartinsBrack:
### Review: changelog NEW badge means "unseen" per browser (re-review @ `0d064dd5`)

This commit is **comment-only** (`docs(web-shell): trim changelog seen-script comments to why-only`); the runtime mechanic is unchanged from the prior approving reviews and I re-verified the load-bearing claims against the current source:

- **No flash / no layout shift on seen loads.** `.changelog-banner--seen .changelog-banner__chip { display: none }` ships in the `<head>` `<style>` and the seen-script is the banner's **last child** (`changelog-banner.ts:110`), so the class lands synchronously *before paint*; the `flex:0 0 auto` chip is never painted on a seen load.
- **No injection surface.** The version is read at runtime via `getAttribute('data-changelog-version')`; the only value interpolated into the script string is the hardcoded `CHANGELOG_SEEN_STORAGE_KEY`. The attribute itself carries a `ChangelogVersion` brand (`^[0-9a-f]{8}$`), Handlebars-escaped.
- **Graceful degradation.** `try/catch` around all storage access → no-JS / private-mode keep the always-NEW baseline; a corrupted stored value self-heals (`!== version` → overwrite + show NEW). CSP is a non-issue (`helmet({ contentSecurityPolicy: false })`).
- **Inline is correct here.** Confirmed direct precedent in the same template — `base.template.ts:104-108` already runs a pre-paint inline script to measure `.banner-area`. A `defer`red external bundle can't run before paint.
- **Comment trim landed correctly.** The `changelog-banner.ts` JSDocs now read as genuine *whys* (before-paint → no flash; attribute-read → no injection; guarded storage → private-mode safe), and the `CHANGELOG_SHELL_TEMPLATE` comment kept only the `returnTo`/`no-referrer` why. Pure comment diff, no code reflowed.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The trim missed one sibling: `base.styles.ts` `__chip` comment still narrates the seen-script's *what/how*
This commit correctly trimmed the seen-script behaviour narration out of `changelog-banner.ts`, but the **same pattern** survives in the chip's CSS comment (`base.styles.ts:604-606`):

> *"The inline seen-script adds `--seen` on a version this browser has already seen, dropping the chip so NEW signals novelty rather than merely 'not yet dismissed'."*

That sentence describes what the JS + the `.changelog-banner--seen .changelog-banner__chip { display:none }` rule (right below it) already do, and restates the feature-level why now carried in `changelog-banner.ts`'s `CHANGELOG_SEEN_SCRIPT` JSDoc and the PR description — exactly the code-comments skill *what/how* category you trimmed elsewhere. The **colour why** in the same block is legitimate and should stay (item 1's "never fully rounded" rule; item 2's first sentence — navy reads as "unseen" against the amber chrome, no collision with `--color-success`). Flagging only the trailing mechanism sentence, in case it wasn't a deliberate keep.

A lighter instance of the same: `CHANGELOG_SEEN_STORAGE_KEY`'s "Shared by the inline seen-script and its test…" (`changelog-banner.ts:91-92`) — the `export` already says "shared"; only "…so a newer post overwrites the old value and the entry never grows unbounded" is the load-bearing why. Per the skill, every **retained** why-comment also needs your explicit keep-approval, which you've been exercising — I'm not self-approving any.

### Still-standing low-priority notes from prior reviews (unchanged, none introduced by this commit)
2. **Chip foreground token** (`base.styles.ts:617`): navy `--color-secondary` bg paired with `--color-on-brand` (white) fg → correct & WCAG AAA (~11.4:1), white in both themes. Purely semantic: the fg token is named for the *amber* fill. A `--color-on-secondary` would self-describe. Defensible to leave.
3. **Full-load granularity:** banner lives in `.banner-area` (`base.template.ts:97`), outside `{{{content}}}`/`<main>` (`:117`), so on `hx-boost` navigations the seen-script runs once per *full* load, not per route change. Reasonable — just a conscious choice.
4. **Single-render threshold** (`changelog-banner.ts:99`): version recorded on first sight, so a reader who never glances at the banner loses the chip on the next full load (hook + "Read more" stay until server-side dismissal). Matches the PR intent.
5. **Private-mode test** throws from `getItem`; the realistic first-sight failure is `setItem` throwing after `getItem` returns. Same `try`, identical behaviour, coverage already 100%.
6. **`CHANGELOG_SEEN_STORAGE_KEY` re-export** (`index.ts:49`) has no out-of-package consumer (in-package test only); knip stays green because `src/index.ts` is a declared entry. API symmetry is a fine reason to keep it.

FagnerMartinsBrack:
@claude would this PR start showing the banner for those who have hidden it? Make sure there's no backward regression on those who already closed the current changelog banner.

PR #856, opened 2026-06-29
Migrate save-html to multipart save-content with PDF support

FagnerMartinsBrack:
Replaces the JSON-based `save-html` endpoint with a new `save-content` action that uses `multipart/form-data` encoding. This enables the iOS app to upload both rendered HTML and binary content (PDFs) without client-side size gating, letting the server decide on payload limits.

## Key Changes

- **API method rename**: `saveHTML()` → `saveContent()` with updated signature
  - Accepts `Data` + `mediaType` instead of `String rawHtml`
  - Builds multipart bodies with UUID boundaries instead of JSON
  - Includes `mediaType` field (`text/html` or `application/pdf`)

- **External content fetching**: New `fetchExternalContent()` method
  - Uses isolated `externalSession` (no bearer token, no redirect delegate)
  - Enforces 25 MiB ceiling to protect share extension memory budget
  - Returns nil on non-2xx or oversized responses, triggering URL-only fallback

- **PDF support in HTMLCaptor**
  - Detects main-frame MIME type via `decidePolicyFor navigationResponse`
  - Cancels PDF loads (avoids rendering large files in memory)
  - Stamps `mediaType` on captured page for save journey routing

- **SaveSharedPage orchestration**
  - New `resolveContentPayload()` helper routes HTML vs. PDF
  - PDFs are fetched directly and validated for `%PDF-` magic header
  - Degrades to URL-only save on fetch failure or invalid content

- **Test infrastructure**
  - Added `TestSupport.multipartParts()` parser (mirrors server's `extractAllParts`)
  - Validates boundary format, part headers, and binary content integrity
  - Updated all save-html tests to assert multipart wire format

- **Error handling**: Refusal (403) no longer triggers fallback save
  - Message-only refusals surface as `.refused` without attempting URL-only path
  - Fallback only fires on errors carrying a `save-article` action

## Implementation Notes

- Multipart boundary is a UUID per request (no collision risk)
- Content part includes `filename="content"` so server keys `isFile` correctly
- Title part omitted entirely when nil (not sent as empty string)
- External fetch never carries `Authorization` or `X-Readplace-Client` headers
- Server decides payload limits; client never gates on size

https://claude.ai/code/session_01C1ai5C5rManihGWkJ7vpoH

### DIFF
--- a/projects/ios-readplace/README.md
+++ b/projects/ios-readplace/README.md
@@ -3,8 +3,9 @@
 An iPhone app that behaves like the Readplace **browser
 extension**: it lists your saved links and lets you save new ones by **sharing a
 URL to it** from any app. When you share a link, the app renders the page in a
-hidden `WKWebView` first and uploads the **rendered HTML** (not just the URL) via
-the server's `save-html` Siren action — exactly what the extension does.
+hidden `WKWebView` first and uploads the **captured content** (the rendered HTML,
+or a shared PDF's bytes — not just the URL) via the server's `save-content` Siren
+action — exactly what the extension does.
 
 It lives under `projects/` as an nx project
 (`ios-readplace`). It builds with its own Swift/fastlane toolchain rather
@@ -93,10 +94,12 @@ That produces `build/Readplace-unsigned.ipa` (the app + its share extension).
   `status` field on its request, not its URL, so the endpoint can move freely.
 - **Save by sharing**: a **Share Extension** appears in the iOS share sheet for
   URLs/web pages. It loads the page in an off-screen `WKWebView`, captures
-  `document.documentElement.outerHTML`, and POSTs `{url, rawHtml, title}` to
-  `POST /queue/save-html`. If the token is missing, capture fails, or the HTML is
-  over the 10 MiB server limit, it degrades to the URL-only `save-article` path —
-  mirroring the extension's own fallback.
+  `document.documentElement.outerHTML`, and uploads it as a `multipart/form-data`
+  `{url, mediaType, title, content}` body to `POST /queue/save-content`. A shared
+  URL that resolves to a PDF is fetched directly (the web view declines to render
+  it) and uploaded as `application/pdf` instead. If the token is missing, capture
+  fails, or the content is over the server's size limit, it degrades to the
+  URL-only `save-article` path — mirroring the extension's own fallback.
 - **Add links via Share**: the `+` button is a client-side control — an
   `add-links-help` affordance the app injects itself and treats as canonical,
   ignoring (deduping) any the server also advertises. It opens a sheet whose
@@ -241,7 +244,8 @@ cases:
   tokens.
 - **API**: entry-point `303` redirect with the `Authorization` header preserved,
   `401` → single refresh → retry (and no retry loop when refresh fails),
-  `save-html` body + fallback to URL-only on an error action, `save-article`,
+  `save-content` multipart body + fallback to URL-only on an error action,
+  the external content fetch never leaking the bearer token, `save-article`,
   the status action posting the urlencoded `status` field, following the `303`
   back to the collection and verifying the status at the protocol level only (any
   non-2xx/3xx surfaces a generic server error — no per-code special-casing),
--- a/projects/ios-readplace/README.md
+++ b/projects/ios-readplace/README.md
@@ -97,9 +97,12 @@ That produces `build/Readplace-unsigned.ipa` (the app + its share extension).
   `document.documentElement.outerHTML`, and uploads it as a `multipart/form-data`
   `{url, mediaType, title, content}` body to `POST /queue/save-content`. A shared
   URL that resolves to a PDF is fetched directly (the web view declines to render
-  it) and uploaded as `application/pdf` instead. If the token is missing, capture
-  fails, or the content is over the server's size limit, it degrades to the
-  URL-only `save-article` path — mirroring the extension's own fallback.
+  it) and uploaded as `application/pdf` instead. If capture fails, or a shared PDF
+  is blocked or larger than the client's own 25 MiB external-fetch ceiling, it
+  degrades to the URL-only `save-article` path — and the server's crawler, which
+  allows far larger PDFs, then fetches it. (A payload that instead exceeds the
+  *server's* cap comes back with that same URL-only fallback action.) This mirrors
+  the extension's own fallback.
 - **Add links via Share**: the `+` button is a client-side control — an
   `add-links-help` affordance the app injects itself and treats as canonical,
   ignoring (deduping) any the server also advertises. It opens a sheet whose
@@ -260,10 +263,13 @@ cases:
   without exchanging the code), then a reading-list load preserving the bearer
   token across the entry-point `303`, and both sign-out paths (`logout` /
   `forceLogout`) dropping the minted `hutch_sid` session cookie.
-  Share-save — `SaveSharedPage` saving rendered HTML when it's under the cap,
-  degrading to URL-only when the capture is empty or the HTML is over the cap,
-  short-circuiting before any network call when logged out or when there's no
-  link, and reporting no-op when the server advertises no save action.
+  Share-save — `SaveSharedPage` saving rendered HTML via `save-content`, saving a
+  shared PDF's fetched bytes as `application/pdf`, degrading to URL-only when the
+  capture is empty, the PDF fetch is blocked, or the server refuses the payload
+  with a fallback action, short-circuiting before any network call when logged out
+  or when there's no link, and reporting no-op when the server advertises no save
+  action; plus the `HTMLCaptor` navigation-response decision (render HTML vs.
+  capture a main-frame PDF as a file) in isolation.
 - **Web-auth flow (shared by Login & Sign up)**: the native authorize requests
   (`readplace://oauth-callback` redirect + `screen_hint=login`/`signup`), the
   deep-link callback exchanging the code with the native `redirect_uri` and
--- a/projects/ios-readplace/ShareExtension/ShareViewController.swift
+++ b/projects/ios-readplace/ShareExtension/ShareViewController.swift
@@ -1,8 +1,9 @@
 import UIKit
 
-/// The share-sheet entry point. Renders the shared page in a WKWebView,
-/// captures its HTML, and saves it via the `save-html` action — degrading to a
-/// URL-only save if the token is missing, capture fails, or the HTML is too big.
+/// The share-sheet entry point. Renders the shared page in a WKWebView (or
+/// fetches a shared PDF's bytes), uploads the captured content via the
+/// `save-content` action — degrading to a URL-only save if the token is missing,
+/// capture fails, or the content is too big.
 @MainActor
 final class ShareViewController: UIViewController {
 	private let store = TokenStore()
--- a/projects/ios-readplace/Shared/HTMLCaptor.swift
+++ b/projects/ios-readplace/Shared/HTMLCaptor.swift
@@ -1,15 +1,27 @@
 import UIKit
 import WebKit
 
-/// The rendered page content captured from a WKWebView.
+/// The rendered page content captured from a WKWebView, plus the media type of
+/// the loaded main-frame resource (`text/html` for a rendered page,
+/// `application/pdf` for a PDF the captor declined to render, nil when the load
+/// failed before a response). The media type lets the save journey pick the
+/// matching `save-content` upload instead of degrading every non-HTML resource
+/// to a URL-only crawl.
 struct CapturedPage {
 	let rawHtml: String?
 	let title: String?
+	let mediaType: String?
+
+	init(rawHtml: String?, title: String?, mediaType: String? = nil) {
+		self.rawHtml = rawHtml
+		self.title = title
+		self.mediaType = mediaType
+	}
 }
 
 /// Loads a URL in an off-screen WKWebView and returns the rendered DOM as
 /// `document.documentElement.outerHTML` plus `document.title` — the same
-/// content the browser extension captures and sends to `save-html`.
+/// content the browser extension captures and uploads via `save-content`.
 ///
 /// Resolves on first main-frame load completion (after a short settle delay so
 /// script-rendered content is present) or when the timeout elapses, whichever
@@ -23,6 +35,7 @@ final class HTMLCaptor: NSObject, WKNavigationDelegate {
 	private var continuation: CheckedContinuation<CapturedPage, Never>?
 	private var timeoutTask: Task<Void, Never>?
 	private var settleSeconds: Double = 0.4
+	private var detectedMediaType: String?
 
 	override init() {
 		let configuration = WKWebViewConfiguration()
@@ -44,6 +57,34 @@ final class HTMLCaptor: NSObject, WKNavigationDelegate {
 				await self?.finish(extractContent: true)
 			}
 			webView.load(URLRequest(url: url))
+		}
+	}
+
+	// Records the main-frame resource's media type before deciding whether to
+	// render it. A PDF is cancelled rather than loaded — a large PDF would blow
+	// the share extension's memory budget — and finished as a non-extract capture
+	// so the save journey fetches the bytes itself and uploads them as a file.
+	// Setting `detectedMediaType` before `.cancel` (which provokes a
+	// didFailProvisionalNavigation re-entry) means the media type is already
+	// stamped no matter which path resolves `finish` first.
+	nonisolated func webView(
+		_ webView: WKWebView,
+		decidePolicyFor navigationResponse: WKNavigationResponse,
+		decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void
+	) {
+		let mimeType = navigationResponse.response.mimeType
+		let isMainFrame = navigationResponse.isForMainFrame
+		Task { @MainActor [weak self] in
+			guard let self else { return decisionHandler(.allow) }
+			if isMainFrame {
+				self.detectedMediaType = mimeType
+				if mimeType == "application/pdf" {
+					decisionHandler(.cancel)
+					await self.finish(extractContent: false)
+					return
+				}
+			}
+			decisionHandler(.allow)
 		}
 	}
 
@@ -75,11 +116,11 @@ final class HTMLCaptor: NSObject, WKNavigationDelegate {
 		timeoutTask?.cancel()
 		timeoutTask = nil
 
-		var page = CapturedPage(rawHtml: nil, title: nil)
+		var page = CapturedPage(rawHtml: nil, title: nil, mediaType: detectedMediaType)
 		if extractContent {
 			let html = (try? await webView.evaluateJavaScript("document.documentElement.outerHTML")) as? String
 			let title = (try? await webView.evaluateJavaScript("document.title")) as? String
-			page = CapturedPage(rawHtml: html, title: title)
+			page = CapturedPage(rawHtml: html, title: title, mediaType: "text/html")
 		}
 		webView.stopLoading()
 		continuation.resume(returning: page)
--- a/projects/ios-readplace/Shared/HTMLCaptor.swift
+++ b/projects/ios-readplace/Shared/HTMLCaptor.swift
@@ -72,20 +72,33 @@ final class HTMLCaptor: NSObject, WKNavigationDelegate {
 		decidePolicyFor navigationResponse: WKNavigationResponse,
 		decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void
 	) {
-		let mimeType = navigationResponse.response.mimeType
-		let isMainFrame = navigationResponse.isForMainFrame
+		let decision = Self.navigationResponseDecision(
+			mimeType: navigationResponse.response.mimeType,
+			isMainFrame: navigationResponse.isForMainFrame
+		)
 		Task { @MainActor [weak self] in
 			guard let self else { return decisionHandler(.allow) }
-			if isMainFrame {
-				self.detectedMediaType = mimeType
-				if mimeType == "application/pdf" {
-					decisionHandler(.cancel)
-					await self.finish(extractContent: false)
-					return
-				}
+			switch decision {
+			case .allow(let detectedMediaType):
+				if let detectedMediaType { self.detectedMediaType = detectedMediaType }
+				decisionHandler(.allow)
+			case .captureAsFile(let mediaType):
+				self.detectedMediaType = mediaType
+				decisionHandler(.cancel)
+				await self.finish(extractContent: false)
 			}
-			decisionHandler(.allow)
 		}
+	}
+
+	enum NavigationResponseDecision: Equatable {
+		case allow(detectedMediaType: String?)
+		case captureAsFile(mediaType: String)
+	}
+
+	nonisolated static func navigationResponseDecision(mimeType: String?, isMainFrame: Bool) -> NavigationResponseDecision {
+		guard isMainFrame else { return .allow(detectedMediaType: nil) }
+		if mimeType == "application/pdf" { return .captureAsFile(mediaType: "application/pdf") }
+		return .allow(detectedMediaType: mimeType)
 	}
 
 	// WKNavigationDelegate's requirements are nonisolated; these hop to the main
--- a/projects/ios-readplace/Shared/ReadplaceAPI.swift
+++ b/projects/ios-readplace/Shared/ReadplaceAPI.swift
@@ -263,7 +263,7 @@ final class ReadplaceAPI {
 	/// — a large scanned PDF, say — never lands in the extension's memory budget
 	/// whole; returns nil on any non-2xx, oversize, or transport failure so the
 	/// caller degrades to a URL-only save.
-	func fetchExternalContent(_ url: URL) async -> (Data, String)? {
+	func fetchExternalContent(_ url: URL) async -> Data? {
 		var request = URLRequest(url: url)
 		request.httpMethod = "GET"
 		guard let (stream, response) = try? await externalSession.bytes(for: request),
@@ -281,8 +281,7 @@ final class ReadplaceAPI {
 		} catch {
 			return nil
 		}
-		let contentType = http.value(forHTTPHeaderField: "Content-Type") ?? ""
-		return (data, contentType)
+		return data
 	}
 
 	/// Builds a `multipart/form-data` request with a UUID boundary for
--- a/projects/ios-readplace/Shared/ReadplaceAPI.swift
+++ b/projects/ios-readplace/Shared/ReadplaceAPI.swift
@@ -512,8 +512,6 @@ private final class RedirectHeaderPreservingDelegate: NSObject, URLSessionTaskDe
 }
 
 private extension Data {
-	/// Appends a string's UTF-8 bytes, for assembling a multipart body whose text
-	/// segments interleave with raw binary content.
 	mutating func append(_ string: String) {
 		append(Data(string.utf8))
 	}
--- a/projects/ios-readplace/Shared/ReadplaceAPI.swift
+++ b/projects/ios-readplace/Shared/ReadplaceAPI.swift
@@ -81,13 +81,6 @@ struct QueuePage {
 final class ReadplaceAPI {
 	private static let logger = Logger(subsystem: "com.readplace", category: "ReadplaceAPI")
 
-	/// Conservative ceiling on bytes pulled into the extension for an external
-	/// content fetch. Well under the server's `MAX_PDF_BYTES` OCR ceiling: that is
-	/// an origin-side limit, whereas the share extension holds the fetched bytes
-	/// plus a duplicated multipart body in a tight memory budget, so an oversize
-	/// resource degrades to a URL-only save before the buffers are ever built.
-	private static let maxExternalContentBytes = 25 * 1024 * 1024
-
 	let baseURL: String
 	private let store: TokenStore
 	private let oauth: OAuthService
@@ -96,12 +89,25 @@ final class ReadplaceAPI {
 	// and never via `send()`, so neither the bearer nor the redirect-preserving
 	// re-attachment can leak the Readplace `Authorization` header to that origin.
 	private let externalSession: URLSession
+	/// Conservative ceiling on bytes pulled into the extension for an external
+	/// content fetch. Well under the server's `MAX_PDF_BYTES` OCR ceiling: that is
+	/// an origin-side limit, whereas the share extension holds the fetched bytes
+	/// plus a duplicated multipart body in a tight memory budget. `fetchExternalContent`
+	/// streams the body and stops the moment the running total crosses this ceiling
+	/// (refusing outright when the response announces an oversize length), so an
+	/// oversize resource degrades to a URL-only save without ever being buffered whole.
+	private let maxExternalContentBytes: Int
 
 	// Defaults to an ephemeral configuration so the session's cookie jar is its
 	// own isolated, in-memory store rather than process-wide `HTTPCookieStorage.shared`:
 	// the `hutch_sid` cookie minted by `bootstrapSession` must not linger in the
 	// shared jar where it would outlive the session and leak across sign-outs.
-	init(baseURL: String, store: TokenStore, sessionConfiguration: URLSessionConfiguration = .ephemeral) {
+	init(
+		baseURL: String,
+		store: TokenStore,
+		sessionConfiguration: URLSessionConfiguration = .ephemeral,
+		maxExternalContentBytes: Int = 25 * 1024 * 1024
+	) {
 		self.baseURL = baseURL
 		self.store = store
 		self.oauth = OAuthService(baseURL: baseURL, store: store, sessionConfiguration: sessionConfiguration)
@@ -113,6 +119,7 @@ final class ReadplaceAPI {
 			delegateQueue: nil
 		)
 		self.externalSession = URLSession(configuration: sessionConfiguration)
+		self.maxExternalContentBytes = maxExternalContentBytes
 	}
 
 	// MARK: - Reading list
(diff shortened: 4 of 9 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #856
- **Branch**: claude/ios-save-content-migration-t15nn3

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: bc2f0f2b33703171eec1bf8459a33133a414961b

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content with PDF support

I cross-checked the client's wire format against the server it talks to (`multipart-upload.ts`, `queue.page.ts` save-content route, `save-content-limit-handler.ts`) and traced the capture → resolve → upload → fallback paths. **The contract matches the server exactly** and the design is careful. One issue is worth fixing before merge.

#### What's correct (verified, not assumed)
- **Wire format matches the server byte-for-byte.** Field names `url` / `mediaType` / `title` / `content`, the `filename="content"` attribute that the server keys `isFile` off (`multipart-upload.ts:117`, looked up at `queue.page.ts:881-884`), the `\r\n` framing and `--<boundary>` delimiters all line up with `extractAllParts`. The new `TestSupport.multipartParts` is a faithful port, so tests assert real server-parseable bytes.
- **Security property is real and tested.** `externalSession` has no delegate and never flows through `send()`, so the bearer and `X-Readplace-Client` can't leak to a third-party origin (`ReadplaceAPI.swift:115`, `:256`); `testFetchExternalContentSendsNoAuthorization` asserts both headers are absent.
- **Reentrancy is handled.** The `.cancel` in `decidePolicyFor` provokes a `didFailProvisionalNavigation` re-entry, but `finish` nil-checks the continuation (`HTMLCaptor.swift:114`) so there's no double-resume, and `detectedMediaType` is stamped before `.cancel`.
- **No breaking change.** The server still advertises `save-html` alongside `save-content`, so a new-app/old-server skew degrades to URL-only rather than failing. Refusal-before-fallback ordering is preserved (`ReadplaceAPI.swift:235`).

### 1. The 25 MiB external-fetch ceiling is checked *after* the whole response is buffered, so it doesn't protect the share-extension memory budget

`ReadplaceAPI.swift:256-263` — `fetchExternalContent` awaits `externalSession.data(for:)`, which fully downloads the response into `data` **before** `data.count <= Self.maxExternalContentBytes` is ever evaluated:

```swift
guard let (data, response) = try? await externalSession.data(for: request),
    let http = response as? HTTPURLResponse,
    (200...299).contains(http.statusCode),
    data.count <= Self.maxExternalContentBytes   // ← runs only after the full body is resident
else { return nil }
```

This directly contradicts the rationale comment on `maxExternalContentBytes` (`:159-165`): *"an oversize resource degrades to a URL-only save **before the buffers are ever built**."* The buffer **is** built first. A large PDF — exactly the input this PR adds support for, and routinely 50–200 MiB for scanned/image-heavy docs — gets fully loaded into the share extension's tight memory budget, risking a jetsam kill **before** the size check runs. When that happens the save is silently lost with *no* URL-only fallback (the process is dead), which is the opposite of the intended graceful degradation.

`data(for:)` offers no hook to abort mid-download. Stream instead so resident memory stays near the ceiling:

```swift
let (stream, response) = try await externalSession.bytes(for: request)
guard let http = response as? HTTPURLResponse, (200...299).contains(http.statusCode) else { return nil }
if http.expectedContentLength > Int64(Self.maxExternalContentBytes) { return nil } // advisory early-out
var data = Data()
do {
    for try await byte in stream {
        data.append(byte)
        if data.count > Self.maxExternalContentBytes { return nil } // abort before the budget blows
    }
} catch { return nil }
```

(While here: `multipartRequest` also copies `content` into a second `Data`, so peak is ~2× the payload even for an in-spec PDF — capping the fetch near the ceiling matters doubly.)

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `fetchExternalContent`'s returned content-type is dead in production
It returns `(Data, String)` (`ReadplaceAPI.swift:256`), but the only production caller discards it: `let (bytes, _) = await api.fetchExternalContent(url)` (`SaveSharedPage.swift:76`) — the `%PDF-` magic check is what actually validates. Only the test reads the string. Consider returning just `Data?`, or use the content-type as a cheap pre-filter before the magic-byte check.

### 2. The riskiest new code path has no automated test
`HTMLCaptor.webView(_:decidePolicyFor:decisionHandler:)` (`HTMLCaptor.swift:70-89`) — the PDF-cancel + double-`finish` reentrancy — is the most subtle addition (the author's own comment flags how delicate the ordering is), yet it's only exercised indirectly via `FakeHTMLCaptor`. WKWebView is the established untestable boundary, but the *decision* itself (`mimeType` + `isMainFrame` → allow/cancel) is pure and could be lifted into a tiny testable helper so a regression in the branching is caught without WebKit.

### 3. The cookie-less re-fetch will degrade many gated PDFs to URL-only
`externalSession` carries no cookies and the WKWebView's session isn't shared, so PDFs behind a publisher/paywall/bot-defence gate will often return a non-PDF challenge, fail the `%PDF-` check, and fall back to URL-only — for exactly the high-value PDFs users most want captured. This is a documented, safe trade-off (degrades cleanly); flagging only so the reduced success rate is a conscious choice.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #856
- **Branch**: claude/ios-save-content-migration-t15nn3

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: 9dac7c022ae6141f53fa0c7e43c21d696f5482c5

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content with PDF support (re-review @ `9dac7c02`)

This commit addresses the prior round's single HIGH issue (unbounded external fetch). I re-traced the capture → resolve → fetch → upload → fallback paths and re-checked the wire format against the server it talks to. **The streaming fix is correct, and there's no contract drift.**

#### The streaming fix is correct (verified, not assumed)
- `fetchExternalContent` (`ReadplaceAPI.swift:266-286`) now uses `externalSession.bytes(for:)` and (a) refuses up front when `expectedContentLength` exceeds the ceiling, and (b) aborts mid-stream the moment the running total crosses `maxExternalContentBytes`. Resident memory is bounded **and** total transfer is bounded — the early `return nil` inside the `for try await` drops the `AsyncBytes` iterator, which cancels the underlying task, so a lying/omitted `Content-Length` can't pull the whole body. Any non-2xx / oversize / transport failure returns nil → URL-only fallback, exactly as before.
- The compression edge is handled correctly: `expectedContentLength` is the *compressed* size, but the loop guard counts *decompressed* bytes, so the real protection is the loop, with the announced-length check as a cheap early-out.
- The ceiling is an injectable init param (default 25 MiB), and **both** rejection branches are now tested: `testFetchExternalContentAbortsWhenStreamedBytesExceedCeiling` (mid-stream) and `testFetchExternalContentRejectsResponseAnnouncingOversizeLength` (announced length). The no-auth-leak test still holds. The corrected `maxExternalContentBytes` doc comment now matches the streaming reality.

#### No contract drift (re-verified against the server)
- save-content route (`queue.page.ts:845-940`) **requires** `mediaType` (422 if absent) and keys the file part off `filename` via `isFile` (`multipart-upload.ts:114-118`). The client sends `mediaType` and `filename="content"`. ✓
- Both `"application/pdf"` and `"text/html"` handlers exist server-side (`queue.page.ts:308-323`); the client only ever sends those two values. ✓
- Server's save-content `maxBytes` is `MAX_PDF_BYTES.bytes` = **500 MB** (`pdf-page-limits.ts:12-15`), so the client comment "well under the server's `MAX_PDF_BYTES` OCR ceiling" is accurate. ✓
- The server re-validates the `%PDF-` magic header (`queue.page.ts:310`), so the client's `resolveContentPayload` pre-check is belt-and-suspenders, not the sole guard. ✓
- `saveHTML`/`rawHtml` are fully gone from the client; `formRequest` is still used by `invocationRequest` (no dead code).

#### Reentrancy still correct
`detectedMediaType` is stamped before `.cancel`, `finish(extractContent:false)` runs synchronously on the MainActor before the `didFailProvisionalNavigation` re-entry is dispatched, and the continuation nil-check makes the second `finish` a no-op (`HTMLCaptor.swift:70-127`).

None found — the prior round's HIGH issue is fixed and verified.

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Per-byte iteration is correct but not the cheapest option
`for try await byte in stream` + `data.append(byte)` (`ReadplaceAPI.swift:277-279`) walks the body one `UInt8` at a time — ~26M iterations and single-byte appends for a 25 MiB PDF. It's bounded and the author flagged the trade-off, so it's fine as-is. If profiling ever shows it matters, a `URLSessionDataDelegate` consuming `didReceive data:` at chunk granularity (cancel when the accumulator crosses the ceiling) gives the same memory/transfer cap with chunk-sized appends.

### 2. The `reserveCapacity` true-branch is untested
`if http.expectedContentLength > 0 { data.reserveCapacity(...) }` (`ReadplaceAPI.swift:275`) is reached only when the announced length is positive **and** within the ceiling — but no test sets an in-bounds `Content-Length` (the no-auth test sends none → −1; the oversize test is rejected by the guard first). A stub with `Content-Length` equal to a small in-bounds body would cover it and assert the known-length happy path.

### 3. "Client never gates on size" is not literally true for the PDF fetch
The PR says *"Server decides payload limits; client never gates on size,"* but the 25 MiB external-fetch ceiling **is** a client-side size gate for shared PDFs. It's justified (a share extension can't buffer a 500 MB download) and degrades cleanly — a >25 MiB PDF becomes a URL-only save and the server's own crawler (500 MB ceiling) fetches it, so reachable large PDFs are still saved; only a PDF that is *both* >25 MiB *and* unreachable by the server crawler is lost. Worth a line in the README/PR so the gate reads as a conscious exception rather than a contradiction. Relatedly, the README's "over the server's size limit" (`README.md:100-101`) is imprecise for the PDF path — that degrade is the *client's* 25 MiB ceiling, not the server's.

### 4. (carried from last round, still open) The riskiest branch has no direct test
`HTMLCaptor.webView(_:decidePolicyFor:…)` (`HTMLCaptor.swift:70-89`) — the `mimeType`+`isMainFrame` → cancel/allow decision — is still only exercised indirectly via `FakeHTMLCaptor`. The pure decision could be lifted into a tiny testable helper so a regression in the branching is caught without WebKit.

### 5. (carried) `fetchExternalContent`'s returned content-type is unused in production
`SaveSharedPage.swift:76` discards it (`let (bytes, _)`); only the test reads the string — the `%PDF-` magic check is what validates. Either drop it to `Data?` or use the content-type as a cheap pre-filter before the magic-byte check.

FagnerMartinsBrack:

**Found 3 comment(s) to fix** (and 7 that look compliant with the skill).

### `projects/ios-readplace/Tests/TestSupport.swift`

> /// One part of a parsed `multipart/form-data` body.

- **Fix:** Remove. The doc restates the type name `MultipartPart` and its multipart/form-data context, which the struct and its fields (name/filename/contentType/body) already convey. No why.

### `projects/ios-readplace/Shared/ReadplaceAPI.swift`

> /// Appends a string's UTF-8 bytes, for assembling a multipart body whose text segments interleave with raw binary content.

- **Fix:** Remove. The body is `append(Data(string.utf8))` — the lead clause restates it verbatim; the trailing motivation is recoverable from the call sites in multipartRequest. If kept at all it needs explicit human approval.

### `projects/ios-readplace/Shared/SaveSharedPage.swift`

> Capture the page → list the queue → when there is content to upload (captured HTML, or the bytes of a shared PDF...) attempt the `save-content` upload... when there is no content to upload, save the U

- **Fix:** Drop the step-by-step flow narration; it mirrors run() line-for-line (a how restatement). The preceding unchanged sentence about the decision tree running against real API/token types (testability) is the genuine why and should remain.

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: b6c13a0f16d4a60428abaf7961b10344222d2d75

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content with PDF support — re-review @ `b6c13a0f` (post-merge)

This commit merges `main` into the branch. I re-confirmed the PR's own diff is unchanged, re-traced capture → resolve → fetch → upload → fallback, and re-checked the wire format against the server's parser **at this commit**. **The merge is clean, there's no contract drift, and the prior round's single HIGH (unbounded external fetch) is still fixed.**

#### The merge is clean (verified)
- `b6c13a0f`'s second parent is `main`. It integrated the `ios-app` OAuth-client rename (README / `AppConfig.swift` / OAuth tests, from #857), the new **`code-comments` skill**, and unrelated inbox changes — none of which touch the save-content logic. `git diff origin/main...HEAD` is still **+515/−115 across the same 8 files**.
- The README merge correctly kept **both** independently-edited paragraphs — "Save by sharing" (this PR) and "Add links via Share" (main) — neither reverted.

#### Wire format still matches the server byte-for-byte (re-verified against `extractAllParts`)
- `multipartRequest` (`ReadplaceAPI.swift:292-324`) ↔ `extractAllParts` (`multipart-upload.ts:89-125`): `url`/`mediaType`/`title` carry **no** `filename` → `isFile=false`; `content` carries `filename="content"` → `isFile=true`. The route looks each up with exactly those predicates (`queue.page.ts:881-884`), **requires** `mediaType` (422 otherwise, `:902-911`), and the trailing `--<boundary>--` is caught by the server's `0x2d 0x2d` end check (`:98`). `TestSupport.multipartParts` is a faithful port, so the tests assert real server-parseable bytes.
- Handlers exist for exactly the two media types the client ever sends — `application/pdf` / `text/html` (`queue.page.ts:308-323`) — and the server re-validates `%PDF-` (`:310`), so `resolveContentPayload`'s magic-byte pre-check is belt-and-suspenders, not the sole guard.

#### Prior HIGH still fixed; reentrancy still correct
- `fetchExternalContent` (`ReadplaceAPI.swift:266-286`) streams via `bytes(for:)`, early-outs when the announced `expectedContentLength` exceeds the ceiling, and aborts mid-stream once the running total crosses `maxExternalContentBytes` — resident memory **and** total transfer bounded. Both branches are tested via an injected tiny ceiling; the no-auth-leak test still holds.
- `detectedMediaType` is stamped before `.cancel`; the non-extract `finish` runs to completion without an `await` suspension, so the `didFailProvisionalNavigation` re-entry hits the continuation nil-check → no double-resume (`HTMLCaptor.swift:70-127`).

None found — the prior round's HIGH is fixed and verified.

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The 3 comments you flagged against the `code-comments` skill are still present at this commit (fix already in flight)
This merge brought the `code-comments` skill into the repo as the single source of truth, and the three what/how comments you flagged remain at `b6c13a0f`: `TestSupport.swift:121` (`/// One part of a parsed multipart/form-data body.`), `ReadplaceAPI.swift:515-516` (`/// Appends a string's UTF-8 bytes…`), and `SaveSharedPage.swift:27-31` (the `run()` flow narration). Your comment-review classification is correct — each narrates *what/how* the code already says. Flagging only so the review reflects the current tree; a targeted fix is already running, so no auto-fix link here to avoid racing it.

### 2. The `reserveCapacity` true-branch is untested
`if http.expectedContentLength > 0 { data.reserveCapacity(...) }` (`ReadplaceAPI.swift:275`) only runs when the announced length is positive **and** within the ceiling, but no test sets an in-bounds `Content-Length` — the no-auth test sends none (→ −1), the oversize test is rejected by the guard first. A stub with `Content-Length` equal to a small in-bounds body covers it. [Fix this →](https://claude.ai/code?q=In%20projects/ios-readplace/Tests/ReadplaceAPITests.swift%20add%20a%20test%20for%20fetchExternalContent%20where%20the%20stub%20sets%20an%20in-bounds%20Content-Length%20equal%20to%20the%20body%20size%20%28positive%20and%20within%20the%20ceiling%29%2C%20so%20the%20reserveCapacity%20branch%20at%20ReadplaceAPI.swift%3A275%20is%20exercised.%20Branch%20claude/ios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

### 3. Per-byte stream iteration is correct but not the cheapest option
`for try await byte in stream` + `data.append(byte)` (`ReadplaceAPI.swift:277-279`) walks the body one `UInt8` at a time (~26M iterations for a 25 MiB PDF). Bounded and the author flagged the trade-off, so fine as-is; a `URLSessionDataDelegate` consuming `didReceive data:` at chunk granularity (cancel when the accumulator crosses the ceiling) would give the same memory/transfer cap with chunk-sized appends if profiling ever shows it matters.

### 4. `fetchExternalContent`'s returned content-type is unused in production
It returns `(Data, String)` (`ReadplaceAPI.swift:266`); the only production caller discards it (`let (bytes, _)`, `SaveSharedPage.swift:76`) — the `%PDF-` magic check is what validates. Only the test reads the string. Either drop it to `Data?` or use the content-type as a cheap pre-filter before the magic-byte check.

### 5. (carried) The riskiest branch has no direct test
`HTMLCaptor.webView(_:decidePolicyFor:…)` (`HTMLCaptor.swift:70-89`) — the `mimeType`+`isMainFrame` → cancel/allow decision — is still only exercised indirectly via `FakeHTMLCaptor`. WKWebView is the established untestable boundary here, but the pure decision could be lifted into a tiny testable helper so a regression in the branching is caught without WebKit.

### 6. README "over the server's size limit" is imprecise for the PDF path
`README.md:100` says shared content degrades to URL-only when "over the server's size limit", but for a PDF that degrade is the **client's** own 25 MiB external-fetch ceiling (`fetchExternalContent` → nil), not the server's. A >25 MiB PDF becomes a URL-only save and the server's own 500 MB crawler then fetches it — worth a word so the client gate reads as a conscious exception. [Fix this →](https://claude.ai/code?q=In%20projects/ios-readplace/README.md%20around%20line%20100%20the%20%27Save%20by%20sharing%27%20section%20says%20content%20degrades%20to%20URL-only%20when%20%27over%20the%20server%27s%20size%20limit%27%2C%20but%20for%20the%20PDF%20path%20that%20degrade%20is%20the%20client%27s%20own%2025%20MiB%20external-fetch%20ceiling%20in%20fetchExternalContent%20%28returns%20nil%29%2C%20not%20the%20server%27s%20limit.%20Clarify%20that%20a%20PDF%20over%20the%20client%27s%2025%20MiB%20ceiling%20or%20a%20failed/blocked%20fetch%20degrades%20to%20URL-only%20and%20the%20server%27s%20500%20MB%20crawler%20then%20fetches%20it.%20Branch%20claude/ios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: 8862e13e793cd09a4cab2d42d0cd9cf2299cc07e

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content — re-review @ `8862e13e`

This commit is the **comment-only cleanup** that resolves the prior round's LOW item #1: it removes the three what/how comments flagged against the `code-comments` skill. `git diff b6c13a0f..8862e13e` is **3 deletions, 0 insertions**, all `///`/`//` doc lines. All substantive Swift is byte-identical to `b6c13a0f`, which already cleared three review rounds with no HIGH/MEDIUM. I re-traced the contract at this commit rather than assume.

#### The comment removals are clean (verified)
- `ReadplaceAPI.swift:514` — `private extension Data` now has `append(_:)` directly; no orphaned `///`, compiles unchanged.
- `SaveSharedPage.swift:23-25` — the `run()` flow narration is gone; the genuine **testability *why*** (*"…lifted out of `ShareViewController` so the full decision tree runs against the real API and token types under test…"*) was correctly kept.
- `TestSupport.swift:121` — `MultipartPart`'s doc line removed; the struct + its `name`/`filename`/`contentType`/`body` fields convey it. Clean.

#### Fallback decode path — deep-dived this round, confirmed correct
The new `saveContent` fallback (`ReadplaceAPI.swift:243-254`) builds a JSON `{url, title?}` and follows the server's `save-article` action **without** setting `Prefer: return=representation` (unlike the standalone `saveArticle`, `:330`). I checked whether that omission breaks the response decode:
- The advertised `save-article` href is `/queue` → the Siren API route at `queue.page.ts:544`, which returns **`201` + `toArticleEntity`** whenever `wantsSiren(req)` — it never reads `Prefer`. The fallback request flows through `send()`, which sets `Accept: application/vnd.siren+json`, so `wantsSiren` holds and `decodeArticle` parses the entity. ✓
- The `303`-redirecting handler (`queue.page.ts:968`) is a **different path** (`/queue/save`, the browser save-bar), which the iOS client never calls. So the `Prefer` header on `saveArticle` is a harmless pre-existing no-op, and the fallback's omission of it is also harmless. **No bug.**

#### Wire format & reentrancy still hold (re-verified at this commit)
- `multipartRequest` (`ReadplaceAPI.swift:292-324`) ↔ `extractAllParts` (`multipart-upload.ts:89-125`): `url`/`mediaType`/`title` carry no `filename` → `isFile=false`; `content` carries `filename="content"` → `isFile=true`. The route looks each up with exactly those predicates (`queue.page.ts:881-884`) and **requires** `mediaType` (422 + fallback action otherwise, `:902-911`). `TestSupport.multipartParts` is a faithful byte-for-byte port. ✓
- Both `application/pdf` and `text/html` handlers exist server-side and the server re-validates `%PDF-` (`queue.page.ts:309-323`), so the client's magic-byte pre-check is belt-and-suspenders.
- Streaming fetch (`bytes(for:)`, `:266-286`) still bounds resident memory **and** total transfer; both rejection branches are tested. `detectedMediaType` is stamped before `.cancel`, and `finish`'s continuation nil-check keeps the `didFailProvisionalNavigation` re-entry a no-op (`HTMLCaptor.swift:70-127`).

None found — comment-only commit; the prior round's HIGH (unbounded external fetch) remains fixed and verified.

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The `reserveCapacity` true-branch is still untested
`if http.expectedContentLength > 0 { data.reserveCapacity(...) }` (`ReadplaceAPI.swift:275`) runs only when the announced length is **positive and within** the ceiling. Confirmed against `StubURLProtocol` (it builds the `HTTPURLResponse` straight from `stub.headers`, so a missing `Content-Length` → `expectedContentLength == -1`): no test sets an in-bounds `Content-Length`, so this branch never executes. A stub whose `Content-Length` equals a small in-bounds body covers it. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FTests%2FReadplaceAPITests.swift%20add%20a%20test%20for%20fetchExternalContent%20where%20the%20stub%20sets%20an%20in-bounds%20Content-Length%20equal%20to%20the%20body%20size%20so%20the%20reserveCapacity%20branch%20at%20ReadplaceAPI.swift%3A275%20is%20exercised.%20Branch%20claude%2Fios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

### 2. README "over the server's size limit" is imprecise for the PDF path
`README.md:99-102` says shared content degrades to URL-only when "over the server's size limit", but for a PDF that degrade is the **client's** own 25 MiB external-fetch ceiling (`fetchExternalContent` → nil), not the server's. A >25 MiB PDF becomes a URL-only save and the server's own 500 MB crawler then fetches it. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FREADME.md%20around%20lines%2099-102%20the%20%27Save%20by%20sharing%27%20section%20says%20content%20degrades%20to%20URL-only%20when%20%27over%20the%20server%27s%20size%20limit%27%2C%20but%20for%20the%20PDF%20path%20that%20degrade%20is%20the%20client%27s%20own%2025%20MiB%20external-fetch%20ceiling%20in%20fetchExternalContent%20%28returns%20nil%29%2C%20not%20the%20server%27s.%20Clarify%20that%20a%20PDF%20over%20the%20client%27s%2025%20MiB%20ceiling%20or%20a%20failed%2Fblocked%20fetch%20degrades%20to%20URL-only%20and%20the%20server%27s%20500%20MB%20crawler%20then%20fetches%20it.%20Branch%20claude%2Fios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

### 3. `fetchExternalContent`'s returned content-type is unused in production
It returns `(Data, String)` (`ReadplaceAPI.swift:266`); the only production caller discards it (`let (bytes, _)`, `SaveSharedPage.swift:70`) — the `%PDF-` magic check is what validates. Only the test reads the string. Either drop it to `Data?` or use the content-type as a cheap pre-filter before the magic-byte check. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FShared%2FReadplaceAPI.swift%20fetchExternalContent%20returns%20%28Data%2C%20String%29%20but%20the%20only%20production%20caller%20in%20SaveSharedPage.swift%3A70%20discards%20the%20content-type.%20Either%20change%20the%20return%20type%20to%20Data%3F%20or%20use%20the%20content-type%20as%20a%20cheap%20pre-filter%20before%20the%20%25PDF-%20magic-byte%20check.%20Branch%20claude%2Fios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

### 4. (carried) The riskiest branch has no direct test
`HTMLCaptor.webView(_:decidePolicyFor:…)` (`HTMLCaptor.swift:70-89`) — the `mimeType`+`isMainFrame` → cancel/allow decision — is still only exercised indirectly via `FakeHTMLCaptor`. WKWebView is the established untestable boundary, but the pure decision could be lifted into a tiny testable helper so a regression in the branching is caught without WebKit. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FShared%2FHTMLCaptor.swift%20lift%20the%20mimeType%2BisMainFrame%20to%20allow%2Fcancel%20decision%20in%20webView%28_%3AdecidePolicyFor%3A%29%20at%20lines%2070-89%20into%20a%20small%20pure%20testable%20helper%2C%20and%20add%20a%20unit%20test%20so%20a%20regression%20in%20the%20PDF-cancel%20branching%20is%20caught%20without%20WebKit.%20Branch%20claude%2Fios-save-content-migration-t15nn3&repo=Readplace/readplace.com)

### 5. (carried) Per-byte stream iteration is correct but not the cheapest option
`for try await byte in stream` + `data.append(byte)` (`ReadplaceAPI.swift:277-279`) walks the body one `UInt8` at a time (~26M iterations for a 25 MiB PDF). Bounded and the author flagged the trade-off, so fine as-is; a `URLSessionDataDelegate` consuming `didReceive data:` at chunk granularity would give the same cap with chunk-sized appends if profiling ever shows it matters.

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: a936cd6fba6a04b97eda78dfa21d94ba195b52ec

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content — @ `a936cd6f`

The tip commit is a **comment-only cleanup** (`git show a936cd6f` = **9 deletions, 0 insertions**, all `///`/`//` doc lines — the three what/how comments flagged against the `code-comments` skill). The substantive Swift is byte-identical to the code that already cleared prior rounds with no HIGH/MEDIUM. I re-traced the contract at this commit against the server rather than assume.

#### Verified at this commit (not assumed)
- **Wire format matches the server byte-for-byte.** `multipartRequest` (`ReadplaceAPI.swift:292-324`) ↔ `extractAllParts` (`multipart-upload.ts:89-125`): `url`/`mediaType`/`title` carry no `filename` → `isFile=false`; `content` carries `filename="content"` → `isFile=true`. The route looks each up with exactly those predicates (`queue.page.ts:958-961`), **requires** both `content` (422, `:968`) and `mediaType` (422, `:979`) with a `save-article` fallback action attached, and the trailing `--<boundary>--` is caught by the server's `0x2d 0x2d` end check (`:98`). `TestSupport.multipartParts` is a faithful port, so the tests assert real server-parseable bytes.
- **Media types line up.** Handlers exist for exactly the two values the client ever sends — `application/pdf` / `text/html` (`queue.page.ts:340-355`) — and the PDF handler re-validates `%PDF-` server-side (`:342`), so `resolveContentPayload`'s magic-byte pre-check (`SaveSharedPage.swift:71`) is belt-and-suspenders, not the sole guard.
- **Prior HIGH (unbounded fetch) still fixed.** `fetchExternalContent` (`ReadplaceAPI.swift:266-286`) streams via `bytes(for:)`: refuses up front when `expectedContentLength` exceeds the ceiling (`:272`) and aborts mid-stream once the running total crosses `maxExternalContentBytes` (`:278-279`) — resident memory **and** total transfer bounded (the early `return nil` drops the `AsyncBytes` iterator, cancelling the task, so an omitted/lying `Content-Length` can't pull the whole body). Both branches are tested via an injected tiny ceiling; `Int(expectedContentLength)` at `:275` is safe because the guard already bounded it to 25 MiB.
- **Security property real and tested.** `externalSession` has no delegate and never flows through `send()`, so the bearer and `X-Readplace-Client` can't leak to a third-party origin; `testFetchExternalContentSendsNoAuthorization` + the PDF-fetch test in `SaveSharedPageTests` both assert the headers are absent.
- **Reentrancy correct.** `detectedMediaType` is stamped before `.cancel`, `finish(extractContent:false)` runs to completion with no `await` suspension before `continuation.resume`, and the continuation nil-check makes the `didFailProvisionalNavigation` re-entry a no-op (`HTMLCaptor.swift:70-127`).
- **No breaking change.** An old server that only advertises `save-article` (no `save-content`) degrades to URL-only via the second `if` in `run()` (`SaveSharedPage.swift:49`); a PDF-fetch failure degrades the same way — both covered by tests.

#### Comment cleanup is clean
The removed comments were pure *what/how*; the genuine testability *why* on `SaveSharedPage` (*"…lifted out of `ShareViewController` so the full decision tree runs against the real API and token types under test…"*) was correctly kept. Deletions don't affect Swift compilation.

None found — comment-only commit; the prior round's HIGH remains fixed and verified.

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The `reserveCapacity` true-branch is still untested
`if http.expectedContentLength > 0 { data.reserveCapacity(...) }` (`ReadplaceAPI.swift:275`) runs only when the announced length is **positive and within** the ceiling. No test sets an in-bounds `Content-Length` — the no-auth test sends none (→ `-1`), and the oversize test is rejected by the guard first — so this branch never executes. A stub whose `Content-Length` equals a small in-bounds body covers it.

### 2. README "over the server's size limit" is imprecise for the PDF path
`README.md:99-102` says shared content degrades to URL-only when "over the server's size limit", but for a PDF that degrade is the **client's** own 25 MiB external-fetch ceiling (`fetchExternalContent` → nil), not the server's. A >25 MiB PDF becomes a URL-only save and the server's own 500 MB crawler then fetches it.

### 3. `fetchExternalContent`'s returned content-type is unused in production
It returns `(Data, String)` (`ReadplaceAPI.swift:266`); the only production caller discards it (`let (bytes, _)`, `SaveSharedPage.swift:70`) — the `%PDF-` magic check is what validates. Only the test reads the string. Either drop it to `Data?` or use the content-type as a cheap pre-filter before the magic-byte check.

### 4. (carried) The riskiest branch has no direct test
`HTMLCaptor.webView(_:decidePolicyFor:…)` (`HTMLCaptor.swift:70-89`) — the `mimeType`+`isMainFrame` → cancel/allow decision — is still only exercised indirectly via `FakeHTMLCaptor`. WKWebView is the established untestable boundary, but the pure decision could be lifted into a tiny testable helper so a regression in the PDF-cancel branching is caught without WebKit.

### 5. (carried) Per-byte stream iteration is correct but not the cheapest option
`for try await byte in stream` + `data.append(byte)` (`ReadplaceAPI.swift:277-279`) walks the body one `UInt8` at a time (~26M iterations for a 25 MiB PDF). It's memory- and transfer-bounded and the author flagged the trade-off, so it's fine as-is — but `URLSession.AsyncBytes` per-byte iteration is notably slow, so a large in-spec PDF could add a user-invisible (background) but multi-second stall. If that ever matters, a `URLSessionDataDelegate` consuming `didReceive data:` at chunk granularity (cancel when the accumulator crosses the ceiling) gives the same memory/transfer cap with chunk-sized appends.

FagnerMartinsBrack:

Also double check the share extension will allow pdf webpages to be shared to readplace on iPhone after this PR

FagnerMartinsBrack:

## Context
- **PR**: #856
- **Commit**: 32fb0528c2420717047ed2e5873e3d3f73e97f45

FagnerMartinsBrack:
### Review: Migrate save-html to multipart save-content — @ `32fb0528`

The tip commit applies four of the five carried LOW suggestions (`git show 32fb0528` = **+94/−23** across 6 files, all client-side Swift/README/tests — no server, no wire-format change). I traced each change and re-checked the contract against the server at this commit rather than assume from prior rounds.

#### The four fixes are correct (verified, not assumed)

**1. `fetchExternalContent` return `(Data, String)? → Data?`** (`ReadplaceAPI.swift:266-285`) — the content-type was dead in production (`SaveSharedPage.swift:70` already discarded it; the `%PDF-` magic check is the real validator). The lone reader was the test, which is updated (`ReadplaceAPITests.swift:345`). Clean removal, no behavior change.

**2. Navigation-response decision lifted into a pure helper — behavior-preserving.** I diffed the pre-refactor delegate body (`a936cd6f`) against `navigationResponseDecision(mimeType:isMainFrame:)` + the switch (`HTMLCaptor.swift:70-102`). Every reachable state maps identically:
| main-frame? | mimeType | old | new |
|---|---|---|---|
| yes | `application/pdf` | stamp → cancel → finish(false) | `.captureAsFile` → stamp → cancel → finish(false) |
| yes | `text/html` | stamp → allow | `.allow(text/html)` → stamp → allow |
| yes | `nil` | stamp nil → allow | `.allow(nil)` → **skip stamp** → allow |
| no | any | (untouched) → allow | `.allow(nil)` → (untouched) → allow |

The *only* divergence is the main-frame-`nil` row: the old code clobbered `detectedMediaType` to `nil`; the new code leaves it. That's unreachable in practice (one main-frame response per navigation, `detectedMediaType` inits `nil`) and strictly safer (never overwrites a set type with `nil`). Crucially, **the reentrancy contract is intact**: `detectedMediaType` is still stamped *before* `.cancel` in the `.captureAsFile` arm, and `finish(extractContent:false)` still runs to `continuation.resume`/`nil` with no suspension before the `.cancel`-provoked `didFailProvisionalNavigation` re-entry hits the nil-check. The new `HTMLCaptorTests` cover all four branches of the pure decision — including the important **sub-frame PDF → allow** (a PDF in an `<iframe>` must not cancel the whole page). The residual untested code is only the thin actor-hopping glue at the genuine WebKit boundary; the branching logic the reviews flagged as risky is now directly tested.

**3. `reserveCapacity` branch now covered.** I confirmed `StubURLProtocol` builds the `HTTPURLResponse` straight from `stub.headers` (`StubURLProtocol.swift:75-80`), so setting `Content-Length: "\(body.count)"` yields a positive in-bounds `expectedContentLength` → the `> 0` guard fires `reserveCapacity` (`ReadplaceAPI.swift:275`). The new test (`ReadplaceAPITests.swift:402`) genuinely exercises the known-length happy path the no-`Content-Length` (→ −1) and oversize (rejected first) tests never reached.

**4. README wording** (`README.md:100-104`, `:266-272`) now correctly attributes the PDF degrade to the *client's* own 25 MiB external-fetch ceiling (with the server's 500 MB crawler picking it up), and separately notes the server-cap case returns the same URL-only fallback action. Accurate.

#### Contract re-verified at this commit (server unchanged)
- `contentPart = parts.find(p => p.name === "content" && p.isFile)`; `url`/`mediaType`/`title` require `!p.isFile` (`queue.page.ts:958-961`). The client's `multipartRequest` sends `filename="content"` only on the content part → matches. `mediaType` and non-empty `content` are each **required** (422 + a `save-article` fallback action, `:968`/`:979`), which the client's `saveContent` fallback decodes and follows. `application/pdf`/`text/html` handlers both exist and the server re-validates `%PDF-` (`:341-343`), so the client pre-check is belt-and-suspenders.
- Prior HIGH (unbounded fetch) still fixed: `bytes(for:)` streams, early-outs on announced oversize, aborts mid-stream; the no-auth-leak property still holds and is tested.

None found — comment/refactor-level commit; the prior round's HIGH remains fixed and verified.

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. (carried, consciously deferred) Per-byte stream iteration
`for try await byte in stream` + `data.append(byte)` (`ReadplaceAPI.swift:277-279`) still walks the body one `UInt8` at a time (~26M iterations for a 25 MiB PDF). It's memory- **and** transfer-bounded and correct; the author explicitly deferred it with reviewer agreement (a `URLSessionDataDelegate` consuming `didReceive data:` at chunk granularity is the riskiest change to make in a share extension and its multi-chunk path can't be exercised by `StubURLProtocol`, which delivers the body in a single `didLoad`). No action needed — flagging only for completeness. If profiling ever shows a large in-spec PDF stalls the background save, revisit it as its own change with a real-device loop.

PR #858, opened 2026-06-29
Keep link count badge in sync during article extraction

FagnerMartinsBrack:
## Summary

Refactors the inbox email detail header to keep the link count badge synchronized with the articles panel during extraction. Previously, the badge was conditionally rendered only when a count was available. Now it's always present as a stable OOB swap target, rendering empty while extraction is pending and updating via out-of-band swap when extraction completes.

## Changes

- **New component:** `renderInboxLinkCount()` — extracts the link count badge into a reusable component that renders with an `id` and optional `hx-swap-oob="outerHTML"` attribute for out-of-band updates
- **Poll response:** The `/inbox/:id/articles?poll=N` endpoint now appends the link count badge as an OOB swap to every response, keeping the header in lockstep with the panel
- **Template refactor:** Replaced conditional `{{#if linkCountLabel}}` rendering with unconditional `{{{linkCountHtml}}}` injection, moving the conditional logic into the component
- **CSS:** Added `.inbox-email-detail__link-count:empty { display: none; }` to hide the empty placeholder during extraction, keeping the meta row's flex layout clean
- **Tests:** Updated assertions to verify the badge always renders (empty while extracting, populated on completion) and carries the OOB swap attribute in poll responses

## Implementation Details

The badge is now a stable anchor point (`id="inbox-email-detail-link-count"`) that htmx can target for out-of-band swaps. While extraction is pending, `label` is `undefined`, so the badge renders empty and is hidden via CSS. Once extraction completes, the terminal poll tick includes the badge with the count as an OOB swap, allowing the header to update instantly without waiting for a full page reload.

https://claude.ai/code/session_011Rgv7zwxRbgXT4VqtD7r1B

### DIFF
--- a/projects/hutch/src/runtime/web/pages/inbox/inbox-email-detail.component.ts
+++ b/projects/hutch/src/runtime/web/pages/inbox/inbox-email-detail.component.ts
@@ -4,6 +4,7 @@ import { render } from "@packages/web-shell";
 import type { PageBody } from "@packages/web-shell";
 import { renderInboxArticlesPanel } from "./inbox-articles-panel.component";
 import { buildInboxEmailIframeSrcdoc } from "./inbox-email-iframe-srcdoc";
+import { renderInboxLinkCount } from "./inbox-link-count.component";
 import { INBOX_EMAIL_DETAIL_STYLES } from "./inbox-email-detail.styles";
 import type { InboxEmailDetailViewModel } from "./inbox-email-detail.viewmodel";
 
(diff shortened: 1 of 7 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #858
- **Commit**: c6da95f1845d84820b14690512def5dd765dd36e

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The now-live badge isn't announced to assistive tech

The badge used to appear only on a full page load (static). After this PR it updates mid-session via `hx-swap-oob` when extraction completes, which makes it a dynamically-updating region — but there's no `aria-live`, so a screen-reader user won't hear "1 link" when it lands. The subtlety: the badge **element itself** is the OOB `outerHTML` target (`id="inbox-email-detail-link-count"`), and replacing a live-region node doesn't announce the new node's content — the live region must be a **stable wrapper** that persists across swaps, with the badge swapped inside it. Optional polish, not a regression.

### 2. The same "withheld during extraction / OOB anchor" rationale is restated in four places

The component JSDoc (`inbox-link-count.component.ts`), the route comment (`inbox.page.ts:174`), the CSS comment (`inbox-email-detail.styles.css:41`), and the viewmodel comment (`inbox-email-detail.viewmodel.ts:89`) each re-explain the same mechanism. Each sits at a different layer and matches the repo's why-comment house style, but the component JSDoc already covers it end-to-end, so the others could be trimmed to a one-line pointer. Purely optional.

PR #862, opened 2026-06-29
Add reader-ready email marker to owner reader links

FagnerMartinsBrack:
## Summary

Adds a `from=reader-ready-email` query parameter to links in the reader-ready notification emails, allowing the system to distinguish email clicks from shared permalinks and route logged-out owners to login (since they already own the article) rather than the public `/view` share page.

## Changes

- **New module `owner-reader-link.ts`**: Exports `buildOwnerReaderPath()` and `wantsOwnerLogin()` to centralize the marker constant and logic, ensuring producer (email handler) and consumer (permalink resolver) stay in sync.

- **`reader-permalink.ts`**: Added early redirect for logged-out visitors with the email marker — sends them to `/login?return=/queue/{id}/view?from=reader-ready-email` instead of the public `/view` share page. Logged-in owners and non-owners ignore the marker and follow existing logic.

- **`reader-ready-notify-handler.ts`**: Updated email generation to include the marker in the reader CTA link via `buildOwnerReaderPath()`.

- **Tests**: Added comprehensive coverage:
  - `queue.reader.route.test.ts`: Logged-out owner redirects to login with return path; logged-in owner renders private reader despite marker.
  - `reader-permalink.test.ts`: Logged-out visitor with marker redirects to login; logged-in owner ignores marker and renders reader; logged-in non-owner ignores marker and gets public `/view` redirect.
  - `reader-ready-notify-handler.test.ts`: Verifies email CTA includes the marker.

## Implementation Details

The marker is an internal, same-deploy contract between the email producer and the permalink resolver. It only affects the redirect decision for unauthenticated requests — authenticated users (owners and non-owners) follow their normal paths regardless of the marker's presence.

https://claude.ai/code/session_01D9N5KLL1LbDd8PLUbGyft1

PR #866, opened 2026-06-30
Persist iOS reader state across opens; destroy all sessions on iOS sign-out

FagnerMartinsBrack:
## Summary

iOS reader state now survives between article opens, and iOS sign-out destroys every server session the user holds — while other devices and the browser extensions keep their tokens and transparently re-mint.

## What actually changes on iOS

**Reader state persists between article opens.** Each article previously opened in a WKWebView backed by a fresh `.nonPersistent()` data store, so anything written inside one reader died when the sheet closed. The reader now uses the persistent, process-wide `WKWebsiteDataStore.default()`.

The state this preserves in practice is the **founder share balloon's localStorage dismissal** (`readplace.share-dismissed`): the iOS reader is the *chromeless* render (`/queue/:id/view?platform=ios`), which by design contains no banners (`queue.reader-platform.route.test.ts` pins `.banner-area` to null), so the changelog banner never renders there and its cookie is not what the store switch fixes on the happy path. Server-set cookies (e.g. `rp_changelog_dismissed`) are preserved too, which matters on the edge path where the reader webview follows a redirect to a full-shell page (e.g. an expired session landing on the public `/view` permalink, where a guest-visible banner with a working dismiss form does render).

**Sign-out wipes the WebKit store without dropping server-set cookies.** `AppSession` deletes only cookies named `hutch_sid` from the default store (a blanket cookie wipe would drop server-set state like a changelog dismissal), then clears every non-cookie data type — cache, localStorage, IndexedDB, service workers — so the signed-out account's reading history doesn't stay on disk. Deliberate tradeoff: that non-cookie wipe also resets the share balloon's localStorage dismissal, so the balloon can reappear once after a sign-out/sign-in cycle (including the `forceLogout()` path triggered by session expiry).

- `hutch_sid` is re-injected per reader open (same name/domain/path overwrites any persisted copy), so the reader always authenticates as the current user.
- `logout()` awaits the WebKit wipe before publishing the logged-out state; `forceLogout()` stays synchronous and fire-and-forgets the wipe (returned as a `Task` so tests can await it).
- The wipe is injected as a closure seam on `AppSession` for testing; the real WebKit call remains the untested OS boundary.

## What changes on the server

**`POST /oauth/revoke` for the `ios-app` client destroys all of the user's sessions, then revokes only the presented token.** iOS mints a server session per reader open and keeps none of their ids, so the only way to honor its sign-out is a bulk destroy by user. Sessions are destroyed before the token is revoked, so a failed destroy leaves the token valid and the client's retry re-runs the whole sign-out. Other clients' tokens (browser extensions, other iOS devices) survive on purpose: their devices transparently re-mint a session on next use instead of being signed out by someone else's phone. Revokes presented by any other client stay scoped to the presented token, as before.

- `DestroyUserSessions` contract added; DynamoDB implementation paginates the new sessions-table `userId-index` (KEYS_ONLY) and deletes each session; in-memory test double mirrors it.
- `revokeDestroysUserSessions(clientId)` names the ios-app-only cascade in `@packages/domain/oauth`.
- Shared `forEachQueryPage` helper extracted in `@packages/hutch-storage-client`; `revokeAllUserTokens` refactored onto it.

**Infra:** adds the `userId-index` GSI to the sessions table and flips its IAM grant to `includeIndexes: true`. (The GSI did not previously exist — earlier drafts assumed it did.)

**Password reset destroys all of the user's sessions** after a successful reset, so a stolen session can't outlive the credential that minted it. A reset token whose account no longer exists now gets a clean 400 instead of an unhandled conditional-check failure.

**Changelog dismiss cookie lifetime: 23 months requested** (up from 1 year). WebKit and Firefox honor the full value for server-set first-party cookies; Chrome 104+ clamps it to 400 days per rfc6265bis. Requesting only 400 days would have shortened every other engine to Chrome's floor.

## Tests

- Integration: iOS revoke destroys all sessions, revokes only the presented token, and the extension bystander token survives; extension revoke stays token-scoped; password reset invalidates pre-reset sessions; ghost-account reset token 400s.
- Unit: `destroyUserSessions` GSI query shape, per-page deletes, pagination via `ExclusiveStartKey`; `revokeDestroysUserSessions` predicate.
- E2E (local + staging): full PKCE sign-in → mint two reader sessions → revoke → every session dead, revoked bearer can't re-mint. The staging flow uses a dedicated `oauth-revoke-e2e@example.com` user so destroying its sessions can't sign out concurrently running suites or a human using the shared staging account.
- iOS: both logout paths assert the WebKit wipe closure is invoked alongside the existing cookie-jar assertions.

https://claude.ai/code/session_0157WZ5dA8MJubodtneb6dFy


FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: db274c11d744783d157f2de2e89df8d426659223

FagnerMartinsBrack:
### Review: Destroy all user sessions on OAuth token revoke

### 1. `destroyUserSessions` queries a `userId-index` GSI that does not exist on the sessions table

`destroyUserSessions` (`projects/hutch/src/runtime/providers/auth/dynamodb-auth.ts:267-281`) runs `sessions.query({ IndexName: "userId-index", ... })`, but the `hutch-sessions` table is provisioned with **no GSIs** — only `hashKey: "sessionId"` and a single declared attribute (`projects/hutch/src/infra/hutch-storage.ts:135-144`). `userId-index` exists only on the **users** (`:110`), **oauth** (`:156`), and **inbox-addresses** (`:221`) tables. Every prior access to the sessions table is by primary key, so this is the first userId-keyed query and the index was never added.

**Impact:** against real DynamoDB this throws `ValidationException: The table does not have the specified index: userId-index`. The call is `await`ed in the `/oauth/revoke` handler (`oauth.routes.ts:315-317`) *after* the token is already revoked, so the handler rejects (500/hang) instead of returning `200` — breaking the iOS sign-out path and RFC 7009 compliance, and no session is ever destroyed. CI passes only because the integration test uses the in-memory provider and the unit test uses a fake client that ignores index existence.

**Fix:** add the `userId` attribute and a `userId-index` GSI to the sessions table with `projectionType: "ALL"` (required — the query helper parses every item through `SessionRow`, which needs `userId`/`expiresAt`/`emailVerified`; a `KEYS_ONLY` projection would make `schema.parse` throw, the same gotcha documented at `dynamodb-auth.ts:360-362`). Add a test asserting the index is provisioned.

## Medium Priority Issues

### 1. Bulk-destroy by `userId` also kills the user's browser/web sessions
Browser login flows (`auth.page.ts:207`, `google-auth.page.ts`) write to the same sessions table via `createSession` under the same `userId`, so `destroyUserSessions(userId)` logs the user out everywhere, not just iOS reader sessions. The PR frames this as iOS reader-session cleanup and it's asymmetric with browser `/logout`. Confirm "sign out all devices" is intended and document it in the revoke handler. (Only manifests once the High issue is fixed.)

### 2. Test coverage doesn't exercise the real query path
The integration test uses the in-memory `Map`-based provider (can't catch a missing index); the unit test's `createQueryFakeClient` returns no `LastEvaluatedKey`, so the `do…while` pagination loop never runs more than once. Add an infra assertion that the sessions table provisions `userId-index`, and a fake-client case with a `LastEvaluatedKey` to cover pagination.

## Low Priority Suggestions

### 1. Sequential deletes inside the pagination loop — `Promise.all` the per-page deletes to cut sign-out latency for users with many accumulated sessions.

### 2. iOS reader cookie now persists on disk + non-awaited wipe — `ReaderWebView` switched to `.default()` (on-disk, process-wide) and `forceLogout()` fires the wipe without awaiting before flipping `isLoggedIn`. Mitigated by per-open re-injection and server-side destroy, but a slightly weaker posture than the prior `.nonPersistent()`.

### 3. GSI eventual consistency — a just-minted reader session may not be visible to the revoke query immediately and could briefly survive sign-out (TTL still bounds it). Matches the documented lag at `dynamodb-inbox-address.ts:41`.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #866
- **Branch**: claude/ios-changelog-dismissal-4j35io

## Review to Address

FagnerMartinsBrack:

I reviewed the comment changes in this PR against the new [`code-comments` skill](https://github.com/Readplace/readplace.com/blob/main/.claude/skills/code-comments/SKILL.md) (now the single source of truth for comment policy: a comment may exist only for a *why* neither the code nor the commit/PR history can carry, or an approved hack — and any added/kept comment needs explicit human approval).

**Found 1 comment(s) to fix** (and 8 that look compliant with the skill).

### `src/packages/provider-contracts/src/auth.ts`

> /** Destroy every live session for a user (queried via the sessions `userId-index`). iOS mints a fresh session per reader open and holds none of their ids, so a sign-out can only guarantee no live ses

- **Fix:** Drop the 'queried via the sessions userId-index' clause. This is a storage/implementation 'how' (a DynamoDB GSI name) leaking into an implementation-agnostic contract type in provider-contracts; it rots if the lookup mechanism changes. Keep only the second sentence — the genuine why (iOS mints a session per reader open and holds none of their ids) — which the type signature cannot carry.

- **Do not add any new comments.** For any comment worth keeping, leave it and say so, noting it needs explicit human approval — don't invent new rationale.

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: 30e5ef9c82e8be1fda26374935deef59268feda0

FagnerMartinsBrack:
**The prior review cycle's fixes never landed on this branch.** The tip is still the original `db274c11` plus a `main` merge (`git log origin/main..HEAD` shows only those two commits), and `projects/hutch/src/infra/hutch-storage.ts` is untouched. The HIGH issue from the last review is fully intact, and the `code-comments` trim also did not apply.

### 1. `destroyUserSessions` queries a `userId-index` GSI that still does not exist on the sessions table
`destroyUserSessions` (`dynamodb-auth.ts:283-288`) runs `sessions.query({ IndexName: "userId-index", … })`, but the `hutch-sessions` table (`hutch-storage.ts:135-144`) is still `hashKey: "sessionId"` with attributes `[{ sessionId }]` only and **no `globalSecondaryIndexes`**. `userId-index` exists on the users (`:110`), oauth (`:156`), and inbox-addresses (`:221`) tables — never on sessions.

**Impact:** against real DynamoDB this throws `ValidationException: The table does not have the specified index: userId-index`. It is `await`ed in the revoke handler (`oauth.routes.ts`, ~L312-315) *after* the token is revoked and *before* `res.status(200)`, so `/oauth/revoke` returns 500 instead of 200 — this breaks the iOS sign-out path and RFC 7009 compliance, and destroys zero sessions (the PR's whole purpose fails). CI is green only because the integration test uses the in-memory `Map` provider and the unit test's fake client ignores index existence.

**Fix:** add `{ name: "userId", type: "S" }` to the sessions-table attributes and `globalSecondaryIndexes: [{ name: "userId-index", hashKey: "userId", projectionType: "ALL" }]`. Projection **must be `ALL`**: `defineDynamoTable.query` parses every item through `SessionRow` (`session-row.ts:16-21`), which requires `expiresAt` (`z.number()`) and `emailVerified`; a `KEYS_ONLY`/`INCLUDE` projection makes `schema.parse` throw — the same gotcha documented at `dynamodb-auth.ts:374-375` for `userIdPrefix-index`. Add a test asserting the index is provisioned.

## Medium Priority Issues

### 1. Cross-surface "sign out everywhere" is only half-documented
The revoke-handler comment (`oauth.routes.ts`) explains why iOS bulk-destroys and notes browser `/logout` stays single-session, but never states the actual cross-device effect: browser logins write to the **same** sessions table under the **same** `userId` (`auth.page.ts` → `createSession`), so revoking from iOS also terminates the user's live **desktop browser** session. Confirm sign-out-everywhere is intended and say so explicitly in the comment. (Only manifests once the High issue is fixed.)

### 2. Tests still don't exercise the real failure modes
Nothing asserts the sessions table provisions `userId-index` — which is exactly why this prod-breaking bug has now survived two review cycles invisibly. The auto-apply note said a Pulumi `setMocks` unit test isn't reachable; fair, but the gap is real — an integration/LocalStack assertion or a deploy-time check would close it. Separately, the pagination multi-page path is untested: `createQueryFakeClient` (`dynamodb-auth.test.ts:31-49`) returns no `LastEvaluatedKey`, so the `do…while (exclusiveStartKey)` loop never iterates twice. Add a fake-client case returning a `LastEvaluatedKey`.

## Low Priority Suggestions

### 1. Sequential deletes inside the loop
`destroyUserSessions` awaits each `sessions.delete` one at a time. `await Promise.all(items.map((s) => sessions.delete({ Key: { sessionId: s.sessionId } })))` per page cuts sign-out latency for users with many accumulated sessions. (Note: `userId-index` is a GSI, so a just-minted session can briefly lag the revoke query and survive sign-out — TTL still bounds it, matching the documented lag at `dynamodb-inbox-address.ts`.)

### 2. iOS reader cookie now persists on disk + non-awaited wipe on forced logout
`ReaderWebView` switched to `.default()` (on-disk, process-wide) and `forceLogout()` fires `wipeReaderSessionCookie {}` without awaiting before flipping `isLoggedIn`. The intended backstop — the server-side destroy on revoke — is precisely what the High issue disables, so until that lands an iOS sign-out can leave both the server session live and `hutch_sid` on disk briefly. Reasonable tradeoff for changelog persistence, just weaker than the prior `.nonPersistent()`.

### 3. Comment-policy leak still present
The `DestroyUserSessions` contract comment (`provider-contracts/src/auth.ts:56-58`) still carries the storage "how" clause *"(queried via the sessions `userId-index`)"* the earlier `code-comments` pass flagged for removal — that trim also did not land. It names a DynamoDB GSI inside an implementation-agnostic contract type; drop the clause and keep only the why.

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **PR**: #866
- **Branch**: claude/ios-changelog-dismissal-4j35io

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: d0817506d45453a918062090d836db0491270053

FagnerMartinsBrack:
### Re-review: Destroy all user sessions on OAuth token revoke

Reviewed at commit `d0817506` (diff against `origin/main`). **The prior cycle's fixes are now on the branch** — `git log origin/main..HEAD` includes `aa1c0ded` and the diff against `origin/main` carries `hutch-storage.ts | 15 ++++-`. The previously-blocking HIGH issue is resolved; all MEDIUM items are addressed. No new blocking issues.

None found — the `userId-index` GSI now exists.

- **Verified:** `hutch-storage.ts:139-152` adds `{ name: "userId", type: "S" }` and `globalSecondaryIndexes: [{ name: "userId-index", hashKey: "userId", projectionType: "ALL" }]` to the `hutch-sessions` table (read at the current tip, not just the diff).
- **Projection is correct.** `defineDynamoTable.query` (`define-table.ts:130-138`) parses every item through `SessionRow` (`session-row.ts:16-21`), which requires `expiresAt`/`emailVerified` — so `KEYS_ONLY`/`INCLUDE` would have thrown. `ALL` is required and present.
- **No backfill gap.** `createSession` writes `userId` on every row.

## Medium Priority Issues

None found.

- **Cross-surface effect documented** — `oauth.routes.ts:311-316` states the by-`userId` delete is intentionally sign-out-everywhere; browser `POST /logout` keeps single-session destroy.
- **Pagination exercised** — `dynamodb-auth.test.ts` `createPaginatedQueryFakeClient` asserts the `do…while` loop iterates twice and feeds `LastEvaluatedKey` back as `ExclusiveStartKey`, deleting across both pages.
- **Wiring complete** through `provider-contracts → providers → server.ts / test-app.ts → bundle.types.ts → oauth.routes`.
- **Boundary validation clean** — `revokedUserId: unknown` validated via `UserIdSchema.parse(...)`, no `as`.

## Low Priority Suggestions

### 1. Sequential deletes inside the pagination loop
`destroyUserSessions` (`dynamodb-auth.ts:280-293`) awaits each `sessions.delete` one at a time; `await Promise.all(items.map(...))` per page is a cheap sign-out latency win.

### 2. "Sign out everywhere" is session-scoped, not token-scoped
The model exposes `revokeAllUserTokens(userId)` (`oauth.ts:10`) but the handler doesn't call it — it revokes only the presented token (RFC 7009) and destroys all sessions, so a second device's refresh token survives and re-mints a session. Matches the PR's cleanup purpose; flagging in case true all-device logout was intended.

### 3. GSI eventual consistency
A just-minted reader session can briefly lag the revoke query and survive sign-out (TTL bounds it). A one-line note would mirror `dynamodb-inbox-address.ts`.

### 4. iOS reader cookie persists on disk + non-awaited wipe on forced logout
`ReaderWebView` → `.default()` (on-disk) and `forceLogout()` fires the wipe without awaiting before flipping `isLoggedIn`. Backstopped by the now-working server-side destroy + per-open re-injection; slightly weaker at-rest posture than the prior `.nonPersistent()`.

### 5. No regression guard that the sessions table provisions `userId-index`
`src/infra/**` is excluded from `tsc`/Jest and there's no Pulumi `setMocks` harness, so a unit assertion isn't reachable without new infrastructure. Production risk is resolved (validated at deploy). Known gap, not a blocker.

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: fcef38e1e56ae18c0ad456e8ef7e4f75956fca7e

FagnerMartinsBrack:
### Re-review: Destroy all user sessions on OAuth token revoke

Reviewed at `fcef38e1` (diff vs `origin/main`). The branch was **force-pushed** since the last cycle — the earlier SHAs (`aa1c0ded`/`d0817506`) are gone, replaced by `cbb433be → 65291eed → b2f5c14a → fcef38e1` — but the tree is equivalent to the previously-approved state. I re-verified every prior HIGH/MEDIUM fix is present **at the current tip** (read the files directly, not just the diff), and traced the query path into the infra + query helper.

None found — the `userId-index` GSI is now provisioned.

- **GSI present & correct** (`hutch-storage.ts:139-152`): `hutch-sessions` now declares `{ name: "userId", type: "S" }` and `globalSecondaryIndexes: [{ name: "userId-index", hashKey: "userId", projectionType: "ALL" }]`, mirroring the users (`:108`) and oauth (`:167`) tables.
- **Projection `ALL` is required, not optional.** `defineDynamoTable.query` (`define-table.ts:134`) maps every returned item through `schema.parse`, and `SessionRow` (`session-row.ts:16-21`) requires `sessionId`/`userId`/`expiresAt`/`emailVerified` — a `KEYS_ONLY`/`INCLUDE` projection would make `parse` throw. `ALL` is present.
- **Non-destructive rollout.** Adding one attribute + one GSI to an existing PAY_PER_REQUEST table is an in-place `UpdateTable`; `createSession` already writes `userId` on every row, so the index backfills with no data gap.
- **Delete key resolves.** `query` returns parsed `items` (each carrying `sessionId` from `SessionRow`), so `sessions.delete({ Key: { sessionId } })` is well-formed.

## Medium Priority Issues

None found.

- **Cross-surface effect documented** (`oauth.routes.ts:311-316`): the by-`userId` delete is stated to be intentionally sign-out-everywhere (also clears browser/desktop `hutch_sid`), while browser `POST /logout` keeps its single-session destroy.
- **Boundary validation clean** (`oauth.routes.ts:294,318`): `revokedUserId: unknown` is validated with `UserIdSchema.parse(...)` — Zod-at-the-boundary, no `as`.
- **Pagination exercised** (`dynamodb-auth.test.ts`): `createPaginatedQueryFakeClient` returns a `LastEvaluatedKey` on page 1 and asserts the `do…while` loop iterates twice, feeds the key back as `ExclusiveStartKey`, and deletes across both pages — plus empty-user and single-page cases.
- **Wiring complete** through `provider-contracts → {dynamodb,in-memory} providers → server.ts / test-app.ts → bundle.types.ts → oauth.routes`; both providers return `destroyUserSessions`.
- **iOS wipe is main-thread-safe & name-scoped.** `AppSession` is `@MainActor` and `onSessionExpired` is invoked from the `@MainActor` `ReadingListViewModel` (`:214`), so `forceLogout()`'s `WKWebsiteDataStore.default()` access is on the main thread. `deleteReaderSessionCookieFromWebKit` deletes only `hutch_sid` (`= AppConfig.sessionCookieName`), preserving `rp_changelog_dismissed`; the `DispatchGroup` fires `completion` immediately when zero cookies match, so `logout()`'s awaited wipe never hangs.

## Low Priority Suggestions

(All carried from prior cycles — none blocking.)

### 1. Sequential deletes inside the pagination loop
`destroyUserSessions` (`dynamodb-auth.ts:280-292`) awaits each `sessions.delete` one at a time. `await Promise.all(items.map((s) => sessions.delete({ Key: { sessionId: s.sessionId } })))` per page cuts sign-out latency for users with many accumulated sessions.

### 2. "Sign out everywhere" is session-scoped, not token-scoped
The handler revokes only the presented token (correct per RFC 7009) and destroys all *sessions*, but doesn't call `revokeAllUserTokens` (`provider-contracts/src/oauth.ts`). A second device's still-valid refresh token survives and re-mints a `hutch_sid` on its next reader open — consistent with the PR's cleanup purpose, just noting the framing doesn't extend to OAuth tokens.

### 3. GSI eventual consistency
`userId-index` is a GSI, so a just-minted reader session can briefly lag the revoke query and survive sign-out (the 7-day TTL still bounds it). A one-line note here would mirror the documented lag at `dynamodb-inbox-address.ts`.

### 4. iOS reader cookie now persists on disk + fire-and-forget wipe on `forceLogout`
`ReaderWebView` → `.default()` puts `hutch_sid` in the on-disk, process-wide store; `forceLogout()` fires the wipe without awaiting before flipping `isLoggedIn`. Backstopped by the now-working server-side destroy + per-open re-injection; a slightly weaker at-rest posture than the prior `.nonPersistent()`, as already noted.

### 5. No regression guard that the sessions table provisions `userId-index`
`src/infra/**` is excluded from `tsc`/Jest and there's no Pulumi `setMocks` harness, so a unit assertion isn't reachable without new infrastructure. The production risk is resolved (GSI provisioned, validated at deploy). Known gap, not a blocker.

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: fc83339abf2493677ff23054d35b0662eae80b3c

FagnerMartinsBrack:
### Re-review: Destroy all user sessions on OAuth token revoke

Reviewed at `fc83339a` (diff vs `origin/main`). The branch was force-pushed again (`094c83db → db8730a6 → c8e8cb97 → 8f63643e → fc83339a`). The new tip commit — *"grant the sessions index and scope revoke sign-out per client"* — meaningfully **improves** on the previously-approved tree and closes the remaining LOW items. I read every changed file at the current tip (not just the diff) and traced the query path into infra, IAM, and the query helper.

None found — the `userId-index` GSI is provisioned **and** the IAM grant to query it is now in place.

- **GSI present** (`hutch-storage.ts:139-152`): `hutch-sessions` declares `{ name: "userId", type: "S" }` + `globalSecondaryIndexes: [{ name: "userId-index", hashKey: "userId", projectionType: "KEYS_ONLY" }]`.
- **Projection changed `ALL` → `KEYS_ONLY`, and it's now correct.** Prior cycles required `ALL` because the query parsed full `SessionRow`s. This commit instead adds a dedicated `sessionKeysByUser` table def with a minimal `z.object({ sessionId: z.string() })` schema (`dynamodb-auth.ts:110-115`). `KEYS_ONLY` projects `sessionId` + `userId`, zod strips the extra `userId`, and only `sessionId` is needed to delete — so `parse` succeeds and the GSI is cheaper (less storage + write amplification). Good call.
- **IAM grant fixed** (`infra/index.ts:174`): `sessionsTable` flipped `includeIndexes: false → true`. Without this the query would `AccessDenied` at runtime even with the GSI present — this is the load-bearing half that the earlier "just add the GSI" fix would still have missed.
- **Delete key resolves**: `query` returns parsed items each carrying `sessionId`, so `sessions.delete({ Key: { sessionId } })` is well-formed. Deleting from the base table while paging the GSI is key-paginated (not snapshot), so no skips/loops; DeleteItem is idempotent anyway.

## Medium Priority Issues

None found — the cross-surface effect is now **scoped, documented, and tested**, and the real query path is now exercised end-to-end.

- **Sign-out-everywhere is scoped per client** (`built-in-clients.ts:59-68`, `oauth.routes.ts:308-319`): only `ios-app` triggers `destroyUserSessions` + `revokeAllUserTokens`; every other client (extensions, dynamically-registered) stays single-token via `revokeToken`. This directly resolves the prior MEDIUM about an iOS revoke silently killing browser sessions, and the `revokeSignsOutEverywhere` predicate is documented with the genuine *why*.
- **Now token-scoped too**: the iOS path also calls `deps.model.revokeAllUserTokens(userId)`, so a second device's still-valid refresh token no longer survives to re-mint a session — closing prior LOW #2.
- **Ordering rationale is explicit** (`oauth.routes.ts:312-313`): sessions-before-tokens so a failed destroy leaves the presented token valid and the client's retry re-runs the whole sign-out. `revokedUserId` is validated with `UserIdSchema.parse(...)` — Zod-at-the-boundary, no `as`.
- **Real path is tested.** Unit: `dynamodb-auth.test.ts` covers single-page, empty-user, and multi-page pagination (feeds `LastEvaluatedKey` back as `ExclusiveStartKey`, deletes across both pages); `oauth.routes.test.ts` asserts both directions (iOS revoke nukes all sessions + both tokens; extension revoke keeps sessions + the iOS bystander token); `built-in-clients.test.ts` pins the predicate. E2E: `oauth-revoke-flow.ts` drives the full HTTP contract (signup → PKCE authorize → token → mint 2 reader sessions → revoke → prove web + both reader sessions are dead + the bearer can't re-mint), wired into **staging** (`run.e2e-staging.ts`, serial) so it exercises the deployed GSI **and IAM grant** — the regression guard prior cycles said was unreachable by unit tests.
- **`forEachQueryPage` extraction** (`define-table.ts:229-249`) de-dupes the pagination loop shared by `destroyUserSessions` and `revokeAllUserTokens`, and both now `Promise.all` deletes per page — closing prior LOW #1 (sequential deletes).
- **iOS wipe verified.** `AppSession` is `@MainActor` (`AppSession.swift:21`), so `WKWebsiteDataStore.default()` access in the static `removeReaderWebStoreData` runs on the main thread. The session cookie is deleted **by name** (preserving `rp_changelog_dismissed`), and non-cookie data (cache, local/session storage, IndexedDB) is now also cleared so a signed-out account's reading history doesn't linger on disk — a stronger at-rest posture than the earlier name-only wipe. Both `logout()` and `forceLogout()` are covered by `LoginFlowTests` via an injected wipe spy.
- **Comment-policy leak resolved** (`provider-contracts/src/auth.ts:56`): the `DestroyUserSessions` type is now comment-free — the `userId-index` "how" clause the `code-comments` pass flagged is gone.

## Low Priority Suggestions

### 1. Staging E2E can flake on GSI eventual consistency
`oauth-revoke-flow.ts` mints two reader sessions, then revokes and asserts all sessions return `303`. Against real DynamoDB, `userId-index` is a GSI (eventually consistent), so a just-minted session may not be visible to the revoke query yet and could survive (its liveness check reads by primary key, so it'd still be `200` and fail the assertion). The 3 liveness GETs between mint and revoke make this unlikely, but a bounded poll/retry on the post-revoke assertions would harden the staging run. (TTL still bounds it in production.)

### 2. PR description is stale vs the code
Two real behavior changes in this diff aren't in the PR body: (a) **password reset now destroys all of the user's sessions** (`forgot-password.page.ts:155-158`, tested) — good security hardening, but unlisted; and (b) an **iOS sign-out now revokes every OAuth token the user holds, including their browser-extension tokens** (broad blast radius; intentional and documented in `built-in-clients.ts`, but worth a human's eyes). Please update the summary so reviewers and any changelog capture both.

### 3. `forceLogout()` wipe remains fire-and-forget
`forceLogout()` now returns a `@discardableResult Task` (so tests can await it) and the wipe clears non-cookie data too — both improvements. Production still discards the task before flipping `isLoggedIn`, but it's called only when the session is already invalid and is backstopped by the working server-side destroy + per-open re-injection, so the residual window is benign.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28606033859
- **PR**: #866
- **Branch**: claude/ios-changelog-dismissal-4j35io

## Failed Jobs

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: 4a20c3c5bc98e84fd18440c6da3d641d645d2fe2

FagnerMartinsBrack:
### Re-review: Persist iOS reader state; destroy sessions on iOS sign-out

Reviewed at `4a20c3c5` (diff vs `origin/main`). Two commits landed since the last clean review at `fc83339a`:
- `2809aeea` — scope iOS sign-out to sessions + the presented token (a behavior change)
- `4a20c3c5` — CI fix: Makefile test-retry flags

I read the changed files at the current tip and re-verified the prior HIGH/MEDIUM fixes are still present.

None found.

- **Revoke handler is correct and simpler** (`oauth.routes.ts:308-321`): iOS (`revokeDestroysUserSessions`) → `destroyUserSessions(userId)` then `revokeToken(refreshRecord)`; every other client → `revokeToken(refreshRecord)` only. `revokeToken` moved out of the old `else` so it runs on every path — the presented token is always revoked (RFC 7009), and only iOS additionally bulk-destroys sessions. Ordering is sessions-before-token so a failed destroy leaves the presented token valid for the client's retry. `userId` validated via `UserIdSchema.parse` (no `as`). All four branches (record/no-record × iOS/other) are covered, including the unknown-token path (`:548`).
- **Prior fixes intact at tip**: `userId-index` GSI (`KEYS_ONLY`) + `includeIndexes: true` IAM grant, `forEachQueryPage` pagination with per-page `Promise.all` deletes, password-reset session destroy + ghost-account 400.

## Medium Priority Issues

None found.

- **Tests match the new behavior**: `oauth.routes.test.ts` now asserts the extension bystander token *survives* an iOS revoke (`assert(bystander, …)` + `user.id` check) while the iOS presented token + all sessions die; `built-in-clients.test.ts` renamed to `revokeDestroysUserSessions` and pins ios-app=true, extensions/dynamic=false.
- **Naming true-up is accurate**: `revokeSignsOutEverywhere` → `revokeDestroysUserSessions` reflects that tokens no longer die everywhere (other devices re-mint). The predicate comment carries the genuine *why*.
- **iOS comment true-ups are correct**: the chromeless reader renders no banners, so the persisted state is the share-hint localStorage dismissal; the sign-out wipe deletes the session cookie *by name* and clears non-cookie data, deliberately resetting that dismissal. Both logout paths remain covered by the injected wipe spy.

## Low Priority Suggestions

### 1. `revokeAllUserTokens` has no production caller (same as `main`)
`2809aeea` deliberately unwired `revokeAllUserTokens` from the iOS revoke path (so other devices' tokens survive). The method stays on the `OAuthModel` contract with both impls + provider tests, but no route calls it — identical to `origin/main` (caller-less since #862), so this PR introduces **no regression** and knip stays green. Flagging only that the capability is now dormant, in case a future "hard sign-out everywhere" was meant to use it.

### 2. Suite-wide iOS test retry (`Makefile`)
`4a20c3c5` adds `-retry-tests-on-failure -test-iterations 3` to the whole `make test` run to absorb the first-launch simulator flake. `-retry-tests-on-failure` re-runs only failed tests and stops on first success, so passing tests still run once and a deterministic failure still fails all 3 iterations — it doesn't mask hard failures. Note it's suite-wide (not scoped to the known-flaky `ReadplaceAPITests` case), so a genuinely flaky *product* test added later would be silently retried too; the rationale lives in the commit body per the repo's comment policy.

### 3. Staging revoke E2E can flake on GSI eventual consistency
`oauth-revoke-flow.ts` revokes, then asserts every session is dead; `userId-index` is eventually consistent, so a just-minted session could briefly lag the revoke query. The dedicated `oauth-revoke-e2e@example.com` user removes the cross-suite blast radius, but a bounded poll/retry on the post-revoke assertions would harden the staging run. (TTL bounds it in prod.)

### 4. `forceLogout()` wipe stays fire-and-forget
Returns a `@discardableResult Task`; production discards it before flipping `isLoggedIn`. Called only when the session is already invalid and backstopped by the server-side destroy + per-open re-injection, so the residual window is benign — carried from prior cycles.

FagnerMartinsBrack:
@claude address these low prio issues and don't report the other 2 again:

1. revokeAllUserTokens has no production caller (same as main)
2809aeea deliberately unwired revokeAllUserTokens from the iOS revoke path (so other devices' tokens survive). The method stays on the OAuthModel contract with both impls + provider tests, but no route calls it — identical to origin/main (caller-less since https://github.com/Readplace/readplace.com/pull/862), so this PR introduces no regression and knip stays green. Flagging only that the capability is now dormant, in case a future "hard sign-out everywhere" was meant to use it.

2. Suite-wide iOS test retry (Makefile)
4a20c3c5 adds -retry-tests-on-failure -test-iterations 3 to the whole make test run to absorb the first-launch simulator flake. -retry-tests-on-failure re-runs only failed tests and stops on first success, so passing tests still run once and a deterministic failure still fails all 3 iterations — it doesn't mask hard failures. Note it's suite-wide (not scoped to the known-flaky ReadplaceAPITests case), so a genuinely flaky product test added later would be silently retried too; the rationale lives in the commit body per the repo's comment policy.

FagnerMartinsBrack:

## Context
- **PR**: #866
- **Commit**: 2dfac194ce080f7c22717bbcd1375bc3e0cd9cf1

FagnerMartinsBrack:
### Re-review: Persist iOS reader state; destroy sessions on iOS sign-out

Reviewed at `2dfac194` (diff vs `origin/main`). Two commits landed since the last clean review at `4a20c3c5`, both closing the LOW items you asked to address:
- `c79e7470` — drop the caller-less `revokeAllUserTokens` capability
- `2dfac194` — pre-boot the iOS simulator instead of retrying the whole suite

CI run [`28632628637`](https://github.com/Readplace/readplace.com/actions/runs/28632628637) on this exact tip is **green** — `ios-tests` ✅ and `web-tests` ✅ — which is the real validation for both commits.

None found.

- **Prior HIGH fix intact at tip.** `hutch-sessions` still declares the `userId` attribute + `userId-index` GSI (`KEYS_ONLY`) (`hutch-storage.ts:139-149`), the IAM grant is `includeIndexes: true` (`infra/index.ts:174`), and `destroyUserSessions` queries it through the dedicated `sessionKeysByUser` `KEYS_ONLY` schema (`dynamodb-auth.ts:287-299`). The revoke handler still destroys sessions **before** revoking the token and validates `userId` via `UserIdSchema.parse` (`oauth.routes.ts:308-319`).

## Medium Priority Issues

None found.

- **`revokeAllUserTokens` removal is complete and clean** (`c79e7470`). Gone from the `OAuthModel` contract, both provider impls, and the shared contract tests; **no lingering references repo-wide**. The now-orphaned `forEachQueryPage`/`UserId` imports, `RevokeItemRow` schema, and `revokeView` table def came out with it — while `forEachQueryPage` stays used by `destroyUserSessions`, and the in-memory `userIdIndex` is still maintained by `saveToken`/`revokeToken` (`oauth-model.ts:288-289`), so no dead bookkeeping is left behind. `web-tests` green confirms tsc + biome + knip + coverage (functions **100%**) all pass with the method and its 3 tests removed together.
- **iOS pre-boot is the right trade and is validated** (`2dfac194`). `ios-tests` passed on this tip with the blanket `-retry-tests-on-failure -test-iterations 3` **gone**, so `simctl bootstatus -b` absorbs the first-launch cold-start race and every product test now runs **exactly once** — a genuinely flaky *product* test would fail the run instead of being silently retried away, which was the concern behind the request. Ordering is correct: pre-boot precedes `xcodebuild test`, and the subsequent `test-staging` sub-make reuses the already-booted device.

## Low Priority Suggestions

### 1. `SIM_DEVICE` extraction is correct for CI; local `name=` form leans on simctl name-resolution
`SIM_DEVICE = $(shell printf '%s' '$(SIM)' | sed -E 's/.*=//')` greedily strips to the last `=`, so CI's `SIM="platform=iOS Simulator,id=$UDID"` (`ci.yml:321`) yields the bare UDID and `simctl bootstatus '<UDID>' -b` boots unambiguously — the path that's now green. The local default (`name=iPhone NN`) yields the device name, which `bootstatus` resolves the same way `xcodebuild -destination` already does, so **no new ambiguity** is introduced. Nothing to change; noted for the record.

*(Previously-raised LOWs — staging-revoke E2E GSI eventual-consistency and `forceLogout()` fire-and-forget — intentionally omitted per your request not to re-report them.)*

PR #880, opened 2026-07-01
Centralize locale constant across date formatting modules

FagnerMartinsBrack:
## Summary
Extracted the hardcoded locale string into a shared constant to ensure consistent date formatting across the application and reduce duplication.

## Changes
- Exported `LOCALE` constant from `local-time.format.ts` to make it reusable across modules
- Updated `trial-countdown.format.ts` to import and use the shared `LOCALE` constant instead of the hardcoded `"en-US"` string

## Implementation Details
The `LOCALE` constant (`"en-US"`) is now defined in a single location (`local-time.format.ts`) and imported where needed. This ensures that any future changes to the application's locale preference only need to be updated in one place, improving maintainability and reducing the risk of inconsistent formatting across date-related features.

https://claude.ai/code/session_018NPG6UwR8zy6xzzfFztJPH

### DIFF
--- a/src/packages/web-shell/src/local-time.format.ts
+++ b/src/packages/web-shell/src/local-time.format.ts
@@ -17,7 +17,7 @@ export interface LocalTime {
 
 export type LocalTimeStyle = "datetime" | "date";
 
-const LOCALE = "en-US";
+export const LOCALE = "en-US";
 
 /** Explicit component options rather than dateStyle/timeStyle, which cannot be
  * combined with timeZoneName. hour12:false yields a 24-hour clock (09:00, not
--- a/src/packages/web-shell/src/trial-countdown.format.ts
+++ b/src/packages/web-shell/src/trial-countdown.format.ts
@@ -1,3 +1,5 @@
+import { LOCALE } from "./local-time.format";
+
 export interface TrialRemaining {
 	days: number;
 	hours: number;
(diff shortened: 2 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #880
- **Commit**: 1716cf68eb6fe3834ce214bd76b1a2b7b74667f9

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. "Single source of truth" is partial — two other `"en-US"` literals remain

The PR's stated goal is that a future locale change "only need[s] to be updated in one place." Two date-formatting `"en-US"` literals still exist after this change:

- **`src/packages/web-shell/src/trial-countdown.format.test.ts:129`** — the test computes its expected value with a hardcoded `"en-US"`. **Leaving this hardcoded is the right call**, not a miss: an independent literal makes the test a genuine oracle that would catch an accidental change to `LOCALE`, whereas importing the constant would make the assertion self-referential (it would pass no matter what `LOCALE` became). Worth being aware that the "one place" claim is intentionally not literally true here.
- **`projects/blog-site/src/runtime/web/pages/blog/blog.posts.ts:101`** — a separate project. It can't import `web-shell`'s constant without violating the repo's "No Cross-Project Relative Imports" rule; genuinely sharing the value would require extracting it into a workspace package. Reasonably out of scope for this PR, but it means the app-wide locale is still duplicated across projects.

No action required — just calibrating the description so a future reader doesn't assume the whole codebase now routes through this constant.

### 2. Minor cross-module coupling: `trial-countdown.format` now depends on `local-time.format`

`trial-countdown.format.ts` now imports from `local-time.format.ts` solely to borrow a constant; the two are otherwise-independent formatters. This is acceptable — both are `*.format.ts` siblings in the same package and `local-time.format.ts` is dependency-free, so the risk is low. If more shared i18n constants accumulate, a dedicated `locale.ts` would read more clearly than sourcing them from the local-time module. Not necessary for this change.

PR #886, opened 2026-07-01
fix(summary): stop doubling receiveCount in exhausted-retries parse-error

FagnerMartinsBrack:
## What

Every DLQ-exhausted summary shows up on the errors dashboard as:

```
summary-generation-failed: exhausted-retries (receiveCount=4) (receiveCount=4)
```

The `(receiveCount=N)` phrase is appended **twice**. This PR makes it appear once.

## Why (root cause)

The receive count travels on `SummaryGenerationFailedEvent` as its own numeric field. Two places stamp it into the human-readable reason string:

1. `mark-summary-exhausted.ts` → `reasonAsString({kind:"exhausted-retries"})` returned `` `exhausted-retries (receiveCount=${n})` `` — count embedded here.
2. `summary-generation-failed-handler.ts:27` — the parse-error sink appends `` ` (receiveCount=${detail.receiveCount})` `` on top of whatever `detail.reason` already is.

Because the generate-summary DLQ handler **always** tags the reason `exhausted-retries` (`generate-summary-dlq-handler.ts:35`), the sink's append always landed on a string that already carried the count → the doubled suffix on **every** line from this source (167 lines over the last 30 days in prod).

## Fix

`reasonAsString` now returns the bare label `"exhausted-retries"` and lets the sink own the single append. Result:

```
summary-generation-failed: exhausted-retries (receiveCount=4)
```

Scope is deliberately minimal — only the published log string changes. The stored `summary.reason` **object** (which still carries the structured `receiveCount`) is untouched, and no retry/DLQ behaviour changes.

## Notes

- The parallel crawl-axis formatter (`mark-crawl-exhausted.ts`) embeds the count the same way, but its event (`CrawlArticleFailedEvent`) has **no** parse-error sink that re-appends, so it never produced a doubled string — left as-is to keep this PR scoped.
- This does **not** reduce the *count* of these lines (that's driven by the summary auto-heal re-enqueue loop) — it only corrects the text. Loop/volume is tracked separately.

## Tests

- `mark-summary-exhausted.test.ts` — asserts the bare `"exhausted-retries"` label.
- `summary-generation-failed-handler.test.ts` — new regression test asserting the sink appends `(receiveCount=N)` exactly once for the `exhausted-retries` reason.
- `pnpm check` green (100% coverage).

### DIFF
--- a/projects/save-link/src/runtime/domain/generate-summary/summary-generation-failed-handler.test.ts
+++ b/projects/save-link/src/runtime/domain/generate-summary/summary-generation-failed-handler.test.ts
@@ -30,7 +30,12 @@ function createSqsEvent(detail: { url: string; reason: string; receiveCount: num
 describe("initSummaryGenerationFailedHandler", () => {
 	it("logs a parse-error record with source=generate-summary", async () => {
 		const infoSpy = jest.fn();
-		const parseErrorLogger = { info: infoSpy } as unknown as HutchLogger.Typed<ParseErrorEvent>;
+		const parseErrorLogger: HutchLogger.Typed<ParseErrorEvent> = {
+			info: infoSpy,
+			error: jest.fn(),
+			warn: jest.fn(),
+			debug: jest.fn(),
+		};
 		const now = () => new Date("2026-04-19T12:00:00.000Z");
 
 		const handler = initSummaryGenerationFailedHandler({ parseErrorLogger, logger: noopLogger, now });
@@ -57,7 +62,12 @@ describe("initSummaryGenerationFailedHandler", () => {
 
 	it("appends receiveCount exactly once for the exhausted-retries reason (no doubled (receiveCount=N))", async () => {
 		const infoSpy = jest.fn();
-		const parseErrorLogger = { info: infoSpy } as unknown as HutchLogger.Typed<ParseErrorEvent>;
+		const parseErrorLogger: HutchLogger.Typed<ParseErrorEvent> = {
+			info: infoSpy,
+			error: jest.fn(),
+			warn: jest.fn(),
+			debug: jest.fn(),
+		};
 		const now = () => new Date("2026-04-19T12:00:00.000Z");
 
 		const handler = initSummaryGenerationFailedHandler({ parseErrorLogger, logger: noopLogger, now });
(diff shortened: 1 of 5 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #886
- **Commit**: 3647769d4f4a4e2d78e49d0fb2cfa7816f712bd2

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Two sibling `reasonAsString` formatters now diverge on `exhausted-retries`

After this change, `mark-summary-exhausted.ts:14` returns the bare `"exhausted-retries"`, while its near-identical sibling `mark-crawl-exhausted.ts:20` still returns `` `exhausted-retries (receiveCount=${reason.receiveCount})` ``.

The asymmetry is **correct** — only `SummaryGenerationFailedEvent` has a re-appending parse-error sink (`summary-generation-failed-handler.ts:27`); I grepped the tree and confirmed `CrawlArticleFailedEvent` has no handler that reads `detail.reason`, so the crawl axis never doubled. The concern is purely latent maintainability: the two functions read like copy-paste siblings, so a future "harmonize these" refactor could reintroduce the double on the summary axis or drop the count on the crawl axis. Both sides are already pinned by tests (the summary test name even spells out the reason), which is this repo's preferred guard — so this is genuinely optional. A one-line *why* comment on the summary `exhausted-retries` case would make the asymmetry self-documenting, subject to the repo's comment-approval gate.

FagnerMartinsBrack:
@claude Address the following on this branch:

1. **Normalize `mark-crawl-exhausted.ts` the same way this PR normalized the summary axis.** Its `reasonAsString` still returns `` `exhausted-retries (receiveCount=${reason.receiveCount})` `` while the `publish-crawl-article-failed` effect already carries `receiveCount` as a structured field — the exact double-encoding shape this PR removed from `mark-summary-exhausted.ts`. No sink appends the count on the crawl axis today, so nothing visible changes, but normalizing (a) unifies the `exhausted-retries` reason vocabulary across the two axes for log queries, and (b) prevents this PR's bug from reappearing if a `CrawlArticleFailedEvent` logging sink modeled on the summary one is ever added. Update `mark-crawl-exhausted.test.ts` to pin the bare label, mirroring the summary test.

2. **Drop the `as unknown as HutchLogger.Typed<ParseErrorEvent>` double assertion** the new test copies at `summary-generation-failed-handler.test.ts:60`. CLAUDE.md's "Avoid TypeScript Type Assertions (`as`)" rule explicitly lists "Faking a partial object with `as`" as BAD, and neither allowed exception applies. `HutchLogger.Typed` requires only `info`/`error`/`warn`/`debug`, so a fully-typed fake — `{ info: infoSpy, error: jest.fn(), warn: jest.fn(), debug: jest.fn() }` — compiles with no assertion. Fix the two pre-existing instances of the same pattern in this file too so it has one idiom.

Re the review's **Low #1** (add a why-comment for the formatter asymmetry): declined per the repo comment policy — the why is already carried by the test name and this PR's description, and item 1 above removes the asymmetry altogether, which is the better resolution than documenting it.

Verify with `pnpm check`; commit to this branch.

FagnerMartinsBrack:

## Context
- **PR**: #886
- **Commit**: 786e429b17b1fae6b7014e9fe5b721a97a72be94

PR #889, opened 2026-07-01
fix(crawl): follow curl redirects (SSRF-guarded) + back off failing stale-checks

FagnerMartinsBrack:
Two fixes for the `fetchCurl` error category (#3 in the 30-day dashboard audit). ENOENT is a documented follow-up (see bottom).

---

## Commit 1 — follow curl-impersonate redirects through the SSRF guard

**Why.** The curl fallback hardcoded `--location --max-redirs 0`, so any origin answering with a 3xx made curl exit 47 (`Maximum (0) redirects followed`) and the crawl failed. That was ~**59** of the fetchCurl failures over 30 days (archive.ph 46×, developer.android.com, dev.w3.org) — sites that block the Node/undici path (forcing the curl fallback) *and* redirect.

`--location` refused redirects deliberately: curl does its own DNS, so following a redirect could reach an unchecked private IP, bypassing the Node-level SSRF guard.

**How.** `fetchCurl` now follows redirects itself, re-applying the **same** guard per hop:
- `--location` dropped; `--max-redirs 0` kept so curl never chases anything on its own.
- A 3xx is returned to `fetchCurl`, which resolves the `Location` (relative against the current hop), runs `resolvePinnedAddress` on the **new host** (the existing SSRF guard — blocks private/❌ addresses), and re-pins curl to that validated address for the next hop.
- Bounded to 5 hops; a loop fails with `too many redirects`.

The SSRF invariant is preserved: every curl spawn still connects only to a `resolvePinnedAddress`-validated IP — now applied to each hop instead of only the first.

Also removes the dead persona-fallback signature `"maximum (5) redirects followed"` (it never matched the real `(0)` message, and curl no longer exits 47 for redirects). The undici `max_redirects` signature stays.

## Commit 2 — back off on refresh failure instead of re-crawling every /view

**Why.** A failed stale-check refresh returned `"skip"` **without** advancing `contentFetchedAt`, so a permanently-broken origin stayed past the 24h staleness gate and was re-crawled on **every** `/view`. That is what turned a handful of unfetchable URLs into the repeat storms: `www.ics.uci.edu` SSL-failed **73×**, `archive.ph` **46×** over 30 days — the same row hammered on each visit.

**How.** On `result.status === "failed"` the handler publishes `UpdateFetchTimestamp` (resets the freshness clock, carries the existing `bodyHash` forward) and returns a new `"backoff"` action. The next attempt waits a full TTL instead of firing on the next visit, collapsing the per-URL failure count to ~1 per TTL. Existing content on a ready row is untouched; only the redundant re-crawls stop.

## Combined effect on #3

- archive.ph (46×) and other redirecting origins now **crawl successfully** (commit 1) → they stop failing *and* stop being stale → drop off the dashboard.
- ics.uci.edu SSL (73×) and any still-failing origin are **re-crawled ~1×/TTL instead of every visit** (commit 2) → the storms collapse.

## Tests
- curl: follows 301→200, relative-Location resolution, cross-host re-pin via `--resolve`, **SSRF guard refuses a redirect to 169.254.169.254 (no curl spawned for it)**, MAX_REDIRECTS cap, and non-redirect-with-Location is not followed.
- stale-check: failed refresh now asserts the backoff timestamp + `bodyHash` carry-forward + the `backed off` log.
- `pnpm check` green (100% coverage).

## Follow-up NOT in this PR — `ENOENT` (curl binary missing on hutch-handler, ~5×/30d)

hutch-handler wires the same crawl fallback chain (`app.ts` `initCrawlFetch`) but the curl-impersonate **Lambda layer** is defined only in the **save-link** Pulumi stack (`projects/save-link/src/infra`), not attached to the hutch-handler Lambda (`projects/hutch/src/infra`). When a synchronous `/view` / `/import` / `/admin/recrawl` crawl escalates to curl, `execFile("curl_chrome131")` finds nothing → `spawn ENOENT`.

Fixing it means exporting the layer ARN from the save-link stack and consuming it via a **cross-stack StackReference** in the hutch stack (or building the layer in both) — a deploy-ordering-sensitive infra change I can't verify without deploying, so it is intentionally left out of this code PR. Recommend a dedicated infra PR.

### DIFF
--- a/projects/save-link/src/runtime/domain/stale-check/stale-check-handler.test.ts
+++ b/projects/save-link/src/runtime/domain/stale-check/stale-check-handler.test.ts
@@ -284,11 +284,12 @@ describe("initStaleCheckHandler", () => {
 		expect(publishRefreshArticleContent).not.toHaveBeenCalled();
 	});
 
-	it("skips when crawl returns failed (no further effect)", async () => {
+	it("backs off when crawl returns failed: resets the freshness clock (carrying the bodyHash) so the next attempt waits a full TTL", async () => {
 		const findArticleFreshness: FindArticleFreshness = async () => ({
 			etag: undefined,
 			lastModified: undefined,
 			contentFetchedAt: "2026-04-01T00:00:00.000Z",
+			bodyHash: "h".repeat(64),
 		});
 		const crawlAndFinalizeArticle: CrawlAndFinalizeArticle = async () => ({ status: "failed", reason: "crawl-failed" });
 		const publishRefreshArticleContent: PublishRefreshArticleContent = jest
@@ -300,6 +301,7 @@ describe("initStaleCheckHandler", () => {
 		const emitSimpleCrawlUnsupported: EmitSimpleCrawlUnsupported = jest
 			.fn()
 			.mockResolvedValue(undefined);
+		const logger = { info: jest.fn(), warn: jest.fn(), error: jest.fn(), debug: jest.fn() };
 
 		const handler = createHandler({
 			findArticleFreshness,
@@ -307,12 +309,21 @@ describe("initStaleCheckHandler", () => {
 			publishRefreshArticleContent,
 			publishUpdateFetchTimestamp,
 			emitSimpleCrawlUnsupported,
+			logger,
 		});
 
 		await handler(createSqsEvent({ url: URL_UNDER_TEST }), buildLambdaContext(), () => {});
 
+		expect(publishUpdateFetchTimestamp).toHaveBeenCalledTimes(1);
+		expect(publishUpdateFetchTimestamp).toHaveBeenCalledWith({
+			url: URL_UNDER_TEST,
+			contentFetchedAt: fixedNow().toISOString(),
+			bodyHash: "h".repeat(64),
+		});
+		expect(logger.info).toHaveBeenCalledWith("[StaleCheckRequested] backed off after refresh failure", {
+			url: URL_UNDER_TEST,
+		});
 		expect(publishRefreshArticleContent).not.toHaveBeenCalled();
-		expect(publishUpdateFetchTimestamp).not.toHaveBeenCalled();
 		expect(emitSimpleCrawlUnsupported).not.toHaveBeenCalled();
 	});
 
--- a/projects/save-link/src/runtime/domain/stale-check/stale-check-handler.ts
+++ b/projects/save-link/src/runtime/domain/stale-check/stale-check-handler.ts
@@ -34,8 +34,12 @@ import type { CrawlAndFinalizeArticle } from "@packages/finalize-article";
  *
  * Action outcomes:
  *   - `"new"`             — row missing; re-published SaveAnonymousLinkCommand.
- *   - `"skip"`            — within TTL, terminal state, parse failure, or
- *                           simpleCrawl failure. No downstream effect.
+ *   - `"skip"`            — within TTL, terminal state, or parse failure. No
+ *                           downstream effect.
+ *   - `"backoff"`         — the refresh fetch failed (cert error, redirect,
+ *                           bot block); reset the freshness clock so the next
+ *                           attempt waits a full TTL instead of re-firing on
+ *                           every /view.
  *   - `"unchanged"`       — 304 Not Modified; published UpdateFetchTimestamp.
  *   - `"refreshed"`       — HTML refetched + parsed; published RefreshArticleContent.
  *   - `"tier-1-deferred"` — non-HTML body; emitted SimpleCrawlUnsupportedEvent
@@ -46,6 +50,7 @@ import type { CrawlAndFinalizeArticle } from "@packages/finalize-article";
 export type StaleCheckAction =
 	| "new"
 	| "skip"
+	| "backoff"
 	| "unchanged"
 	| "refreshed"
 	| "tier-1-deferred";
@@ -115,7 +120,18 @@ export function initStaleCheckHandler(deps: {
 			return "unchanged";
 		}
 
-		if (result.status === "failed") return "skip";
+		if (result.status === "failed") {
+			/* A permanently-failing origin (cert error, redirect, bot block) would
+			 * otherwise be re-crawled on every /view, since the failed branch never
+			 * advanced contentFetchedAt. Reset the freshness clock so the next
+			 * attempt waits a full TTL instead of firing on the next visit. */
+			await publishUpdateFetchTimestamp({
+				url,
+				contentFetchedAt: now().toISOString(),
+				bodyHash: freshness.bodyHash,
+			});
+			return "backoff";
+		}
 
 		if (result.status === "unsupported") {
 			/* Write the stage marker so the reader's progress bar advances
@@ -176,6 +192,8 @@ export function initStaleCheckHandler(deps: {
 						url: detail.url,
 						action,
 					});
+				} else if (action === "backoff") {
+					logger.info(`${logPrefix} backed off after refresh failure`, { url: detail.url });
 				} else {
 					logger.info(`${logPrefix} no-op`, { url: detail.url, action });
 				}
--- a/src/packages/crawl-article/src/curl-fetch.test.ts
+++ b/src/packages/crawl-article/src/curl-fetch.test.ts
@@ -125,7 +125,7 @@ describe("fetchCurl argument construction", () => {
 		const args = fake.calls[0].args;
 		expect(args).toContain("--http2");
 		expect(args).toContain("--compressed");
-		expect(args).toContain("--location");
+		expect(args).not.toContain("--location");
 		const sepIdx = args.indexOf("--");
 		expect(sepIdx).toBeGreaterThan(0);
 		expect(args[sepIdx + 1]).toBe("https://example.com/page?q=1");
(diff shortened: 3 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #889
- **Commit**: a9a7da528291702c1644e2921dfc50add738dd5d

FagnerMartinsBrack:
### Review: follow curl redirects (SSRF-guarded) + back off failing stale-checks

The design is solid. Re-running `resolvePinnedAddress` per hop (so the SSRF check moves from "first host only" to "every host"), resolving a relative `Location` against the current hop, re-pinning cross-host targets via `--resolve`, and capping at 5 hops are all correct and well-covered by the new tests — especially the negative-space assertion that a blocked redirect target spawns **no** curl (`seq.calls` length 1). Commit 2's backoff is the right fix for the per-`/view` storm and correctly leaves a ready row's existing content untouched while only advancing the freshness clock. One security hardening gap worth addressing before merge.

None found

## Medium Priority Issues

### 1. Redirect loop has no scheme allowlist — the curl fallback will follow `gopher://`/`dict://`/`file://`/… redirects

`src/packages/crawl-article/src/curl-fetch.ts:119` computes the next hop as `new URL(location, currentUrl).href` and hands it straight to `fetchOnce` → curl, with no check on the resulting **scheme**. The SSRF guard (`resolvePinnedAddress`) is purely *IP*-based: it resolves the host and refuses private/loopback/link-local addresses. It is not a *scheme* guard.

That matters because the primary (undici) path only ever follows `http`/`https` redirects, but the curl fallback is now strictly more permissive. I verified against the bundled libcurl behaviour (curl-impersonate is built on the same libcurl):

- **Enabled protocols** include `dict file ftp ftps gopher gophers imap ldap mqtt pop3 rtmp rtsp scp sftp smb smtp telnet tftp`, and a `gopher://host:port/…` URL **dials the authority host** (confirmed: connect attempt → timeout, not a parse error).
- A malicious origin (reachable **logged-out** via self-serve import per `CLAUDE.md`, and able to force the curl fallback by 403-ing undici) can answer with `301 Location: gopher://<public-host>:<port>/<payload>`. The guard resolves `<public-host>`, sees a public IP (not blocked), pins curl to it, and curl speaks gopher — turning the crawler into an **open egress proxy** to arbitrary public host:port over non-HTTP protocols (port-scan/reflection/laundering from our egress IP). This is curl-version-independent.
- `file://` internal-read is *mostly* closed today, but only incidentally: `file:///path` (empty host) is caught by the no-address branch (`dns.lookup("") → []`), `file://localhost/path` is caught because loopback is IP-blocked, and `file://<ip>/path` is rejected by **curl 8.x itself** (`curl: (3) URL rejected: Bad file:// URL`). That last defence is a version-specific libcurl behaviour, not something this code guarantees — an older/differently-built curl-impersonate that reads a non-local `file://` authority would bypass the IP guard entirely (the public IP passes, but curl never connects to it), giving LFI.

Internal SSRF (private IPs / `169.254.169.254`) remains blocked — that invariant *is* preserved. But the PR's framing ("every curl spawn still connects only to a validated IP") doesn't cover `file://` (no connection to the checked IP) or the open-egress-to-public case.

**Fix:** after computing `currentUrl`, refuse to follow unless the target protocol is `http:`/`https:` (throw naming the scheme), matching undici's redirect contract. Cheap, and closes the whole class regardless of curl version/protocol set. A `301 → gopher://…` test asserting refusal + no second curl spawn mirrors the existing SSRF-refusal test.

## Low Priority Suggestions

### 1. Stale `--location` references in the parser doc comments
`curl-fetch.ts:202-203` ("With `--location`, intermediate redirect headers appear before the final ones") and `curl-fetch.ts:232` ("With `--location`, each redirect response has its own header block") now describe a flag this PR removed. The multi-header-block handling in `findLastHeaderBlock` is still worth keeping (HTTP `1xx` interim responses like `100 Continue` / `103 Early Hints` also emit multiple status lines), so update the *rationale* rather than the code.

### 2. Per-hop timeout, no aggregate deadline when no `signal` is passed
`fetchOnce` uses `DEFAULT_TIMEOUT_MS` (10s) only when `init.signal` is absent (`curl-fetch.ts:78`). With app-layer redirect following, a 5-hop chain from a slow origin can now consume up to ~60s of wall-clock on the curl path (vs. 10s before) when the caller supplies no signal. Bounded by `MAX_REDIRECTS`, so minor — but a deliberate origin could pad the crawl budget.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #889
- **Branch**: fix/curl-follow-redirects-ssrf-guarded

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #889
- **Commit**: 7a34dd1d28e6e3ceba2c630261ffb828f83bafdb

FagnerMartinsBrack:
### Re-review at `7a34dd1d`: follow curl redirects (SSRF-guarded) + back off failing stale-checks

The Medium flagged in the prior review — the missing scheme allowlist on the curl redirect loop — is now **correctly fixed**. `curl-fetch.ts:121-125` refuses any redirect whose resolved target is not `http:`/`https:` *before* the next `fetchOnce`/curl spawn, closing the `gopher://`/`file://`/`dict://`/… open-egress + LFI class regardless of the libcurl protocol set. Using a `Set` (mirroring `REDIRECT_STATUS_CODES`) keeps both branches covered, and the new negative-space test (`seq.calls` stays length 1 for a `gopher://` Location) mirrors the existing SSRF-refusal test. I traced the layered defense and it holds: `validateSaveableUrl` (`saveable-url.ts:24,92`) already constrains saved URLs to http(s) and blocks private hostnames at save time, the import path extracts only `http(s)://`, `resolvePinnedAddress` re-runs per hop (private-IP + DNS-rebinding pin), and the redirect `Location` — the one attacker-controlled scheme vector — is now allowlisted.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Stale `--location` references in the parser doc comments
`curl-fetch.ts:209` ("With `--location`, intermediate redirect headers appear before the final ones") and `curl-fetch.ts:239` ("With `--location`, each redirect response has its own header block") describe a flag this PR removed — with `--max-redirs 0`, curl now emits a single response's headers per spawn. Keep the multi-block parsing (it still fires for HTTP `1xx` interim responses — `100 Continue` / `103 Early Hints`), but update the *rationale* so the comments stop pointing at behaviour that no longer exists.

### 2. No aggregate deadline across redirect hops when no `signal` is passed
`fetchOnce` applies `DEFAULT_TIMEOUT_MS` (10s) per hop only when `init.signal` is absent (`curl-fetch.ts:79`). With app-level redirect following, a chain from slow origins can now consume up to ~6×10s (~60s) of wall-clock on the curl path (vs. 10s pre-PR) when the caller supplies no signal — bounded by `MAX_REDIRECTS`, so minor, but a deliberate origin could pad the crawl budget. When a `signal` *is* passed the caller's deadline spans all hops (correct), so this only affects the no-signal callers. Worth a conscious call on whether the 10s should be an overall budget.

### 3. (Optional, defense-in-depth) The curl entry URL scheme is not self-guarded
The new `ALLOWED_REDIRECT_PROTOCOLS` check guards redirect hops but not the initial `fetchOnce(url)` — the entry URL's scheme is trusted. Production is safe today (`validateSaveableUrl` enforces http(s) at save and the import extractor only yields http(s)), and I confirmed a non-http(s) entry *would* otherwise cascade undici→h2→curl (`shouldTryFallback` in `h2-fetch.ts:206` defaults to `true`). Since `crawl-article` is a reusable, security-sensitive package, hoisting the allowlist into `fetchOnce` (or a one-line assert at the top of `fetchCurl`) would make the curl path uniformly http(s)-only regardless of caller — matching how the IP guard already runs at *both* save time and crawl time. Purely optional.

FagnerMartinsBrack:
@claude Address the following on this branch:

1. **(bug) Register the new redirect-loop error as block-class.** `fetchCurl` now throws `too many redirects (>5)`, but `BLOCK_ERROR_SIGNATURES` no longer matches any curl redirect-loop message — this PR deleted `"maximum (5) redirects followed"` and the remaining `"max_redirects"` is undici's spelling. An origin that 302-loops datacenter IPs (the documented bot-block behavior the deleted signature existed for) now aborts the persona chain on the first persona when it surfaces via the curl leg, while the identical loop via undici (`UND_ERR_MAX_REDIRECTS`) still rotates to the next persona. Add `"too many redirects"` to `BLOCK_ERROR_SIGNATURES` plus a test case in `persona-fallback.test.ts`.

2. **(bug — also review Low #2) One deadline across the redirect chain.** Each hop gets a fresh 10s `DEFAULT_TIMEOUT_MS` when no signal is passed, and the reachable `h2ThenCurl` last-resort branch (h2-fetch.ts:190) deliberately calls `curlFetchImpl` without the caller's aborted signal — so a slow 5-redirect chain can now hold the Lambda up to ~60s past the crawl deadline, where main was bounded to a single 10s attempt. Enforce a single overall budget: when `init.signal` is absent, create one `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` in `fetchCurl` and thread it into every `fetchOnce` hop.

3. **(bug) Wrap Location parsing in the error convention.** `new URL(location, currentUrl)` throws a bare `TypeError: Invalid URL` on a malformed Location header (e.g. a space in the host), escaping the `` `fetchCurl failed for ${url}: …` `` prefix every other failure path in this function carries — persona-fallback classifies on `error.message`, and logs lose the crawl context. Catch and rethrow prefixed, naming the offending Location value.

4. **(hardening) Sensitive-header hygiene on cross-origin hops.** The loop re-sends `init.headers` verbatim to every hop; undici strips `cookie`/`authorization`/`proxy-authorization` on cross-origin redirects, this loop doesn't. No current caller passes credentials (referer + persona headers only), so this is one caller away from a cross-origin credential replay. Drop those headers when the next hop's origin differs from the current one, with a test.

5. **(review Low #1 + inaccurate comment) Fix the stale/incorrect comments.** `parseCurlOutput` and `findLastHeaderBlock` docs still cite `--location`, which this PR removed — the multi-block parsing now matters for HTTP 1xx interim responses (100 Continue / 103 Early Hints), so update the rationale to say that. And the args comment credits `--max-redirs 0` with stopping redirects: without `--location` that flag is inert (curl only consults it under `-L`) — the actual guard is the omission of `--location`. Keep the flag as fail-closed defense (if `--location` is ever re-added, curl exits 47 instead of following unguarded) and make the comment say exactly that.

6. **(review Low #3) Hoist the scheme allowlist to the entry URL** — assert http(s) at the top of `fetchCurl` so the curl path is uniformly http(s)-only regardless of caller, with a refused-entry test (no curl spawned).

**Follow-up, not this PR:** crawl-article now carries three hand-rolled redirect loops (curl-fetch, h2-fetch, aia-fetch) with real drift — only curl-fetch has the scheme allowlist, and h2-fetch resolves relative Locations against `parsed.origin`, silently dropping path-relative targets. Extract a shared redirect-following helper (or at least align the guards) in a dedicated PR.

**Accepted trade-off, no change requested:** the backoff commit fires for transient failures too (a single 503/timeout at stale-check time delays the next refresh attempt by a full TTL). That's deliberate — the freshness clock is scheduling metadata only, content is untouched, and the alternative re-arms the per-view re-crawl storm.

Verify with `pnpm check`; commit to this branch.

FagnerMartinsBrack:

## Context
- **PR**: #889
- **Commit**: d5cda534078408ceb453665049fb8619df20dae7

FagnerMartinsBrack:
### Re-review at `d5cda534`: harden curl redirect-follow + register loop error as block-class

All six follow-ups from the previous round landed and are covered by tests. I traced each security-relevant path in `curl-fetch.ts` and the surrounding fallback chain:

- **Entry-URL scheme guard** (`curl-fetch.ts:106-111`) — `fetchCurl` refuses a non-http(s) entry *before* any `resolvePinnedAddress`/spawn. Combined with the per-hop redirect check (`:134-138`), every `fetchOnce` call is now provably http(s)-only regardless of caller (hop 1 via the entry guard, hops 2+ via the redirect guard). The `file://` entry test asserts zero spawns.
- **Single timeout budget** (`curl-fetch.ts:112`) — one `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` is synthesized once and threaded into every hop, so a worst-case 6-spawn redirect chain shares a 10s budget instead of 6×10s. This also correctly bounds the `h2ThenCurl` last-resort branch (`h2-fetch.ts:189-190`) that deliberately drops the caller's already-aborted signal — it now gets one fresh 10s budget rather than one-per-hop. Spy tests assert `timeout` is called once with `10000`, and not at all when the caller supplies a signal.
- **Wrapped `Location` parse** (`curl-fetch.ts:128-133`) — a malformed `Location` rethrows under the `fetchCurl failed for …` convention instead of a bare `TypeError`, so persona-fallback classifies on a stable message. Test asserts the prefix **and** no second spawn.
- **Cross-origin credential hygiene** (`curl-fetch.ts:139-141`, `:220-228`) — `cookie`/`authorization`/`proxy-authorization` are dropped when `nextUrl.origin` differs from the current hop, case-insensitively (`key.toLowerCase()`), and — because `headers` is reassigned to the stripped set — stay dropped for the rest of the chain (matching undici, which never restores them even on a hop back to the original origin). Same-origin retains them. Both directions tested. No caller passes credentials today, so this is pure hardening.
- **Block-class symmetry** (`persona-fallback.ts:33`) — `too many redirects` makes the curl-leg app-level loop rotate personas exactly like undici's `max_redirects` (`UND_ERR_MAX_REDIRECTS`). The removed `maximum (5) redirects followed` was genuinely dead (never matched curl's real `(0)` message). Matching is `.toLowerCase().includes(...)`, and only the two in-repo fetchers emit this phrase, so no false positives.
- **Comment fixes** — the `--location`/`--max-redirs 0` args comment and the `parseCurlOutput`/`findLastHeaderBlock` docs now describe reality (inert-flag fail-closed backstop; 1xx interim header blocks).

I also confirmed two things outside the immediate diff:
- The `ExecCurl` signature change (dropping the `options` arg) is clean — the only other referent, `crawl-fetch-ssrf.test.ts:127`, derives its mock via `Parameters<ExecCurl>` and never invokes it, so it adapts automatically and still type-checks.
- The stale-check `backoff` branch (`stale-check-handler.ts:123-134`) uses the identical `publishUpdateFetchTimestamp({ contentFetchedAt: now, bodyHash })` shape as the already-reviewed 304 "unchanged" path (`:115-119`), so `bodyHash` carry-forward (including the `undefined` case) is consistent, and terminal-failed rows still short-circuit at `:97` rather than backing off.

The core SSRF invariant holds and is now stronger than the PR framing: every curl spawn connects only to a `resolvePinnedAddress`-validated IP, over an http(s)-only URL, with cross-origin credentials stripped, bounded to 5 redirects under one deadline.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The new `too many redirects` block signature is the only entry without an explanatory comment
Every other entry in `BLOCK_ERROR_SIGNATURES` (`persona-fallback.ts:28-33`) carries a `/* … */` naming its origin (undici mapping, curl exit code, h2 RST_STREAM); `too many redirects` (`:33`) stands alone. You already flagged this and correctly deferred to the human on the code-comments approval gate. A one-liner such as `/* fetchCurl/fetchH2 app-level redirect cap — mirrors undici's max_redirects */` would restore the array's self-documentation. Worth noting the signature is doubly earned: `fetchH2` also throws `too many redirects for …` (`h2-fetch.ts:55`), so this one entry now covers redirect loops on **both** the h2 and curl legs.

### 2. (Acknowledgment) The three hand-rolled redirect loops remain divergent
Your PR note already scopes this to a follow-up, and I independently confirmed it's worth prioritizing: only `curl-fetch` has the scheme allowlist, and `h2-fetch.ts:44` resolves a relative `Location` against `parsed.origin` (scheme+host only) whereas `curl-fetch.ts:130` correctly resolves against the full `currentUrl`. So a *path-relative* redirect (`Location: sub/page`) resolves to `https://host/sub/page` on the h2 leg instead of `https://host/current/path/sub/page` — a real latent correctness bug, not just style drift. The dedicated extraction PR is well justified.

FagnerMartinsBrack:

## Context
- **PR**: #889
- **Commit**: 1b5134b21310b72d574db74198c68c58735bdbd7

FagnerMartinsBrack:
### Review at `1b5134b2`: name the origin of the too-many-redirects block signature

The only code delta since the last (approving) review at `d5cda534` is a single-line doc comment on `persona-fallback.ts:33`, with explicit human approval per the code-comments gate. It addresses Low #1 from the prior round. I re-traced the security-critical paths in `curl-fetch.ts` and the surrounding fallback chain — entry + per-hop scheme allowlist, per-hop `resolvePinnedAddress`, single shared `AbortSignal.timeout` budget, wrapped `Location` parse, cross-origin credential strip, and the `MAX_REDIRECTS` cap all still hold and are covered by tests (including the negative-space `seq.calls`/`fake.calls`-length assertions that a refused hop spawns no curl). Commit 2's stale-check backoff is unchanged and correct.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The new block-signature comment names 2 of 3 emitters, and neither of the two it *would* add reaches the classifier in the live composition
`persona-fallback.ts:33` now reads `/* fetchCurl/fetchH2 app-level redirect cap — mirrors undici's max_redirects */`. Two precision gaps:

- **Omits a third emitter.** `git grep "too many redirects"` shows three app-level fetchers throw the phrase, not two: `curl-fetch.ts:122`, `h2-fetch.ts:55`, **and `aia-fetch.ts:88`** (`fetchAia: too many redirects for ${url}`). The comment's siblings each name what emits them (`curl exit 92`, `undici's mapping`, …), so by that convention `fetchAia` belongs in the list.
- **Only the curl leg actually reaches `isBlockClassError` today.** Tracing the `initCrawlFetch` wiring (`crawl-fetch.ts:54-61`): `withPersonaFallback( withH2Fallback( withAiaChasing(guardedFetch, fetchAia), fetchH2, fetchCurl ) )`. When `fetchAia` throws its redirect-loop error it surfaces in `withH2Fallback`'s catch (`h2-fetch.ts:151`); `shouldTryFallback` returns `true` (no network code, signal not aborted) so it is swallowed and replaced by an `h2ThenCurl` attempt. Inside `h2ThenCurl`, `fetchH2`'s redirect-loop error hits the same `shouldTryFallback` catch (`h2-fetch.ts:185`) and falls through to curl. Only `fetchCurl` — the terminal call, not wrapped in a fallthrough — propagates out to `isBlockClassError`. So in the current composition the h2/aia variants never independently trigger persona rotation.

The comment's *rationale* ("app-level redirect cap — mirrors undici's `max_redirects`") is correct and useful, and the human already approved the annotation — this is purely about the enumeration. Cheapest precise fix: add the missing emitter → `/* fetchCurl/fetchH2/fetchAia app-level redirect cap — mirrors undici's max_redirects */`. The signature matching is a lowercase `.includes(...)`, so all three forms are caught regardless, and the single `persona-fallback.test.ts:43` case remains sufficient coverage.

### 2. (Acknowledgment) The three hand-rolled redirect loops remain divergent
Your PR note already scopes this to a dedicated follow-up, and reviewing `aia-fetch.ts` for the point above independently reconfirmed it: `curl-fetch.ts:130` resolves a relative `Location` against the full `currentUrl`, `h2-fetch.ts:44` resolves against `parsed.origin` (scheme+host only, dropping a path-relative target), and `aia-fetch.ts:80` resolves against `currentUrl` (correct) but carries no scheme allowlist. Only `curl-fetch` has the scheme guard. A shared redirect-following helper would collapse all three — the extraction PR is well justified.

PR #899, opened 2026-07-02
Add in-flight loader to mark-read/unread and toast actions

FagnerMartinsBrack:
Adds a three-dot animated loader affordance to the mark-read/unread status toggle in queue cards and to toast action buttons (e.g., "Undo"). This mirrors the reader's mark-read control behavior, providing visual feedback during the htmx round-trip while the `<main>` fragment is being swapped.

## Changes

- **Queue card status button**: Wraps the button text in a `.queue-article__action-btn-label` span and adds a `.queue-article__action-btn-loader` with three animated dots. The loader displays while the form carries the `htmx-request` class, and the label is hidden via `visibility` to keep button width stable. The form gains `hx-disabled-elt="find button"` to disable the button during the request.

- **Toast action button**: Applies the same pattern with `.toast__action-label` and `.toast__action-loader`. The form always includes `hx-disabled-elt="find button"` since toast actions always trigger a full-page swap.

- **Delete action**: Remains a bare icon with no loader, label wrapper, or request-disable behavior—only the mark-read/unread toggle gets the affordance.

- **Styling**: Adds shared animation keyframes (`queue-mark-read-dot` and `toast-action-dot`) with staggered delays (0s, 0.4s, 0.8s) and respects `prefers-reduced-motion`.

- **Tests**: Verifies the status button has the loader structure, label span, and `hx-disabled-elt` attribute; confirms the delete button has none of these; and validates the toast action button follows the same pattern.

## Implementation Details

- The loader uses `position: absolute` with `inset: 0` to overlay the button, and the button gains `position: relative` to establish the positioning context.
- The label is kept in flow via `visibility: hidden` (not `display: none`) so the button dimensions remain stable while the loader animates.
- During the request, the disabled button shows `cursor: progress` and `opacity: 1` to distinguish it from the pre-fetch dimmed state.
- The `isStatus` flag in `ActionDisplayModel` gates the loader and `hx-disabled-elt` behavior, allowing the delete action to opt out.

https://claude.ai/code/session_016KkBLx75jUk4XeN3NGUJnm


FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: 4b07206ff804b186ee5ef0dd131ab0d3a5b9f893

FagnerMartinsBrack:

None found.

The core mechanism is sound. I specifically verified the concern that mattered most: the new selectors `.queue-article__action-form.htmx-request` and `.toast__action-form.htmx-request` require htmx to place the `htmx-request` class on the **form** element — a stricter requirement than the reader control, which uses an ambient `.htmx-request` descendant selector that would match regardless. This is correct: for a boosted form, htmx 2.x adds `htmx-request` to the requesting element (the form), and `hx-disabled-elt="find button"` resolves its descendant search from that same form — the two are internally consistent. Toasts render inside the swapped `<main>` (per `toast.client.ts`), so htmx processes/boosts them and the affordance activates.

Notably, the divergence from the reader (form-scoped vs. ambient) is **deliberate and prevents a real bug**: the queue card self-polls (`hx-get` on the root `every 3s`), which puts `htmx-request` on the card root. A naive copy of the reader's ambient `.htmx-request .…-label` selector would have flashed the loader on every poll. The scoping avoids that, and the CSS comment documents the reasoning accurately.

## Medium Priority Issues

### 1. The delete-control test proves its point with the `querySelector(...).toBeNull()` anti-pattern

`queue-card.component.test.ts` — "leaves the delete control as a bare icon…" distinguishes the delete button from the status button entirely through negative assertions:

```ts
expect(deleteButton.textContent).toBe("×");
expect(deleteButton.querySelector(".queue-article__action-btn-loader")).toBe(null);
expect(deleteButton.querySelector(".queue-article__action-btn-label")).toBe(null);
expect(deleteButton.closest("form")?.hasAttribute("hx-disabled-elt")).toBe(false);
```

The [test-driven-design skill](.claude/skills/test-driven-design/SKILL.md) calls this out explicitly ("Never Rely on `querySelector(...).toBeNull()`"): a typo in `.queue-article__action-btn-loader`/`-label` also returns `null`, so the assertion passes for the wrong reason. The one positive assertion here doesn't compensate — because the loader dot-spans are empty, `deleteButton.textContent` is `"×"` **even if the delete button wrongly received the `isStatus` treatment**, so it cannot catch the regression it appears to guard against. If a future change made `delete` satisfy `isStatus`, this test could silently keep passing.

A single positive structural assertion is typo-proof and actually distinguishes the two shapes — the delete button renders bare text (no element children), whereas the status button renders two `<span>` wrappers:

```ts
expect(deleteButton.textContent).toBe("×");
expect(deleteButton.children.length).toBe(0); // no label/loader wrapper spans
```

[Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fmark-read-loading-behavior-vkgauf%2C%20in%20projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fpages%2Fqueue%2Fqueue-card%2Fqueue-card.component.test.ts%2C%20replace%20the%20negative%20assertions%20in%20the%20%27leaves%20the%20delete%20control%20as%20a%20bare%20icon%27%20test%20%28querySelector%28...%29.toBe%28null%29%20and%20hasAttribute%28...%29.toBe%28false%29%29%20with%20a%20positive%20structural%20assertion%20like%20expect%28deleteButton.children.length%29.toBe%280%29%20so%20a%20selector%20typo%20cannot%20make%20the%20test%20pass%20for%20the%20wrong%20reason%2C%20per%20the%20test-driven-design%20%27Never%20Rely%20on%20querySelector%28...%29.toBeNull%28%29%27%20rule.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. Button now has two internal shapes via `{{#if isStatus}}…{{else}}…{{/if}}`

`queue-card.template.html` branches the button's inner markup on `isStatus`, producing two shapes (label+loader spans vs. bare `{{text}}`). The [web skill](.claude/skills/web/SKILL.md) ("Iterate Lists, Don't Branch in Templates") favors one template shape per iterated item. This is a defensible exception — the `isStatus` decision is already computed in TypeScript (`toActionDisplayModel`, the testable part), and the bare `×` delete icon genuinely shouldn't carry loader dot-spans — so I'd treat it as acceptable. Flagging only because a reviewer walking the web-skill checklist will notice it. An alternative is to always render the label wrapper and gate the loader purely via the flag/CSS, but that would put dead dot-spans inside the delete button, which is arguably worse.

### 2. The dot-loader idiom is now duplicated three times

`queue-mark-read-dot` (queue.styles.css), `toast-action-dot` (web-shell toast.styles.ts), and the pre-existing `article-body-mark-read-dot` are byte-for-byte-identical keyframes, and the surrounding loader block (`position:absolute; inset:0; display:none/flex; gap; nth-child delays; prefers-reduced-motion`) is near-identical across all three. Since this is now an established "mirror the reader's control" idiom, consider extracting a shared CSS partial/constant to prevent drift (e.g. one animation tuned in one place). Mitigating factor: they live in separate bundles (two hutch stylesheets + the shared web-shell package), so some duplication is expected and this is genuinely optional.

### 3. Toast button loses its accessible name mid-request

During the request, `.toast__action-label` is `visibility: hidden` (removed from the a11y tree) and `.toast__action-loader` is `aria-hidden="true"`, so the (disabled) toast button has no accessible name while in flight. Impact is low — the state is transient and the button is disabled — and it mirrors the reader control exactly. Note this only affects the toast: the queue status button carries `title="{{title}}"`, so it retains an accessible name throughout. If you want parity, a `title`/`aria-label` on `.toast__action` would cover the in-flight window.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #899
- **Branch**: claude/mark-read-loading-behavior-vkgauf

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: f65e2b4c08e099075cfe21aa52aab8bbb38bfb8b

FagnerMartinsBrack:

None found.

The core mechanism remains sound, and the one MEDIUM from the previous round is now resolved. The fix commit (`f65e2b4c`) is scoped to a single test file: it replaced the delete-control test's three negative assertions (`querySelector(...).toBe(null)` ×2 + `hasAttribute(...).toBe(false)`) with one positive structural assertion, `expect(deleteButton.children.length).toBe(0)`. Because `queue-card.template.html` gates *both* the button's inner shape (the `label` + `loader` spans) **and** the form's `hx-disabled-elt` on the same `{{#if isStatus}}` flag, `children.length === 0` is a typo-proof proxy for `isStatus === false` — the status button renders two `<span>` children, the delete button renders bare `×`. A selector typo can no longer make the test pass for the wrong reason, and a future regression that made `delete` satisfy `isStatus` would fail the assertion. This is precisely the resolution the review proposed, and it aligns the test with the test-driven-design *"Never Rely on `querySelector(...).toBeNull()`"* rule.

I also re-verified the load-bearing behaviors:
- **Form-scoped `htmx-request` selectors are still correct.** `.queue-article__action-form.htmx-request` / `.toast__action-form.htmx-request` require htmx to place `htmx-request` on the *form*, not an ancestor. The card self-polls (`hx-get … every 3s`), which puts `htmx-request` on the card root — an ambient descendant selector (as the reader uses) would flash the loader on every poll. The scoping avoids that, and the CSS comment documents it accurately.
- **The queue vs. toast `:disabled` asymmetry is intentional and correct.** The queue scopes the progress-cursor rule under `.htmx-request` (`.queue-article__action-form.htmx-request .queue-article__action-btn--status:disabled`) because the status button has *two* disabled states — the pre-fetch processing state (`disabled` + dimmed `opacity: 0.5`) and the in-flight state — that must look different. The toast button has only the in-flight disabled state, so its unscoped `.toast__action:disabled` is fine and mirrors the reader's `.article-body__mark-read:disabled` exactly.
- **The toast affordance activates.** Toasts render inside the swapped `<main>` (they arrive via an htmx swap and the initial-load process), so htmx boosts the form and adds `htmx-request` on submit; `hx-disabled-elt="find button"` resolves its descendant search from that same form.
- **Existing `textContent` assertions still hold.** `queue.listing.route.test.ts:231-232` (`"Mark as read"`, `"×"`) pass because the loader dot-spans are empty — the label span carries the whole visible text.

## Medium Priority Issues

None found.

The sole MEDIUM from the prior review (the delete-control test's `querySelector(...).toBeNull()` anti-pattern) has been fixed correctly, as described above.

## Low Priority Suggestions

These three are unchanged from the prior review and were intentionally deferred (the auto-apply flow addresses HIGH/MEDIUM only). Re-listed for completeness — all remain acceptable, none are blocking.

### 1. Button has two internal shapes via `{{#if isStatus}}…{{else}}…{{/if}}`

`queue-card.template.html:12` branches the button's inner markup on `isStatus` (label+loader spans vs. bare `{{text}}`). The [web skill](.claude/skills/web/SKILL.md) ("Iterate Lists, Don't Branch in Templates") prefers one shape per iterated item. This is a defensible exception: the `isStatus` decision is computed in TypeScript (`toActionDisplayModel`, the testable part), and the bare `×` delete icon genuinely shouldn't carry loader dot-spans. Always rendering the label wrapper and gating purely via CSS would put dead dot-spans inside the delete button, which is arguably worse. Acceptable as-is.

### 2. The dot-loader idiom is now duplicated three times

`queue-mark-read-dot` (`queue.styles.css`), `toast-action-dot` (`web-shell/toast.styles.ts`), and the pre-existing `article-body-mark-read-dot` are byte-identical keyframes, and the surrounding loader block (`position:absolute; inset:0; display:none/flex; gap; nth-child delays; prefers-reduced-motion`) is near-identical across all three. Since this is now an established "mirror the reader's control" idiom, a shared CSS partial/constant would prevent drift. Mitigating factor: they live in separate bundles (two hutch stylesheets + the shared web-shell package), so some duplication is expected — genuinely optional.

### 3. Toast button loses its accessible name mid-request

During the request, `.toast__action-label` is `visibility: hidden` (removed from the a11y tree) and `.toast__action-loader` is `aria-hidden="true"`, so the (disabled) toast button has no accessible name while in flight. Impact is low — transient and disabled — and it mirrors the reader control. Note this affects only the toast: the queue status button carries `title="{{title}}"`, so it retains a name throughout. A `title`/`aria-label` on `.toast__action` would close the in-flight window if you want parity.

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Branch**: claude/mark-read-loading-behavior-vkgauf

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: 218018fc3aeb3399c00cc05652d9d09d85cb08b0

FagnerMartinsBrack:

None found.

The core mechanism is sound and unchanged since the prior approving review; this pass re-verified it at `218018fc` (post-rebase, CI green) with a focus on regressions from the conflict resolution.

- **No data loss from `hx-disabled-elt="find button"`.** The mutation payload lives in a hidden `<input name="status">`, not in the button's `name`/`value`, and htmx captures form data at submit time before applying the `disabled` attribute. Disabling the button therefore cannot drop the `status=…` field from the POST body.
- **CSS specificity is correct for the two disabled states.** The scoped rule `.queue-article__action-form.htmx-request .queue-article__action-btn--status:disabled { opacity: 1; cursor: progress }` has specificity (0,4,0), which beats both the base `.queue-article__action-btn:disabled { opacity: .5 }` (0,2,0) and `.queue-article__action-btn--status:disabled:hover { opacity: .5 }` (0,3,0). So the in-flight button stays full-opacity with a progress cursor even under hover, visibly distinct from the dimmed 0.5 pre-fetch processing state — exactly as the CSS comment claims.
- **Form-scoped `.htmx-request` remains the correct divergence from the reader.** The queue card self-polls (`hx-get … hx-trigger="every 3s"` on the root), which puts `htmx-request` on the card root; an ambient descendant selector (as the reader uses) would flash the loader on every poll. The reader page has no polling, so its ambient selector is safe there. The scoping is deliberate and documented.
- **Toast affordance activates and its unscoped `:disabled` is correct.** The toast renders inside the swapped `<main>` (`statusToastHtml` in `queue.component.ts`), so htmx boosts its form and adds `htmx-request` on Undo submit. Its `.toast__action:disabled` is intentionally unscoped because the toast has only the in-flight disabled state (no pre-fetch processing state) — this mirrors the reader's `.article-body__mark-read:disabled` exactly.
- **Existing assertions still hold.** `queue.listing.route.test.ts:231-232` (`"Mark as read"`, `"×"`) and `toast.component.test.ts:39` (`"Undo"`) pass because the loader dot-spans are empty; the label span carries the whole visible text.

## Medium Priority Issues

None found.

- The sole prior MEDIUM (delete-control test relying on the `querySelector(...).toBeNull()` anti-pattern) stays fixed. The positive `expect(deleteButton.children.length).toBe(0)` is a typo-proof proxy for `isStatus === false`: `queue-card.template.html` gates *both* the label+loader spans *and* the form's `hx-disabled-elt` on the same `{{#if isStatus}}` flag, so zero element children ⟺ the delete control opted out of the whole treatment. Both `isStatus` branches are exercised (the status test asserts the label span, three loader dots, and `hx-disabled-elt="find button"`; the delete test asserts the bare shape).
- The rebase/conflict resolution in `src/packages/web-shell/src/shared/toast/toast.styles.ts` is clean: it kept both main's `padding: var(--button-padding-sm)` (design-token refactor) and this PR's `position: relative` (positioning context for the absolute loader overlay). No PR change was reverted and no unrelated file was touched.
- Tests query the new spans by BEM class with **positive** assertions (`assert(label)`, `.length).toBe(3)`), which fail loudly on a rename. This does not violate the web skill's rule — that rule is one-directional ("never use `data-test-*` in CSS selectors"), and CSS here does not reference any test attribute.

## Low Priority Suggestions

These three are unchanged from the prior two reviews and were intentionally deferred (the auto-apply flow addresses HIGH/MEDIUM only). Re-listed for completeness — all remain acceptable, none are blocking.

### 1. Button has two internal shapes via `{{#if isStatus}}…{{else}}…{{/if}}`

`queue-card.template.html:12` branches the button's inner markup on `isStatus` (label+loader spans vs. bare `{{text}}`). The [web skill](.claude/skills/web/SKILL.md) ("Iterate Lists, Don't Branch in Templates") prefers one shape per iterated item. Defensible exception: the `isStatus` decision is computed in TypeScript (`toActionDisplayModel`, the testable part), and the bare `×` delete icon genuinely shouldn't carry loader dot-spans. Always rendering the label wrapper and gating purely via CSS would put dead dot-spans inside the delete button, which is arguably worse. Acceptable as-is.

### 2. The dot-loader idiom is now duplicated three times

`queue-mark-read-dot` (`queue.styles.css`), `toast-action-dot` (`web-shell/toast.styles.ts`), and the pre-existing `article-body-mark-read-dot` are byte-identical keyframes, and the surrounding loader block (`position:absolute; inset:0; display:none/flex; gap; nth-child delays; prefers-reduced-motion`) is near-identical across all three. Since this is now an established "mirror the reader's control" idiom, a shared CSS partial/constant would prevent drift. Mitigating factor: they live in separate bundles (two hutch stylesheets + the shared web-shell package), so some duplication is expected — genuinely optional.

### 3. Toast button loses its accessible name mid-request

During the request, `.toast__action-label` is `visibility: hidden` (removed from the a11y tree) and `.toast__action-loader` is `aria-hidden="true"`, so the (disabled) toast button has no accessible name while in flight. Impact is low — transient and disabled — and it mirrors the reader control (which also has no `title`/`aria-label` and hides its label the same way). Note this affects only the toast: the queue status button carries `title="{{title}}"`, so it retains a name throughout. A `title`/`aria-label` on `.toast__action` would close the in-flight window if you want parity.

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: 7d65da0321196a2ac3d3e9904ba5169604fc56c2

FagnerMartinsBrack:

None found.

This pass focused on the new commit `7d65da03`, which addresses low-priority item #3 by adding `title="{{label}}"` to the toast action button (plus a positive test assertion). The change is correct and introduces no regressions:

- **The `title` genuinely closes the in-flight a11y gap.** Mid-request the toast label is `visibility: hidden` (removed from the accessibility tree) and the loader dots are `aria-hidden="true"`, so the disabled, in-flight button previously had **no accessible name**. `title="{{label}}"` supplies one for exactly that window — the correct fix for the flagged item.
- **No injection risk.** `render()` compiles the template with Handlebars (`render.ts`), so the double-stache `{{label}}` in `title="{{label}}"` is HTML-attribute-escaped. A label containing `"` / `<` / `&` cannot break out of the attribute.
- **Existing text assertions still hold.** `queue.listing.route.test.ts:231-232` assert `textContent` is `"Mark as read"` / `"×"`; the label span carries the whole visible text and the loader dot-spans are empty, so concatenated `textContent` is unchanged. `toast.component.test.ts:39` (`"Undo"`) holds for the same reason.
- **Toast `:disabled` styling is in-flight-only and correct.** The toast button is never server-disabled (no `{{#if disabled}}` in `toast.template.html`) and there is no base `:disabled` dimming rule, so the unscoped `.toast__action:disabled { opacity: 1; cursor: progress }` fires only while htmx has disabled the button via `hx-disabled-elt="find button"`. This mirrors the reader's `.article-body__mark-read:disabled` exactly; the queue's scoped-under-`.htmx-request` variant is correctly different because the queue status button has a *second* (pre-fetch dim) disabled state the toast lacks.
- **Prior MEDIUM stays fixed.** The delete-control test's positive `expect(deleteButton.children.length).toBe(0)` remains a typo-proof proxy for `isStatus === false` (the template gates both the label+loader spans and the form's `hx-disabled-elt` on that one flag).

## Medium Priority Issues

None found.

The core mechanism (form-scoped `.htmx-request` divergence from the reader's ambient selector to avoid a loader flash on the card's `every 3s` self-poll; `hx-disabled-elt` capturing form data before disabling; CSS specificity keeping the in-flight state full-opacity even under hover) was verified sound across prior rounds and is unchanged here. The new commit is scoped to the toast template + its test only.

## Low Priority Suggestions

### 1. The reader's mark-read control still has the same in-flight nameless window this PR just closed for the toast

The fix gives the **toast** an accessible name in flight, and the **queue** status button already carries `title="{{title}}"`. But the control this whole feature mirrors — the reader's `.article-body__mark-read` — renders with **no `title` and no `aria-label`** in both `reader-actions-top.template.html:15` and `reader-actions-bottom.template.html:15`, and hides its label under the same `visibility: hidden` (`article-body.styles.css:79`) with an `aria-hidden` loader. So the reader button still goes nameless for its in-flight window. This is **pre-existing and out of this PR's scope** (the PR doesn't touch those files), but a one-line follow-up adding `title="{{markRead.label}}"` there would give app-wide parity. Non-blocking. [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fmark-read-loading-behavior-vkgauf%2C%20add%20title%3D%22%7B%7BmarkRead.label%7D%7D%22%20to%20the%20.article-body__mark-read%20button%20in%20both%20projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fshared%2Farticle-body%2Freader-actions%2Freader-actions-top.template.html%20and%20reader-actions-bottom.template.html%20so%20the%20reader%20mark-read%20control%20keeps%20an%20accessible%20name%20during%20its%20in-flight%20loader%20window%2C%20matching%20the%20queue%20and%20toast%20buttons.%20Add%20a%20positive%20title%20assertion%20to%20the%20reader-actions%20component%20test.&repo=Readplace/readplace.com)

### 2. `title="{{label}}"` duplicates the visible label when the toast is *not* in flight

Because the toast's `title` equals its visible text, the attribute is redundant in the normal (visible-label) state: hovering shows an "Undo" tooltip over an "Undo" button, and screen readers may announce the name twice (content name + `title` description). This is the accepted trade-off of mirroring the queue button's `title` pattern — note the queue differs slightly in that its `title` is a distinct, fuller string than its short `text`, so the queue doesn't get the redundant tooltip. An `aria-label="{{label}}"` would supply the in-flight name **without** a visible hover tooltip or the doubled announcement, at the cost of not matching the queue's `title` idiom. Either is defensible; flagging only so the redundant-visible-tooltip trade-off is a conscious choice. Non-blocking.

### 3. Two previously-deferred items remain, intentionally

- **Template shape-branch** (`queue-card.template.html:12`, `{{#if isStatus}}…{{else}}…`) — the `isStatus` decision is computed and unit-tested in TS (`toActionDisplayModel`), and every branch-free rewrite is a net regression (dead `aria-hidden` dot-spans inside the delete button, or changing delete's product behavior). Acceptable as-is; the author documented the rationale.
- **Triplicated dot keyframes** (`queue-mark-read-dot`, `toast-action-dot`, and the pre-existing `article-body-mark-read-dot`) — colocated keyframes are this repo's established convention (`trial-countdown-pulse`, `queue-processing-pulse`, etc.), the three blocks aren't byte-identical (different BEM selectors), and they live in separate bundles. Genuinely optional.

FagnerMartinsBrack:
@claude change isStatus to status: 'loading', 'loaded' and address all remaining low prio issues

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: 850974edc22c7a2429b4a16e8a8fc48b3709fb88

FagnerMartinsBrack:

None found.

This pass focused on the new commit `850974ed`: the `isStatus` boolean → `status: 'loading' | 'loaded'` enum refactor and the two in-flight `aria-label` a11y fixes. All verified sound.

- **The `eq`-helper wiring is correct and would fail loudly if it weren't.** `render` applies caller helpers at invocation time (`compiled(data, { helpers: options?.helpers })`, `render.ts:29`), and `render.test.ts:49` covers exactly this path. In `{{#if (eq status "loading")}}` Handlebars invokes `eq(status, "loading", options)` — the trailing `options` arg is harmlessly ignored by `const eq = (a, b) => a === b`, so the result is `status === "loading"`. An unresolved subexpression helper is a hard render error (not a silent falsy), so a broken wiring couldn't pass silently; both the `"loading"` and `"loaded"` branches are exercised by the status-toggle and delete tests, keeping `functions` at 100%.
- **The rename is complete.** No dangling `isStatus` remains in the model, template, or tests (`isStatusAction` is just the local computing variable; the reader/toast keep their own class-based markup).
- **Both `aria-label` fixes are correct and drift-free.** `aria-label="{{markRead.label}}"` / `aria-label="{{label}}"` draw from the *same* binding as the visible `<span>` label, so the accessible name and visible text are guaranteed identical — WCAG 2.5.3 (Label in Name) holds and there's no double announcement in the resting state. In flight, the label span is `visibility:hidden` (dropped from the a11y tree) and the loader is `aria-hidden`, so `aria-label` is what keeps the disabled button named. Swapping the toast's `title` for `aria-label` removes the redundant hover tooltip that merely repeated the visible label; double-stache escaping keeps the attribute injection-safe.
- **Prior MEDIUM stays fixed.** The delete-control test's positive `expect(deleteButton.children.length).toBe(0)` remains a typo-proof proxy for the non-loader (`"loaded"`) shape — the template gates both the label+loader spans and the form's `hx-disabled-elt` on `(eq status "loading")`.

## Medium Priority Issues

None found.

The load-bearing mechanics verified across prior rounds are unchanged by this commit (the CSS wasn't touched): the form-scoped `.htmx-request` divergence from the reader's ambient selector that avoids a loader flash on the card's `every 3s` self-poll; `hx-disabled-elt` capturing form data before disabling; and the scoped `:disabled` specificity keeping the in-flight state full-opacity even under hover. This commit is scoped to the queue-card model/template + the reader/toast templates and their tests.

## Low Priority Suggestions

### 1. `status: 'loading' | 'loaded'` reads as a request state but encodes a static per-action classification

The field is named `status` with temporal-sounding values, but it is a *static* property of the action decided once at render time (`status: isStatusAction ? "loading" : "loaded"`, `queue-card.component.ts:46`) — it never reflects an actual in-flight request. The tell is the delete action being permanently `'loaded'`: delete has nothing to do with loading, and it triggers the same full-`<main>` swap as mark-read — it just opts out of the loader affordance. A future reader seeing `ActionDisplayModel.status: 'loading'` could reasonably expect the server computed a live loading state. What the flag actually encodes is "does this action render the in-flight loader affordance," which a boolean `hasLoader` or an enum like `'with-loader' | 'bare'` would map to more directly.

This is entirely your call and **non-blocking** — you explicitly requested `status: 'loading' | 'loaded'` and it's implemented faithfully and correctly. Flagging only so the semantic subtlety is on the record. Related note for accuracy: the enum swap did not remove the template shape-branch (Low item from prior rounds) — `{{#if (eq status "loading")}}…{{else}}…{{/if}}` is the same branch as before, now expressed via the enum plus a small `eq` subexpression helper. That remains an acceptable exception (the branch decision lives in unit-tested TS), just no longer accurately described as "eliminated."

### 2. The dot-loader keyframes remain triplicated (unchanged, deferred by design)

`queue-mark-read-dot`, `toast-action-dot`, and the pre-existing `article-body-mark-read-dot` still share a byte-identical 6-line easing curve across three bundles. As established over the prior rounds, colocated keyframes are this repo's universal convention (zero shared keyframes today), a shared keyframe would be coupled only by a bare `animation:` name string in `.css` files with no compile-time check, and the three blocks aren't fully identical (reader dots are `0.45em` vs `0.4em`). Genuinely optional; noted for completeness.

FagnerMartinsBrack:
@claude 'with-loader' | 'bare' would map to more directly. Do that change

FagnerMartinsBrack:
@claude 'with-loader' | 'bare' would map to more directly. Do that change.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28694296403
- **PR**: #899
- **Branch**: claude/mark-read-loading-behavior-vkgauf

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #899
- **Commit**: 77a379561f546ba4fb2b1b54afad8288dd20fc56

FagnerMartinsBrack:

None found.

This pass focused on the CI-fix **merge** (`77a37956`), which pulled `origin/main` into the branch — the first review since the merge. The failure CI caught was *semantic drift*, not a textual conflict (git auto-merged the four shared queue-card files cleanly, but `main` had made `deviceClass` a required param and split `linkUrl` into three tracked links). I verified the merged tree is coherent:

- **`queue-card.component.ts` kept both sides.** Main's `deviceClass` param and 3-link split (`titleLinkUrl`/`excerptLinkUrl`/`thumbnailLinkUrl`, lines 55, 63–65) coexist with this PR's `affordance: "with-loader" | "bare"` enum, the `eq` helper, and the `hx-disabled-elt`/loader wiring (lines 20, 49, 78–83). No PR logic was reverted and no `main` logic was dropped.
- **The two new tests route through `display()`.** The helper defaults `deviceClass: "desktop"` (lines 30–38), so neither new affordance test calls `toQueueCardDisplayModel` directly — the exact `TS2345` CI reported is eliminated, and both tests now match every other call site in the file. This is the minimal, idiomatic fix; no assertion was weakened.
- **Both production callers pass `deviceClass`** (`queue.component.ts:124-126`, `queue.page.ts:1211-1213`) — `main`'s code, untouched by the PR, so the merge took `main`'s version.
- **CI is green on `77a37956`** (run `28694841031`), confirming the auto-merged tree type-checks and all suites pass.

The load-bearing mechanics verified across prior rounds are unchanged by the merge and still hold: the `render()` per-call helper wiring (`compiled(data, { helpers: options?.helpers })`, `render.ts:29`) resolves `(eq affordance "with-loader")` correctly with both branches exercised (`functions` 100%); the form-scoped `.…-form.htmx-request` selectors avoid the loader flash on the card's `every 3s` self-poll; `hx-disabled-elt` captures form data before disabling (the `status=…` payload is a hidden input, not the button's `name`/`value`); and the toast's unscoped `.toast__action:disabled` fires in-flight only (re-confirmed against `origin/main` — no base `:disabled` dimming rule exists there).

## Medium Priority Issues

None found.

The sole prior MEDIUM (the delete-control test's `querySelector(...).toBeNull()` anti-pattern) remains fixed: the positive `expect(deleteButton.children.length).toBe(0)` (`queue-card.component.test.ts:319`) is a typo-proof proxy for `affordance === "bare"`, since the template gates *both* the label+loader spans *and* the form's `hx-disabled-elt` on `(eq affordance "with-loader")`. No dangling `isStatus`/`loading`/`loaded` references survived the rename (the only `isStatus` hit is the local `isStatusAction` computing variable).

## Low Priority Suggestions

### 1. The string enum requires an `eq` equality helper in the template (unchanged, deferred by design)

`queue-card.template.html` branches with `{{#if (eq affordance "with-loader")}}` at both sites, backed by the generic `const eq = (a, b) => a === b` passed per-call (`queue-card.component.ts:80-83`). This is the direct tradeoff of the `affordance: 'with-loader' | 'bare'` string enum you explicitly requested: a string enum can't be tested by Handlebars' built-in `{{#if}}` without an equality helper, so the enum reintroduces a small amount of template-level comparison that a boolean `hasLoader` would avoid. It remains acceptable — the *decision* still lives in unit-tested TS (`affordance: isStatusAction ? "with-loader" : "bare"`), the helper is a documented one-liner scoped to this render (not a global), and `functions` coverage stays at 100%. Flagging only so the enum-vs-boolean tradeoff is on the record; the enum reads better in the model, which is why it was chosen. Non-blocking.

### 2. The dot-loader keyframes remain triplicated (unchanged, deferred by design)

`queue-mark-read-dot` (`queue.styles.css`), `toast-action-dot` (`web-shell/toast.styles.ts`), and the pre-existing `article-body-mark-read-dot` still share a byte-identical 6-line easing curve across three bundles. As established over prior rounds: colocated keyframes are this repo's universal convention (zero shared keyframes today), a shared keyframe would be coupled only by a bare `animation:` name string in `.css` files with no compile-time check, and the three blocks aren't fully identical (reader dots are `0.45em` vs `0.4em` in queue/toast). Genuinely optional; noted for completeness.

PR #906, opened 2026-07-03
fix(save-link): fit select-content prompt in DeepSeek context via condense + budget cap

FagnerMartinsBrack:
## Why

The article "select content" selector asks `deepseek-chat` to pick the more complete of two candidate bodies (tier-0 vs tier-1). Commit `e58be7ac` patched a production incident (`en.wikipedia.org/wiki/Reading` produced a 1,117,671-token prompt against a 1,048,565-token context → deterministic HTTP 400 → DLQ flipped a successful crawl to `crawlStatus='failed'`) with a blunt `MAX_CANDIDATE_HTML_CHARS = 200_000` per-candidate front-slice plus a `400 → { winner: "tie" }` net.

Three problems remained:

1. **200k was untied to DeepSeek's real limit** and was a *per-candidate* bound, not a total — it only fit today because there happen to be exactly two tiers.
2. **A front-slice discards the rest of the article**, degrading the "most complete, least chrome" decision on large pages (differentiating chrome text can live past the first 200k chars) and can cut mid-tag.
3. **The embedded HTML was dense with markup the model never judges** — `href`/`class`/`style`/`data-*` attributes, comments, whitespace runs. The system prompt keys only on text and tag structure, so attributes just waste the token budget.

## What

**Condense, then budget-cap** each candidate:

- **New `condense-candidate-html.ts`** — walks the parsed DOM with `linkedom` (never regex, so code samples whose text contains literal `<`, `class=`, `href="…"` survive intact) and drops attributes, comments, `<script>`/`<style>` subtrees, and whitespace runs, keeping tag structure and all text. It is total and never throws (it runs inside select-content's `try`, where a non-400 throw would re-create the DLQ failure class). On a markup-heavy Readability body this roughly halves the characters (~49% in a spot-check) while title, prose, and chrome anti-signals ("5 min read", "Join Medium") all survive.
- **New `deepseek-limits.ts`** — extracts `DEEPSEEK_CONTEXT_TOKENS` (1,048,565 — the exact value the provider reports in its context-length 400) and `DEEPSEEK_MAX_OUTPUT_TOKENS` (8,192, moved out of `select-content.ts`) so both the model caller and the budget builder can import them without a cycle.
- **`select-content-prompt.ts`** — removes `MAX_CANDIDATE_HTML_CHARS`/`capCandidateHtml`; adds an exported, unit-testable `perCandidateHtmlCap(candidateCount)` derived from a total char budget (`INPUT_TOKEN_BUDGET × CHARS_PER_INPUT_TOKEN × SAFETY − reserve`). HTML tokenises at ~2–3 chars/token, so `CHARS_PER_INPUT_TOKEN = 2` keeps the char cap conservatively under the token limit (the char→token math `A ≤ R` is documented, including the CJK caveat, which degrades to the `400 → tie` net rather than a DLQ). Each candidate is condensed, then truncated to its per-candidate share; the total is now bounded (~1.76M chars ≈ ≤ ~880k tokens for a pair).

The existing `400 → tie` net stays as the ultimate backstop.

## Observability

When a condensed candidate still overflows its share and must be truncated (real article signal is lost), it now **logs an ERROR**. The `select-most-complete-content` worker was not in the shared `SAVE_LINK_LOG_GROUPS` constant, so its errors never reached the analytics dashboard's "Recent errors" widget — this wires it in (via `SAVE_LINK_LAMBDA_NAMES`, and names the Lambda from that constant in infra so the log-group and dashboard wiring can't drift from the deployed name). Those truncation ERRORs now surface on the widget.

## Verification

- `pnpm check` passes across the monorepo (100% coverage on all new/changed save-link files).
- New end-to-end fit test builds the prompt from two Wikipedia-sized (1.2M-char) candidates and asserts estimated input tokens (`message.length / CHARS_PER_INPUT_TOKEN`) stay under `INPUT_TOKEN_BUDGET` — the machine check that the original 400 can no longer occur.
- New `condense-candidate-html.test.ts` covers attribute/comment/script-style stripping, whitespace collapse, code-sample text preservation, anti-signal retention, and the empty-string early return.
- Dashboard tests derive from `SAVE_LINK_LOG_GROUPS` dynamically, so the newly-added worker log group is automatically asserted as surfaced by the errors widget.

https://claude.ai/code/session_01FoabC24fp2xadLbzYbMm2V

---
_Generated by [Claude Code](https://claude.ai/code/session_01FoabC24fp2xadLbzYbMm2V)_

### DIFF
--- a/projects/save-link/src/infra/index.ts
+++ b/projects/save-link/src/infra/index.ts
@@ -934,7 +934,7 @@ const selectMostCompleteContentDynamodb = new HutchDynamoDBAccess("select-most-c
 	actions: ["dynamodb:GetItem", "dynamodb:UpdateItem"],
 });
 
-const selectMostCompleteContentLambda = new HutchLambda("select-most-complete-content", {
+const selectMostCompleteContentLambda = new HutchLambda(SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent, {
 			entryPoint: "./src/runtime/select-most-complete-content.main.ts",
 		outputDir: ".lib/select-most-complete-content",
 		assetDir: "./src",
--- a/projects/save-link/src/infra/index.ts
+++ b/projects/save-link/src/infra/index.ts
@@ -925,11 +925,11 @@ eventBus.subscribe(StaleCheckRequestedEvent, staleCheckRequestedLambdaWithSQS);
 // AnonymousLinkSavedEvent (only on canonical change) and
 // CrawlArticleCompletedEvent (every successful selection).
 
-const selectMostCompleteContentQueue = new HutchSQS("select-most-complete-content", {
+const selectMostCompleteContentQueue = new HutchSQS(SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent, {
 	visibilityTimeoutSeconds: SELECT_CONTENT_TIMEOUTS.sqsVisibilitySeconds,
 });
 
-const selectMostCompleteContentDynamodb = new HutchDynamoDBAccess("select-most-complete-content-dynamodb", {
+const selectMostCompleteContentDynamodb = new HutchDynamoDBAccess(`${SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent}-dynamodb`, {
 	tables: [{ arn: articlesTableArn, includeIndexes: false }],
 	actions: ["dynamodb:GetItem", "dynamodb:UpdateItem"],
 });
@@ -955,13 +955,13 @@ const selectMostCompleteContentLambda = new HutchLambda(SAVE_LINK_LAMBDA_NAMES.s
 		...selectMostCompleteContentDynamodb.policies,
 		...contentBucket.readPolicies("select-most-complete-content-content-read"),
 		...contentBucket.writePolicies("select-most-complete-content-content-write"),
-		...renamePolicies(generateSummaryQueue.policies, "select-most-complete-content"),
+		...renamePolicies(generateSummaryQueue.policies, SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent),
 	],
 });
 
 eventBus.grantPublish(selectMostCompleteContentLambda);
 
-const selectMostCompleteContentLambdaWithSQS = new HutchSQSBackedLambda("select-most-complete-content", {
+const selectMostCompleteContentLambdaWithSQS = new HutchSQSBackedLambda(SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent, {
 	lambda: selectMostCompleteContentLambda,
 	queue: selectMostCompleteContentQueue,
 	alertEmailDLQEntry: alertEmail,
@@ -971,7 +971,7 @@ const selectMostCompleteContentLambdaWithSQS = new HutchSQSBackedLambda("select-
 eventBus.subscribe(TierContentExtractedEvent, selectMostCompleteContentLambdaWithSQS);
 
 // --- SelectMostCompleteContent DLQ consumer ---
-	new HutchDLQEventHandler("select-most-complete-content-dlq", {
+	new HutchDLQEventHandler(`${SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent}-dlq`, {
 	sourceQueue: selectMostCompleteContentQueue,
 	tableArn: articlesTableArn,
 	tableName: articlesTableName,
--- a/projects/save-link/src/runtime/domain/dom-node-types.ts
+++ b/projects/save-link/src/runtime/domain/dom-node-types.ts
@@ -1,0 +1,9 @@
+/**
+ * The two `Node.nodeType` codes this project's linkedom DOM walks branch on,
+ * named so callers read intent instead of a bare integer. Shared so the
+ * select-content condenser and the summary text-stripper reference one source
+ * rather than each spelling the codes out (one named, one magic) and drifting.
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType
+ */
+export const TEXT_NODE = 3;
+export const COMMENT_NODE = 8;
--- a/projects/save-link/src/runtime/domain/generate-summary/strip-html.ts
+++ b/projects/save-link/src/runtime/domain/generate-summary/strip-html.ts
@@ -1,8 +1,9 @@
 import assert from "node:assert";
 import { parseHTML } from "linkedom";
+import { TEXT_NODE } from "../dom-node-types";
 
 function extractText(node: Node): string {
-	if (node.nodeType === 3) {
+	if (node.nodeType === TEXT_NODE) {
 		assert(node.textContent !== null, "Text node must have textContent");
 		return node.textContent;
 	}
--- a/projects/save-link/src/runtime/domain/select-content/condense-candidate-html.test.ts
+++ b/projects/save-link/src/runtime/domain/select-content/condense-candidate-html.test.ts
@@ -1,9 +1,14 @@
 import { condenseCandidateHtml } from "./condense-candidate-html";
 
+// Comfortably above every fixture here so condensing runs; the raw-return guard
+// (input longer than the cap) is exercised on its own with a tiny cap below.
+const CAP = 10_000_000;
+
 describe("condenseCandidateHtml", () => {
 	it("drops href, class, style, id, and data-* attributes but keeps tags and prose", () => {
 		const out = condenseCandidateHtml(
 			'<article class="post" id="a" data-id="x"><h1 style="color:red">Title</h1><a href="/foo">link</a></article>',
+			CAP,
 		);
 
 		expect(out).not.toContain("href");
@@ -17,9 +22,7 @@ describe("condenseCandidateHtml", () => {
 	});
 
 	it("removes comment nodes at any depth", () => {
-		const out = condenseCandidateHtml(
-			"<!-- top --><div><p>text<!-- inner --></p></div>",
-		);
+		const out = condenseCandidateHtml("<!-- top --><div><p>text<!-- inner --></p></div>", CAP);
 
 		expect(out).not.toContain("<!--");
 		expect(out).not.toContain("top");
@@ -30,6 +33,7 @@ describe("condenseCandidateHtml", () => {
 	it("removes script and style subtrees including their text", () => {
 		const out = condenseCandidateHtml(
 			"<div><script>var x = 1 < 2;</script><style>.a{color:red}</style><p>keep</p></div>",
+			CAP,
 		);
 
 		expect(out).not.toContain("var x");
@@ -38,7 +42,7 @@ describe("condenseCandidateHtml", () => {
 	});
 
 	it("collapses whitespace runs to single spaces and trims", () => {
-		const out = condenseCandidateHtml("  <p>a\n\n\t  b</p>   ");
+		const out = condenseCandidateHtml("  <p>a\n\n\t  b</p>   ", CAP);
 
 		expect(out).toBe("<p>a b</p>");
 	});
@@ -46,6 +50,7 @@ describe("condenseCandidateHtml", () => {
 	it("keeps chrome anti-signal text the model keys on", () => {
 		const out = condenseCandidateHtml(
 			'<div class="byline">5 min read</div><div>Join Medium for free</div>',
+			CAP,
 		);
 
 		expect(out).toContain("5 min read");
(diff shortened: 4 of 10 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #906
- **Commit**: 37d0e6f4e38c3c5bc5269ad1d8ee4d96b9771e4a

FagnerMartinsBrack:

None found

## Medium Priority Issues

### 1. `condenseCandidateHtml`'s documented "must never throw" contract is assumed, not enforced — in the exact `try` whose non-400 throws re-open the DLQ path

`condenseCandidateHtml` runs inside `select-content.ts`'s `try` (via `buildSelectContentUserMessage` at `select-content.ts:54`). That `catch` only swallows HTTP 400s (`NonRetryableRejectionSchema`); **any other throw re-throws → SQS retry → DLQ → `crawlStatus='failed'` on a crawl that actually succeeded** — the precise incident `e58be7ac` and this PR exist to prevent. The function's own JSDoc states it *must never throw* for exactly this reason, but the guarantee is only *assumed*, resting on two unstated preconditions:

- `parseHTML(...)` is a third-party call (`linkedom`), and
- `assert(root, "Root element must exist")` throws if `document.getElementById("root")` is ever `null`.

Both hold in practice — the literal `<div id="root">` wrapper guarantees `#root`, and `linkedom` is already trusted on arbitrary crawled HTML in `strip-html.ts` and `sanitize-fragment.ts` — so this is **not a live bug**. But because the failure mode is catastrophic and load-bearing (a silent failed-crawl), the contract is worth *enforcing* rather than assuming: wrap the parse/condense body in `try/catch` and return the raw `html` on failure (it is still bounded by `perCandidateHtmlCap` in the caller), so a future `linkedom` change or a pathological Readability body can't re-open the DLQ path. Alternatively, soften the JSDoc to document the two assumptions instead of promising an absolute.

[Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fdeepseek-context-budget-hwa2r3%2C%20in%20projects%2Fsave-link%2Fsrc%2Fruntime%2Fdomain%2Fselect-content%2Fcondense-candidate-html.ts%2C%20enforce%20the%20documented%20%22must%20never%20throw%22%20contract%20of%20condenseCandidateHtml%3A%20wrap%20the%20parseHTML%2Fcondense%20body%20in%20try%2Fcatch%20and%20return%20the%20original%20html%20on%20any%20error%20(still%20bounded%20by%20perCandidateHtmlCap%20in%20the%20caller)%2C%20so%20a%20non-400%20throw%20here%20cannot%20DLQ%20a%20successful%20crawl.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. Full candidate DOM is parsed before the cap is applied
`condenseCandidateHtml` builds a full `linkedom` DOM over the *entire* candidate before `perCandidateHtmlCap` truncates it — replacing the previous O(1) `slice(0, 200_000)` with a full-DOM parse on unbounded input, for every candidate (the Wikipedia incident showed candidates reaching ~1.1M tokens ≈ multiple MB). The finalize Lambda is 3008 MB and already loads full tier HTML from S3, so this is likely fine, but it's worth confirming worst-case memory/time under `SELECT_CONTENT_TIMEOUTS.lambdaSeconds`.

### 2. Boilerplate + magic-number drift with `strip-html.ts`
The wrap-in-`<div>` / `parseHTML` / `assert(root)` / `.replace(/\s+/g, " ").trim()` sequence duplicates `generate-summary/strip-html.ts`. The two intentionally produce different outputs (structure-preserving vs text-only), so a full merge isn't warranted — but the shared wrapper boilerplate could be factored, and `strip-html.ts` uses the magic literal `nodeType === 3` while the new file introduces a named `COMMENT_NODE = 8`; a shared helper with named `TEXT_NODE`/`COMMENT_NODE` would unify them.

### 3. Partial constant adoption in `infra/index.ts`
The rename to `SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent` fixes the log-group/dashboard drift (only the `HutchLambda` name feeds the log group). But `"select-most-complete-content"` still appears as a bare string literal ~5× in the same file (the SQS queue, DynamoDB access, `HutchSQSBackedLambda`, the DLQ handler, and `renamePolicies` args). Not required for the dashboard fix, but pointing the remaining literals at the constant would remove the residual drift surface.

### 4. `"never grows the input"` test asserts a property that isn't universal
Verified locally that `linkedom` auto-closes tags (`<p>hi` → `<p>hi</p>`, `<li>a<li>b` → `<li>a</li><li>b</li>`), so `condenseCandidateHtml` *can* grow malformed/unclosed-tag input. Correctness never depends on this — the per-candidate cap bounds the output regardless — but the test name reads as a general invariant it isn't. Consider renaming to reflect that it shrinks well-formed, markup-heavy input.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #906
- **Branch**: claude/deepseek-context-budget-hwa2r3

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #906
- **Commit**: ca3eb648f190d2a9d88f2678b5a6822bab0fa2ac

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

The prior Medium (`condenseCandidateHtml`'s "must never throw" contract being assumed, not enforced) is resolved at `ca3eb648`: the JSDoc now honestly documents the two throw-site assumptions instead of promising an absolute — a resolution the prior reviewer explicitly offered, and consistent with how every sibling `parseHTML` call site in the repo (`strip-html.ts`, `sanitize-fragment.ts`, `resolve-relative-urls.ts`, `download-media.ts`) trusts linkedom inline. Verified end-to-end: the char→token budget is conservative (`perCandidateHtmlCap(2) ≈ 880k` chars ≈ ~880k input tokens at 2 chars/token, under DeepSeek's `1,040,373` input budget with SAFETY 0.85 + 8k reserve); the dashboard wiring surfaces the new worker log group automatically (`WORKER_DASHBOARD_LOG_GROUPS = Object.values(SAVE_LINK_LOG_GROUPS)`, dynamically asserted in `analytics-dashboard.test.ts:123`); log-group naming matches (`HutchLambda` appends `-handler` at `hutch-lambda.ts:210`); and the truncation ERROR reaches CloudWatch (composition root passes `consoleLogger`).

## Low Priority Suggestions

### 1. Full candidate DOM is parsed before the cap is applied
`condenseCandidateHtml` (`select-content-prompt.ts:65`) builds a full linkedom DOM over the entire candidate before `perCandidateHtmlCap` truncates it, replacing the previous O(1) `slice(0, 200_000)`. The select Lambda is 3008 MB (account ceiling) and already holds both full tier HTMLs in memory; realistic candidates run ~1–2 MB (the incident's Wikipedia page) up to ~40 MB → ~200 MB DOM at the documented 3–5× linkedom expansion — so this is very likely fine. Worth noting because the failure mode is asymmetric: an OOM is a process crash, not a throw, so it bypasses the `400 → tie` net entirely → SQS retry → DLQ → `crawlStatus='failed'`, re-creating the exact incident class this PR fixes, on the same large-page inputs. A worst-case memory/time spot-check under `SELECT_CONTENT_TIMEOUTS.lambdaSeconds` (300s) would retire the question.

### 2. Boilerplate + magic-number drift with `strip-html.ts`
The wrap-in-`<div>` / `parseHTML` / `assert(root)` / `.replace(/\s+/g, " ").trim()` sequence duplicates `generate-summary/strip-html.ts`. The two intentionally produce different outputs (structure-preserving vs text-only), so a full merge isn't warranted — but the shared wrapper boilerplate could be factored, and `strip-html.ts` uses the magic literal `nodeType === 3` while the new file introduces a named `COMMENT_NODE = 8`; a shared helper with named `TEXT_NODE`/`COMMENT_NODE` would unify them.

### 3. Partial constant adoption in `infra/index.ts`
The rename to `SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent` (`index.ts:937`) fixes the log-group/dashboard drift (only the `HutchLambda` name feeds the log group). But `"select-most-complete-content"` still appears as a bare string literal ~5× in the same file (HutchSQS queue `:928`, HutchDynamoDBAccess `:932`, `renamePolicies` arg `:958`, HutchSQSBackedLambda `:964`, HutchDLQEventHandler `:974`). These are Pulumi resource names, independent of the log group, so this is cosmetic — and since the constant value is identical, pointing them at it replaces no resource.

### 4. `"never grows the input"` test asserts a property that isn't universal
linkedom auto-closes tags (`<p>hi` → `<p>hi</p>`, `<li>a<li>b` → `<li>a</li><li>b</li>`), so `condenseCandidateHtml` can grow malformed/unclosed-tag input. Correctness never depends on this — the per-candidate cap bounds the output regardless — but the test name (`condense-candidate-html.test.ts:69`) reads as a general invariant it isn't. Consider renaming to reflect that it shrinks well-formed, markup-heavy input.

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #906
- **Commit**: f0adefa07e015d098c482594d4f88a32233be774

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

Re-reviewed at `f0adefa0`, the commit addressing all four prior Low-priority items. The verdict is unchanged from the previous two passes — no High, no Medium. I re-verified the load-bearing claims end-to-end and confirmed the four Low fixes are genuinely resolved rather than papered over:

- **DOM-parse OOM (prior Low 1)** — `condenseCandidateHtml` now takes a `maxInputChars` bound; above it the raw html is returned unparsed and the caller front-slices it, so a giant base64-heavy page degrades selection instead of OOM-crashing. This matters because an OOM is a process crash, not a 400, so it would bypass the `400 → tie` net → SQS retry → DLQ → `crawlStatus='failed'` on a crawl that in fact succeeded — the exact incident class this PR exists to prevent. The author ran the requested spot-check (~40 MB → ~1.6 GB RSS, ~80 MB → ~3.2 GB, OOM past ~74 MB vs the 3008 MB ceiling) and set `MAX_CONDENSE_INPUT_CHARS = 40_000_000` at ~½ the observed worst case. Candidates are condensed sequentially (at most one linkedom DOM live at a time), and the **fit guarantee is untouched** — the per-candidate cap always front-slices after condense whether it parsed or returned raw, so the machine-checked fit test still holds for >40M inputs. New test covers the raw-return branch.
- **Magic-number drift (prior Low 2)** — `dom-node-types.ts` exports `TEXT_NODE`/`COMMENT_NODE`; both are consumed (`strip-html.ts` drops its magic `3`, `condense-candidate-html.ts` drops its local `COMMENT_NODE = 8`), so the two spellings can't drift. It's a within-project shared module, not a cross-project relative import.
- **Infra constant adoption (prior Low 3)** — the five enumerated resource names now point at `SAVE_LINK_LAMBDA_NAMES.selectMostCompleteContent` with a byte-identical value, so no Pulumi resource is replaced; the `-dynamodb`/`-dlq` template literals reproduce the original names exactly. Import present at `index.ts:36`.
- **Test name (prior Low 4)** — renamed to `"shrinks well-formed markup-heavy input"`, honest given linkedom auto-closes tags.

## Low Priority Suggestions

### 1. JSDoc points at a constant defined in the caller, not in-file
The closing paragraph of `condense-candidate-html.ts` says *"See `MAX_CONDENSE_INPUT_CHARS` for the measured value,"* but that constant lives in the **caller** (`select-content-prompt.ts:53`) and reaches the function as the injected `maxInputChars` parameter — so the reference doesn't resolve from within the file the doc is written in. The parameter-injection design itself is correct (it's exactly what lets `condense-candidate-html.test.ts` inject a tiny `5` cap); only the doc pointer dangles. Consider referencing the `maxInputChars` parameter, or noting the concrete value is set by the caller. (Minor related note: the bound is expressed in `.length` chars while the spot-check that motivates it is in MB — for the ASCII/base64 outliers that actually trip it, chars ≈ bytes and the 40M-vs-~74M margin absorbs the difference, so this is a naming nuance, not a risk.) [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fdeepseek-context-budget-hwa2r3%2C%20in%20projects%2Fsave-link%2Fsrc%2Fruntime%2Fdomain%2Fselect-content%2Fcondense-candidate-html.ts%20the%20closing%20JSDoc%20says%20%22See%20MAX_CONDENSE_INPUT_CHARS%20for%20the%20measured%20value%22%20but%20that%20constant%20is%20defined%20in%20the%20caller%20select-content-prompt.ts%2C%20not%20in%20this%20file%3B%20reword%20to%20reference%20the%20maxInputChars%20parameter%20or%20note%20the%20constant%20lives%20in%20the%20caller.&repo=Readplace/readplace.com)

_The prior four Low items (DOM-parse-before-cap, magic-number drift, infra constant adoption, test name) are all resolved at `f0adefa0` — see the verification notes above._

PR #909, opened 2026-07-03
fix(ios-auth): reliably reuse Chrome web session on login

FagnerMartinsBrack:
## Problem

When the Readplace iOS app opens a browser to log in, a user already signed in to `readplace.com` **in Chrome** is forced to log in again. Two things defeated the existing session-reuse mechanism:

1. **Client fallback fired into the wrong browser.** `chromeURLForHTTPS` rewrote the authorize URL to `googlechromes://` only when a `canOpenURL` pre-check passed; a false-negative probe silently routed login into the **OS default browser (Safari)**, where the user isn't signed in. Login is required to be **Chrome-first** — most users browse in Chrome but never change the iOS default-browser setting.
2. **The session cookie didn't survive a browser restart.** `hutch_sid` was a bare session cookie (no `maxAge`/`expires`), so Chrome dropped it the moment it fully closed — even though the server-side session record lives for `SESSION_TTL_SECONDS`.

### Client — land login in Chrome, never Safari (`projects/ios-readplace`)
- `chromeURLForHTTPS` now rewrites `https` → `googlechromes://` **unconditionally**; whether Chrome can actually be opened is decided by the system at open time, not by a weak `canOpenURL` pre-check.
- New App-seam `openAuthorizeURLChromeFirst` opens the Chrome URL via `UIApplication.open(_:options:completionHandler:)` and falls back to the original https URL in the default browser **only** when the completion handler reports the open failed (Chrome not installed) — never as the default path.
- The UI-free `WebAuthFlow` core now exposes a single `openAuthorizeURL` seam handed the **raw https** authorize URL; the Chrome rewrite + open-failure fallback live in the App layer, matching the existing live-`UIApplication` pattern.
- `ExternalBrowser.open` gains a completion flag (mirrors `UIApplication.open`); the reader's plain "open externally" call ignores it.

### Server — persist the web session (`hutch` + `@packages/web-session`)
- Add `SESSION_COOKIE_MAX_AGE_MS` (`= SESSION_TTL_SECONDS * 1000`) and give the `hutch_sid` cookie a `maxAge` at every set-cookie site in `auth.page.ts` and `google-auth.page.ts`, so a Chrome login survives Chrome closing. Bounded by — never outliving — the same server TTL that evicts the session row. The Google `hutch_gstate` state cookie keeps its own short TTL (built from `baseCookieOptions` directly); `baseCookieOptions` (shared by visitor/click cookies) is untouched.

### Docs
- New "iOS login is Chrome-first" Product Constraint in `CLAUDE.md` and updated `projects/ios-readplace/README.md` so the requirement isn't re-broken.

## Testing
- `pnpm check` — green across all 36 projects, 100% coverage thresholds met.
- Added server test asserting the session cookie carries `Max-Age=SESSION_TTL_SECONDS`.
- Added iOS unit tests for the Chrome-first open: Chrome-opens → no fallback; Chrome-can't-open → falls back to https; and the core hands the raw https URL to the seam. (iOS suite runs via `make test` on macOS/simulator — not executable in this Linux CI environment.)

https://claude.ai/code/session_01PSa5oav1spo7JN4uTqThDQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01PSa5oav1spo7JN4uTqThDQ)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,6 +24,10 @@ Workflow for a canary failure:
 3. Re-run the canary locally until the failing source passes — never commit until it does.
 4. Only then push and watch CI.
 
+### iOS login is Chrome-first
+
+When the iOS app opens the browser to log in, it must reuse the user's existing **Chrome** web session — most users browse in Chrome but never change the iOS default-browser setting, so their OS default stays **Safari**, where they are not signed in. The app therefore opens the authorize URL via Chrome's `googlechromes://` scheme ([`openAuthorizeURLChromeFirst`](./projects/ios-readplace/App/WebAuthOpeners.swift)) and must **not** fall back to the OS default browser when Chrome is installed — the fallback fires only when the system reports Chrome can't be opened (not installed). Do not reintroduce a `canOpenURL` pre-check to decide the browser (a false-negative probe silently routes login into Safari and regresses session reuse); let the actual open result decide. Reuse also depends on the `hutch_sid` session cookie being **persistent** — it carries `maxAge: SESSION_COOKIE_MAX_AGE_MS` (bounded by the server session TTL) so a Chrome login survives Chrome fully closing; do not downgrade it back to a bare session cookie.
+
 ## Architecture Guidelines
 
 ### Brand & Design Guidelines
--- a/projects/hutch/src/runtime/web/auth/auth.page.ts
+++ b/projects/hutch/src/runtime/web/auth/auth.page.ts
@@ -51,7 +51,7 @@ import { LoginSchema } from "./auth.schema";
 import { LoginPage, SignupPage, VerifyEmailPage } from "./auth.component";
 import { extractReturnUrl, parseReturnUrl } from "./parse-return-url";
 import { baseCookieOptions } from "../cookie-options";
-import { SESSION_COOKIE_NAME } from "@packages/web-session";
+import { SESSION_COOKIE_MAX_AGE_MS, SESSION_COOKIE_NAME } from "@packages/web-session";
 import { buildVerificationEmailHtml } from "./verification-email";
 import { flattenZodErrors } from "./flatten-zod-errors";
 import { initFetchUserCount } from "./fetch-user-count";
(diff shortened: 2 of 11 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #909
- **Commit**: 072fef5cc908f210ed3bb8a17f5740505d0c5fa1

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `LSApplicationQueriesSchemes` Chrome entries are now dead config
`projects/ios-readplace/App/Info.plist:55-59` still declares `googlechrome`/`googlechromes` under `LSApplicationQueriesSchemes`. Now that the `canOpenURL` pre-check is removed, that key is no longer consulted — it only governs `UIApplication.canOpenURL(_:)`, not `open(_:options:completionHandler:)`. Harmless to keep, but it's dead configuration that could mislead a future reader into thinking a `canOpenURL` probe still exists. Optional cleanup.

### 2. No `Max-Age` assertion on the Google login path
The new persistent-cookie assertion (`Max-Age=${SESSION_TTL_SECONDS}`) was added only to `auth.route.test.ts` (password login). `google-auth.page.ts` sets `hutch_sid` with the same `sessionCookieOptions` at 5 sites, so line coverage is met, but `google-auth.route.test.ts` only asserts `hutch_sid=` is present — not that it's persistent. A parallel `Max-Age` assertion (via the existing `cookiesFrom` helper) would lock in the behavior for the Google path and guard a future regression that touches only that file.

### 3. Cookie persistence now applies to all web logins, not just iOS Chrome
The `hutch_sid` change in `src/packages/web-session/src/session-cookie.ts` converts a bare session cookie into a 7-day persistent cookie for every browser (desktop Safari/Chrome web included), not only the iOS Chrome-first flow. This is the intended mechanism, is documented in `CLAUDE.md`, and is strictly safer than before (bounded by the same server TTL; `httpOnly`+`secure`+`sameSite=lax`). Flagged only so the broader blast radius is a conscious acceptance: users who relied on "closing the browser logs me out" now stay signed in for up to 7 days.

PR #919, opened 2026-07-04
fix(ios-readplace): remove hard-coded server internals from the client

FagnerMartinsBrack:
## Purpose

The iOS app is meant to be a **generic hypermedia client**: it knows only the entry point, the Siren media type, and the action names / link rels it supports — everything else (URLs, methods, field values, presentation) it must **discover from the server or tolerate**. A review found a cluster of places where the client instead hard-coded server internals or duplicated server domain logic, so a routine server-side change would break the app or force an App Store release.

This PR fixes the **independent, client-only** items from that review (the ones that need no coordinated server change). Each removes a concrete coupling and is individually shippable.

### A4 — Inject *all* bootstrap cookies; scrub by host, not by name
**Problem:** `bootstrapSession()` selected the reader session cookie by the hard-coded name `hutch_sid`, and sign-out deleted it by name. If the server ever renamed its session cookie (a server-internal change by definition), cookie injection would return nothing and **every in-app reader open would fail** until an App Store release.
**Fix:** the client now forwards *every* cookie the `/auth/session` response set into the reader web view, and scrubs sign-out cookies by the app's own server **host**. `AppConfig.sessionCookieName` is gone. The client no longer needs to know the cookie's name.

### B1 — Re-read the collection on a reader status change instead of inferring "read"
**Problem:** the reader bridge fired on the *presence* of a `status` field and the client unconditionally treated it as "marked read", dropping the row. But `update-status` is a server **toggle** (`value: "read"` *or* `"unread"`), so a toggle back to unread wrongly removed the item.
**Fix:** on a reader status change the client re-reads the collection and adopts the server's truth (a shallow list) rather than inferring a direction. The row leaves only if the server's unread-only collection no longer lists it. (A deep-scrolled list holds its position and reconciles on the next pull-to-refresh.) The durable server-owned signal is a follow-up PR; this removes the incorrect inference now.

### C1 — Tolerate an evolving/malformed collection warning
**Problem:** `SirenWarning.code` was a **required** decode key, decoded atomically, even though the client reads only `warning.message`. Dropping or renaming `code` server-side — a property no client reads — would throw and **blank the entire collection**. The non-fatal warning channel could take down the page.
**Fix:** `code` is now optional and the warning decodes leniently, so an evolving or malformed warning degrades to *no banner*, never a blank page.

### C2 — Exclude a link from controls when *any* rel is structural
**Problem:** the client keyed structural-link detection on `rel.first` only. A multi-rel link like `["alternate", "next"]` keyed as `alternate`, escaped the structural filter, and would render as a tappable control *while also* being followed for pagination.
**Fix:** a link carrying any structural rel (`self`/`root`/`prev`/`next`/`item`) is never a control, mirroring the browser extension's walker.

### C3 — Surface an item's other semantic links as controls
**Problem:** each item reduced its links to a single `read` pick; every other semantic entity link (a future `share`, `summary`, …) was discarded and could never render — the exact failure the affordance loop exists to prevent.
**Fix:** the row now surfaces every non-structural, non-`read` entity link as a control (keeping `read` as the primary tap), so a newly-advertised item link renders with no client change.

### D — Drop comments that reintroduce URL coupling in prose
Removed comments naming concrete server routes/methods/symbols (`/queue`, `/auth/session`, `/oauth/token`, `SESSION_COOKIE_NAME`, `isFile`, `/login`, `/signup`). The code discovers these from the response; a route baked into a comment rots silently and re-teaches the coupling the contract removes.

## Not in this PR (need coordinated server changes)
Split out because they can't be done client-only without regressing test-pinned behaviour:
- **A2** — discover the session-bootstrap action instead of hard-coding `POST /auth/session`.
- **B2** — a server-owned reader mark-read signal (replaces the htmx-event sniffing; completes B1).
- **B3** — a server-provided read-state signal (the client hard-codes the `"read"` status literal in two places, both pinned by explicit tests encoding intended dual-signal/toggle behaviour).

## Testing
Full XCTest suite green locally (**207 tests, 0 failures**) plus the STAGING smoke pass, run exactly as CI runs `make test`.

### DIFF
--- a/projects/ios-readplace/App/AffordancePresentation.swift
+++ b/projects/ios-readplace/App/AffordancePresentation.swift
@@ -121,13 +121,34 @@ extension Affordance {
 		return fields.allSatisfy { $0.value != nil }
 	}
 
+	/// Whether this affordance is a link carrying any structural navigation rel
+	/// (`self`/`root`/`prev`/`next`/`item`) — client plumbing the client follows for
+	/// its own pagination/identity/item resolution, never a user control. Tests every
+	/// rel, not just the presentation token, so a multi-rel link like
+	/// `["alternate", "next"]` can't slip through as a tappable control while the
+	/// client also follows it for paging.
+	var isStructuralLink: Bool {
+		guard case let .link(link) = invocation else { return false }
+		return link.rel.contains { Affordance.structuralRels.contains($0) }
+	}
+
+	/// Whether the row surfaces this link as a discrete control: a semantic link that
+	/// is neither structural plumbing nor the `read` rel (already the row's primary
+	/// tap target). Keeps a future item link (e.g. `share`) rendering as a control
+	/// instead of being discarded, without double-rendering `read`.
+	var isSemanticControlLink: Bool {
+		guard case let .link(link) = invocation else { return false }
+		return !link.rel.contains { Affordance.structuralRels.contains($0) || $0 == "read" }
+	}
+
 	/// Whether the toolbar should surface this affordance as a control: it must be
-	/// both presentable in the toolbar (a structural navigation link or a
-	/// capture-only save is excluded by its presentation) and actually invokable
+	/// presentable in the toolbar (a structural navigation link or a capture-only
+	/// save is excluded), not carry any structural rel, and be actually invokable
 	/// from a bare control (a field-requiring action with no server value is
 	/// excluded).
 	var isToolbarControl: Bool {
-		presentation.isToolbarControl && isInvokableByBareControl
+		guard !isStructuralLink else { return false }
+		return presentation.isToolbarControl && isInvokableByBareControl
 	}
 
 	/// Whether invoking this affordance removes the item it acts on from the
@@ -148,12 +169,17 @@ extension Affordance {
 
 extension Article {
 	/// The advertised item affordances a row surfaces as swipe and accessibility
-	/// controls, filtered to those a bare control can actually invoke. Like the
+	/// controls: every action a bare control can actually invoke, plus every semantic
+	/// link that is neither structural plumbing nor the `read` tap target — so a
+	/// future item link (e.g. `share`) renders instead of being discarded. Like the
 	/// toolbar, the row drops a field-requiring action with no server value so a
 	/// future such item action is never rendered as a swipe that errors on tap. The
 	/// selection lives here, beside the symmetric toolbar rule
-	/// (`Affordance.isToolbarControl`) and the shared predicate it reuses
-	/// (`isInvokableByBareControl`), so the row's choice of controls is unit-testable
-	/// without standing up a view.
-	var rowControls: [Affordance] { affordances.filter(\.isInvokableByBareControl) }
+	/// (`Affordance.isToolbarControl`) and the shared predicates it reuses
+	/// (`isInvokableByBareControl`, `isSemanticControlLink`), so the row's choice of
+	/// controls is unit-testable without standing up a view.
+	var rowControls: [Affordance] {
+		affordances.filter(\.isInvokableByBareControl)
+			+ links.compactMap(Affordance.init(link:)).filter(\.isSemanticControlLink)
+	}
 }
--- a/projects/ios-readplace/App/AppSession.swift
+++ b/projects/ios-readplace/App/AppSession.swift
@@ -118,7 +118,8 @@ final class AppSession: ObservableObject {
 	private static func removeReaderWebStoreData() async {
 		let store = WKWebsiteDataStore.default()
 		if let host = URL(string: AppConfig.serverBaseURL)?.host {
-			for cookie in await store.httpCookieStore.allCookies() where cookie.domain.contains(host) {
+			for cookie in await store.httpCookieStore.allCookies()
+			where cookie.domain == host || cookie.domain.hasSuffix("." + host) {
 				await store.httpCookieStore.delete(cookie)
 			}
 		}
--- a/projects/ios-readplace/App/AppSession.swift
+++ b/projects/ios-readplace/App/AppSession.swift
@@ -28,8 +28,8 @@ final class AppSession: ObservableObject {
 
 	// Defaults to an ephemeral configuration so the API/OAuth sessions keep their
 	// cookie jar in an isolated, in-memory store rather than process-wide
-	// `HTTPCookieStorage.shared` — the minted `hutch_sid` reader cookie must not
-	// linger in the shared jar where it would outlive a sign-out.
+	// `HTTPCookieStorage.shared` — the minted reader session cookie must not linger
+	// in the shared jar where it would outlive a sign-out.
 	//
 	// `wipeReaderWebStore` defaults to the real WebKit deletion; it's the
 	// OS-boundary seam tests replace with a spy.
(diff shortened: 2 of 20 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #919
- **Commit**: 90a1ecc0685d4af9761b86d65e4d5418787784d8

FagnerMartinsBrack:
### Review: remove hard-coded server internals from the iOS client

This is a coherent, well-reasoned PR that does what it says: each item removes a real coupling and is independently shippable. I traced every change end-to-end. The B1 fix is correct (`convergeWithServer(droppingId: nil)` no-ops for a deep-scrolled list and adopts server truth for a shallow one — verified in `ReadingListViewModel.swift:187`), C2's `isStructuralLink` is exercised by the existing toolbar test, C3's link controls are actually invocable (`activate` → `viewModel.open(link:)` in `ReadingListView.swift:204`), and the D comment removals align with the code-comments skill. No security or data-loss concerns: the reader only ever receives the server's own same-host cookies, and sign-out over-deletes rather than under-deletes.

None found

## Medium Priority Issues

### 1. The headline C3/C1 behaviors ship without tests
The two most novel guarantees have no test pinning them, and the iOS suite has **no coverage gate** (`make test` runs `xcodebuild test` with no `-enableCodeCoverage`/threshold), so CI stays green regardless.

- **C3 — `Affordance.isSemanticControlLink` is never invoked by any test.** The only `rowControls` test (`SirenDecodingTests.swift:52`) builds an article with no `links`, so the new property (`AffordancePresentation.swift:139`) and the item-links branch of `rowControls` (`AffordancePresentation.swift:183`) are never exercised. Nothing pins: a `share` link surfaces as a control; the `read` link is not double-surfaced; a structural item link is excluded; a multi-rel `["alternate","next"]` link is excluded.
- **C1 — the degradation paths are untested.** `testCollectionWarningSurfaced` only covers a warning *with* `code`. Nothing covers a warning missing `code` still surfacing its `message`, or a malformed warning degrading to `warning == nil` while the rest of the collection still decodes — the "never blank the page" claim in `CollectionProperties.init(from:)` (`SirenModels.swift:179`).

## Low Priority Suggestions

### 1. `removeReaderWebStoreData` matches the cookie host by substring
`AppSession.swift:123` — `cookie.domain.contains(host)` over-matches (`readplace.com.attacker.example`). Only over-deletes on sign-out, so not a bug; a precise predicate is `cookie.domain == host || cookie.domain.hasSuffix("." + host)`.

### 2. The "no cookie ⇒ failed mint" guard is weaker than its doc comment
`ReadplaceAPI.swift:382` returns every cookie the shared session jar holds for the URL; a `/auth/session` that sets nothing could pass `guard !cookies.isEmpty` on a stale same-host cookie. Harmless today; snapshot the jar / response `Set-Cookie` to keep the strong guarantee.

### 3. `Affordance.init(link:)` still keys presentation on `rel.first`
`SirenModels.swift:438` — control-worthiness is now multi-rel-aware but the presentation token isn't. Cosmetic only (structural links are pre-filtered; unknown tokens fall back to neutral).

### 4. B1 drops the optimistic removal (intended, flagging for awareness)
`ReadingListViewModel.swift:145` — a just-read row lingers for one `loadQueue()` round-trip before the server collection replaces it. Correct trade for the toggle-direction bug; B2/B3 are the durable follow-up.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #919
- **Branch**: ios/independent-contract-fixes

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #919
- **Commit**: e469a72bc3a1e597af63e4de8db2c0f4eee75ad8

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

The prior review's sole medium — the C3/C1 test-coverage gap — is resolved by commit `e469a72b`. I traced all three added tests against the real decode logic:

- **C3** `testRowControlsSurfaceANonStructuralItemLinkButNeverReadOrStructuralLinks` genuinely exercises `Affordance.isSemanticControlLink` (`AffordancePresentation.swift:139`) and the link branch of `rowControls` (`AffordancePresentation.swift:183`): a `share` link surfaces, `read` stays the primary tap via `readHref` (not double-surfaced), and both a structural `item` link and a multi-rel `["alternate","next"]` link are excluded. The multi-rel case pins C2's "any rel is structural" logic.
- **C1** `testCollectionWarningMissingCodeStillSurfacesItsMessage` and `testMalformedCollectionWarningDegradesToNoBannerButKeepsTheCollection` cover both degradation paths — a code-less warning still shows its message, and a malformed warning degrades to `warning == nil` while the collection alongside it still decodes (`["good"]`), pinning the "never blank the page" claim.

Assertions are positive and non-vacuous (the malformed-warning test pairs `XCTAssertNil(page.warning)` with a positive `page.articles == ["good"]`, so nil can't pass for the wrong reason). The commit changed no production code, so it introduces no new production risk. Cross-client contract verified: the iOS `structuralRels` set and "any rel is structural ⇒ not a control" logic match the browser-extension walker (`siren-reading-list.ts:291`) exactly.

## Low Priority Suggestions

### 1. `removeReaderWebStoreData` matches the cookie host by substring
`AppSession.swift:121` — `cookie.domain.contains(host)` over-matches (a cookie for `readplace.com.attacker.example` contains `readplace.com`). On sign-out this only ever *over-*deletes (privacy-positive, and such a foreign cookie is unlikely in the app's own WebKit store), so it's not a bug — but a precise predicate can't drift: `cookie.domain == host || cookie.domain.hasSuffix("." + host)`. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FApp%2FAppSession.swift%3A121%20on%20branch%20ios%2Findependent-contract-fixes%2C%20replace%20cookie.domain.contains%28host%29%20in%20removeReaderWebStoreData%20with%20cookie.domain%20%3D%3D%20host%20%7C%7C%20cookie.domain.hasSuffix%28%22.%22%20%2B%20host%29%20so%20sign-out%20scrubbing%20does%20not%20over-match%20a%20domain%20like%20readplace.com.attacker.example&repo=Readplace/readplace.com)

### 2. The "no cookie ⇒ failed mint" guard is weaker than its doc comment
`ReadplaceAPI.swift:216` — `sessionCookies` returns every cookie the shared session jar holds for the URL, but the `bootstrapSession` doc comment (`ReadplaceAPI.swift:191`) says "a response that sets no cookie is a failed mint." If the server ever set a same-host cookie on an earlier endpoint, a `/auth/session` that sets nothing would still pass `guard !cookies.isEmpty`. Harmless today (the jar is ephemeral and bootstrap is realistically the only cookie-setter), but the guarantee is now assumption-based where the old name-filter made it explicit.

### 3. `Affordance.init(link:)` still keys presentation on `rel.first`
`SirenModels.swift:438` — C2/C3 correctly made *control-worthiness* inspect every rel, but `token = rel.first` still keys the icon/tint off the first rel alone. No bug (structural links are filtered out before presentation, and an unknown token falls back to the neutral glyph), so this is purely cosmetic: a future `["alternate","share"]` link would render `ellipsis.circle` rather than a share glyph. Noted only because it's the same first-rel assumption this PR is otherwise removing.

### 4. A B1 comment now describes the pre-B1 disposition
`ReaderWebView.swift:128` — the `ReaderBridge.script` doc says the bridge "signals the native side, which closes the sheet and **drops the row**." After B1, `onMarkedRead` routes to `readerStatusChanged()` → `convergeWithServer`, so the row leaves *only* if the server's unread-only collection no longer lists it (a toggle-back-to-unread keeps it). "Drops the row" now overstates the disposition. Minor comment staleness (the B2 follow-up will replace this bridge entirely). [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FApp%2FReaderWebView.swift%20around%20line%20128%20on%20branch%20ios%2Findependent-contract-fixes%2C%20update%20the%20ReaderBridge.script%20doc%20comment%20that%20says%20the%20native%20side%20closes%20the%20sheet%20and%20drops%20the%20row%3A%20after%20B1%2C%20readerStatusChanged%20converges%20with%20the%20server%2C%20so%20the%20row%20leaves%20only%20if%20the%20server%27s%20unread-only%20collection%20no%20longer%20lists%20it.%20Reword%20drops%20the%20row%20to%20reflect%20converge-with-server%20reconciliation&repo=Readplace/readplace.com)

### 5. C2's toolbar multi-rel path isn't directly tested
The new C3 test pins the multi-rel exclusion through `isSemanticControlLink` (the row predicate), and `testToolbarExcludesCaptureOnlySavesAndStructuralLinks` covers `isToolbarControl` with *single*-rel structural links. Nothing directly exercises `isToolbarControl` → `isStructuralLink` with a multi-rel link like `["alternate","next"]`. The two predicates share `structuralRels` and the same `rel.contains { … }` shape, so the behavior is pinned in substance and the regression risk is low — an optional completeness test, not a gap that matters.

FagnerMartinsBrack:
@claude Please address the following **low-priority** review suggestion. It's a clean, in-scope precision fix — no design decision needed.

**Tighten the cookie-host match in `removeReaderWebStoreData`** — `projects/ios-readplace/App/AppSession.swift`

The sign-out scrub matches cookies with `cookie.domain.contains(host)`, which over-matches any domain that merely *contains* the host string (e.g. `readplace.com.attacker.example`). Replace the `where` predicate with an exact-or-subdomain match:

```swift
where cookie.domain == host || cookie.domain.hasSuffix("." + host)
```

This still matches the exact host and every legitimate subdomain (`.readplace.com`, `app.readplace.com`) while excluding spoofed foreign domains, so the sign-out wipe doesn't regress. Please follow CLAUDE.md and run the relevant iOS checks before pushing.

FagnerMartinsBrack:

## Context
- **PR**: #919
- **Commit**: d2a5aed4d09d4d2d172491cce652b171139020d7

FagnerMartinsBrack:
### Review — PR #919 (commit `d2a5aed4`)

The only change since the last review (`e469a72b` → `d2a5aed4`) is the one-line cookie-host predicate in `AppSession.swift`, which **resolves the previous review's low-priority #1**. I verified the fix and re-traced the surrounding changes; nothing regressed.

**The cookie-host fix is correct and complete.** `AppSession.swift:121` now scrubs sign-out cookies with `cookie.domain == host || cookie.domain.hasSuffix("." + host)`:
- Matches the exact host (`readplace.com`), the leading-dot domain cookie (`.readplace.com`), and legitimate subdomains (`app.readplace.com`).
- Excludes the substring spoof (`readplace.com.attacker.example`) the old `contains(host)` reached.
- The new match set is a **strict subset** of the old one, and the only domains removed from the wipe are foreign origins the app never authenticates against — so there is **no under-deletion regression** for the app's own origins.

I also re-confirmed the C3/C1 tests added in `e469a72b` are correct and non-vacuous, and the cross-client contract is intact — the browser-extension walker (`siren-reading-list.ts:305`) excludes a link when `rel.some(...)` is structural over the identical `{self, root, prev, next, item}` set, matching iOS `isStructuralLink`/`isSemanticControlLink`.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

Standing (non-blocking) items carried from the last review; #1 is now fixed by this commit.

### 1. `ReaderBridge.script` doc comment describes the pre-B1 disposition
`ReaderWebView.swift:130` — the doc says the native side "closes the sheet and **drops the row**." After B1, `onMarkedRead` (`ReadingListView.swift:72`) routes to `readerStatusChanged()` → `convergeWithServer(droppingId: nil)`, so the row leaves *only* if the server's unread-only collection no longer lists it. "Drops the row" overstates it. Minor comment staleness (B2 will replace this bridge).

### 2. The "no cookie ⇒ failed mint" guard is weaker than its doc comment
`ReadplaceAPI.swift:216` — `sessionCookies` returns every cookie the shared jar holds for the URL, but the `bootstrapSession` doc (`ReadplaceAPI.swift:191`) says "a response that sets no cookie is a failed mint." Harmless today (ephemeral jar, bootstrap is the only cookie-setter), but the guarantee is now assumption-based.

### 3. `Affordance.init(link:)` still keys presentation on `rel.first`
`SirenModels.swift:438` — control-worthiness is now multi-rel-aware but `token = rel.first` isn't. Cosmetic (structural links pre-filtered; unknown tokens fall back to the neutral glyph).

### 4. C2's toolbar multi-rel path isn't directly tested
The new C3 test pins multi-rel exclusion through `isSemanticControlLink`; `isToolbarControl` → `isStructuralLink` isn't directly exercised with a multi-rel link. Both share `structuralRels` and the same `rel.contains { … }` shape, so risk is low — an optional completeness test.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28741531518
- **PR**: #919
- **Branch**: ios/independent-contract-fixes

## Failed Jobs

FagnerMartinsBrack:

PR #921, opened 2026-07-04
test(ios-readplace): make the share extension logic testable

FagnerMartinsBrack:
> Stacked on #920 (base = `ios/coverage-gate`). Review/merge after it.

## Purpose

`ShareViewController` and `ShareURLExtractor` were compiled **only** into the extension target, which no test target sees. So their real branching logic — the multi-item payload scan, the URL-item type fallbacks (URL / Data / String / plain-text), the PDF size ceiling, and the outcome→status mapping — had **zero coverage and was structurally untestable**. A host delivering an unexpected payload shape would regress silently.

## What changed & the value

- **`ShareURLExtractor` → `Shared/`**: it is Foundation-only, so moving it makes it compile into the app target that `@testable import Readplace` can reach. `extract(from:)` is split into a thin `NSExtensionContext` wrapper plus an **item-list core** a test drives with fabricated `NSItemProvider`s — no live extension context (which no test can construct) required.
- **Extract `ShareStatusPresentation`**: the seven-case outcome→(message, SF Symbol, tone) decision — including joining a refusal's messages and choosing error-vs-warning — moves out of the UIKit controller into a **pure value in `Shared/`**. `ShareViewController` keeps only the trivial tone→`UIColor` mapping at the boundary.
- **Coverage**: `ShareURLExtractor` now at **96%** (URL/Data/String/plain-text extraction, multi-item split, PDF provider + suggested-name fallback, the byte ceiling), `ShareStatusPresentation` at **100%** (every outcome, the refusal join, both tone branches).
- **Baseline** (`scripts/coverage-baseline.json`): `ShareURLExtractor` moves out of the exclusions to a 96% floor; `ShareStatusPresentation` pinned at 100%. The gate now enforces **21 files** (was 19), **9 excluded** (was 10).

## Testing
Full suite green with the gate active locally (**227 tests**, +20; `iOS coverage gate passed: 21 files at or above floor, 9 excluded`), plus the STAGING smoke pass.

### DIFF
--- a/projects/ios-readplace/Makefile
+++ b/projects/ios-readplace/Makefile
@@ -26,9 +26,13 @@ open: generate
 # someone builds `make ipa-staging` by hand.
 SIM ?= platform=iOS Simulator,name=$(shell xcrun simctl list devices available | grep -m1 -oE 'iPhone [0-9]+')
 SIM_DEVICE = $(shell printf '%s' '$(SIM)' | sed -E 's/.*=//')
+COVERAGE_RESULT = build/TestResults.xcresult
 test: generate
 	xcrun simctl bootstatus '$(SIM_DEVICE)' -b
-	xcodebuild test -project Readplace.xcodeproj -scheme Readplace -destination '$(SIM)'
+	rm -rf $(COVERAGE_RESULT)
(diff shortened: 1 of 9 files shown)


FagnerMartinsBrack:

PR #922, opened 2026-07-04
test(ios-readplace): backfill untested logic seams and ratchet coverage

FagnerMartinsBrack:
> Stacked on #921 (base = `ios/share-extension-testable`). Review/merge after it.

## Purpose

Several logic seams had error/edge paths with **no coverage** — the kind of gaps that surface as a silent regression. E3 covers them and ratchets the coverage floors up (the gate PR #920 seeded them at today's numbers).

## What got covered (and the floors raised)

| File | Before | After | New coverage |
|---|---|---|---|
| `OAuthService` | 90% | **98%** | malformed 200 body → `.malformedResponse`; 200 without `refresh_token` → `.malformedResponse`; refresh with no stored token → `.noRefreshToken` (no request made); revoke with no token → skips the network, still clears |
| `KeychainTokenStorage` | 57% | **98%** | a real **simulator-keychain round trip** through the production path (add / read / update / remove, key isolation) — the `UserDefaults` double never exercised the Keychain code |
| `TokenStore` | 85% | **99%** | extracted the embedded-provisioning-profile parsing into a pure `parseAppGroupId(fromProvisioningProfile:)` and covered it with fixtures (valid, no plist, no app-group entitlement, empty array); only the `Bundle.main` read stays untested |
| `ReadingListViewModel` | 94% | **95%** | a 401 whose refresh also fails routes to the logout callback with **no** error banner; a collection warning populates `warningText` |
| `AppSession` (completeSignIn) | 79% | **83%** | a callback carrying `error=` is denied; a callback with no `code` is `.missingCode`; a failed code exchange surfaces as a failure — none logging in |

The one production change is the `parseAppGroupId` extraction (behaviour-preserving) so the profile parsing is unit-testable without a real `Bundle.main`.

## Still deferred (noted for a later ratchet)
`HTMLCaptor`'s WKWebView state machine and the reader `htmx` JS detector remain in the OS-boundary exclusions — covering them needs a `JSContext`/WebKit harness, a larger change than this backfill. `AppSession` is at 83% (the remaining lines are the concurrent WebKit-wipe logout choreography).

## Testing
Full suite green with the gate active (**242 tests**, +15; `iOS coverage gate passed: 21 files at or above floor, 9 excluded`), plus the STAGING smoke pass.

### DIFF
--- a/projects/ios-readplace/Shared/TokenStore.swift
+++ b/projects/ios-readplace/Shared/TokenStore.swift
@@ -61,10 +61,23 @@ struct TokenStore {
 		guard
 			let url = Bundle.main.url(forResource: "embedded", withExtension: "mobileprovision"),
 			let data = try? Data(contentsOf: url),
+			let group = parseAppGroupId(fromProvisioningProfile: data)
+		else { return AppConfig.appGroupId }
+		return group
+	}()
+
+	/// Extracts the first application-group entitlement from an embedded
+	/// `.mobileprovision` blob, or nil when it can't be read. The profile is a
+	/// CMS-signed container with a plist inside; the signature isn't verified here
+	/// (the OS already did at install), so this just scans out the `<?xml…</plist>`
+	/// slice and reads the entitlement. Pulled out of `resolvedAppGroupId` so the
+	/// parsing is unit-testable without a real `Bundle.main`.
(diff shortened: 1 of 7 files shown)


FagnerMartinsBrack:

PR #923, opened 2026-07-04
test(ios-readplace): assert exact values instead of existence/negatives

FagnerMartinsBrack:
> Stacked on #922 (base = `ios/untested-seams`). Review/merge after it. Test-only, no production changes.

## Purpose

Three tests passed **for the wrong reason** — an existence check or a negative assertion that a buggy implementation would still satisfy. This replaces them with assertions that pin the actual value, per the repo's test-driven-design conventions ("Test Behavior, Not Element Existence" / "Avoid Negative Test Assertions").

## What changed & why it matters

- **`SirenDate` parsing** — `XCTAssertNotNil(SirenDate.parse(…))` passed even if the parser returned the *wrong instant* (dropped the fractional part, misread the zone). Now it asserts the parsed value equals a known UTC instant, and that the fractional variant is exactly 123 ms later. Same fix for the fixture's `savedAt` (was `XCTAssertNotNil`).
- **PKCE challenge** — `testChallengeIsURLSafe` proved the challenge only by what it *didn't* contain (`+`, `/`, `=`), so a wrong-length or wrong-alphabet challenge passed. Now it asserts the exact length (43) and a positive base64url-alphabet check, mirroring the verifier test.
- **Login/Signup authorize** — the tests pinned the literal host `"readplace.com"`, which is *why* the full suite can't run under the STAGING build. They now derive the expected host from `AppConfig.serverBaseURL` (the same config the code uses), making them environment-agnostic — a prerequisite for a later widening of `make test-staging`.

## Testing
Full suite green (**242 tests**), coverage gate passes, STAGING smoke passes.

### DIFF
--- a/projects/ios-readplace/Tests/LoginFlowTests.swift
+++ b/projects/ios-readplace/Tests/LoginFlowTests.swift
@@ -23,7 +23,7 @@ final class LoginFlowTests: XCTestCase {
 		let request = makeService(store: TestSupport.loggedInStore()).makeNativeLoginAuthorizationRequest()
 
 		let components = URLComponents(url: request.url, resolvingAgainstBaseURL: false)!
-		XCTAssertEqual(components.host, "readplace.com")
+		XCTAssertEqual(components.host, URL(string: AppConfig.serverBaseURL)?.host)
 		XCTAssertEqual(components.path, "/oauth/authorize")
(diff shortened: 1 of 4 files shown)


FagnerMartinsBrack:

PR #924, opened 2026-07-04
test(ios-readplace): extract per-item routing and web 4xx policy into tested seams

FagnerMartinsBrack:
> Stacked on #923 (base = `ios/test-quality`). Review/merge after it.

## Purpose

Two behavioural decisions lived inside SwiftUI/WebKit glue with **no unit-test seam**, unlike their already-extracted twins (`ToolbarRoute`, `HTMLCaptor.navigationResponseDecision`). This extracts them into pure, gated values, leaving the views with only their effect/plumbing.

## What changed & the value

- **`ItemRoute`** lifts the per-item routing out of `ReadingListView.activate`: a link **opens**, a destructive action **confirms**, any other action **invokes**. The contract-relevant behaviours — a link-only item control opens rather than being silently dropped, and a destructive control never invokes without a confirmation — are now unit-tested without standing up a view. Mirrors `ToolbarRoute` exactly.
- **`WebResponsePolicy`** lifts the "a 4xx/5xx status is a failure" rule out of the `WebPageView` navigation delegate. WKWebView delivers error statuses through `didFinish` (not `didFail`), so the web view would otherwise paint the server's error body; the policy is tested at the **399/400 boundary**.

Both new files are gated at **100%** (the coverage gate requires a new logic file to be fully covered); `ReadingListView` and `WebPageView` keep only their effect/plumbing and stay in the OS-boundary exclusions.

## Testing
Full suite green with the gate (**249 tests**, +7; `iOS coverage gate passed: 23 files at or above floor, 9 excluded`), plus the STAGING smoke pass.

### DIFF
--- a/projects/hutch/src/runtime/web/api/collection-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.test.ts
@@ -258,6 +258,21 @@ describe("toArticleCollectionEntity", () => {
 		expect(saveArticlesAction?.fields?.map((f) => f.name)).toEqual(["manifest", "content"]);
 	});
 
+	it("advertises a create-session action so a client discovers the reader session mint", () => {
+		const result: FindArticlesResult = {
+			articles: [],
+			total: 0,
+			page: 1,
+			pageSize: 20,
+		};
+
+		const entity = toArticleCollectionEntity(result, {});
+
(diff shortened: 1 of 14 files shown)


FagnerMartinsBrack:

PR #926, opened 2026-07-04
feat(hutch,ios): emit an explicit read-state so the client stops hard-coding "read"

FagnerMartinsBrack:
> Stacked on #925 (base = `ios/discover-session-action`). Review/merge after it. **Server + client**, additive/non-breaking — deploy either order.

## Purpose

The iOS client derived an article's read indicator from the **status vocabulary** — `isRead = status == "read" || readAt != nil` — a domain literal baked into the client. It goes stale the moment the server's status vocabulary evolves (an `archived` state that's also "read", or keeping `readAt` after a mark-unread), leaving the checkmark disagreeing with server truth. Two explicit tests pinned the dual-signal derivation, so this couldn't be fixed client-only without a server signal — which is why it was split out of the earlier client-only PR.

## What changed & the value

- **Server** (`article-siren.ts`): emit an explicit `isRead` boolean property on the article entity (the value it already computes for the toggle label). Read-state is now **server-authored**, one property.
- **Client**: render read-state from the server's `isRead` when present, falling back to the old `status`/`readAt` derivation only for an **older server** that doesn't emit it (App Store back-compat). Once the server is deployed the fallback never runs.

The property is additive: an older client ignores it, a newer client prefers it. No breaking change.

## Scope note
The optimistic row-removal prediction on swipe-mark-read (`removesItemFromUnreadList`) still reads the `update-status` field value — that path is the client-side optimistic synthesis that was **deliberately kept** (it's not the read-state *rendering* this PR targets), so its literal is intentionally out of scope here.

## Testing
- Server: `article-siren` / `collection-siren` suites green (incl. new `isRead` assertions and the updated exact-property-key lists); 508 article/route/view tests pass; `pnpm check` passes.
- Client: full iOS suite green (**253 tests**) — the server `isRead` overrides the status literal in both directions, and the fallback derivation still holds when the server omits it; coverage gate + STAGING smoke pass.

### DIFF
--- a/projects/hutch/src/runtime/web/api/article-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/article-siren.test.ts
@@ -51,6 +51,7 @@ describe("toArticleSubEntity", () => {
 				status: "unread",
 				savedAt: "2026-03-04T10:00:00.000Z",
 				readAt: null,
+				isRead: false,
 			},
 			links: [
 				{ rel: ["read"], title: "Read", href: `/queue/${ARTICLE_ID}/view` },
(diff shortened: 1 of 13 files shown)


FagnerMartinsBrack:

PR #927, opened 2026-07-04
feat(hutch,ios): make the iOS app a generic hypermedia client + coverage gate (consolidated)

FagnerMartinsBrack:
Consolidates what were 9 stacked PRs (#919–#927) into **one PR targeting `main`** so CI runs and the whole thing is independently deployable. They were stacked because the changes share 14 files and form a dependency chain (everything builds on the base client refactor; the coverage work shares one baseline file), so they can't be separated into independent `main`-targeting PRs without cumulative diffs — they need coordination, hence one PR. The branch is merged up to current `main`; future drift is left to the conflict-resolver.

This makes the iOS app behave as a **generic hypermedia client** — it knows only the entry point, the Siren media type, and the action names / link rels it supports, and discovers or tolerates everything else — and closes the coverage/testability gaps a review found. Server changes are additive and non-breaking.

---

## 1. Client contract fixes — stop hard-coding server internals

- **Inject all bootstrap cookies (A4).** The reader session mint selected the cookie by the hard-coded name `hutch_sid`; now the client forwards *every* cookie the `/auth/session` response set and scrubs sign-out by server host. A server cookie rename no longer breaks the reader.
- **Re-read on a reader status change (B1).** `update-status` is a server toggle; the client used to infer "read" from a status POST and drop the row. It now re-reads the collection and adopts the server's truth, so a toggle back to unread no longer wrongly removes the item.
- **Tolerate an evolving warning (C1).** A collection `warning` decoded atomically with a required `code` the client never reads; dropping/renaming `code` would blank the whole page. Now `code` is optional and the warning decodes leniently → degrades to no banner.
- **Structural-rel safety (C2).** A link is excluded from controls when **any** of its rels is structural, not just the first — a multi-rel `["alternate","next"]` can't render as a control while also driving pagination.
- **Item semantic links (C3).** The row surfaces every non-structural, non-`read` entity link as a control, so a future item link (e.g. `share`) renders instead of being discarded.
- **Comment de-coupling (D).** Removed comments that named concrete routes/methods/symbols (`/queue`, `/auth/session`, `/oauth/token`, `SESSION_COOKIE_NAME`, `isFile`, `/login`, `/signup`) — they rot silently and reintroduce the URL coupling the contract removes.

## 2. Coverage & testability

- **A per-file coverage gate (E1).** `make test` now runs with `-enableCodeCoverage` and fails below a per-file floor (`scripts/check-coverage.py` + `coverage-baseline.json`) — the iOS suite previously ran with no coverage measurement at all. It's a ratchet: pure SwiftUI views and WebKit/OS-boundary glue are excluded; every other file must stay at or above its floor; a new logic file must be 100%.
- **Share Extension made testable (E2).** `ShareURLExtractor` moved into `Shared/` (with an item-list core tests can drive), and the outcome→status mapping extracted into a pure `ShareStatusPresentation` — both previously uncoverable in the extension-only target.
- **Untested seams backfilled (E3).** Keychain 57→98%, TokenStore 85→99% (extracted `parseAppGroupId`), plus OAuth malformed/refresh/revoke branches, `completeSignIn` error paths, and the VM 401→logout route.
- **Exact-value assertions (E4).** Replaced `XCTAssertNotNil`/negative assertions on dates, the PKCE challenge, and host pins with assertions that pin the actual value.
- **View decisions extracted (E5).** `ItemRoute` (per-item routing) and `WebResponsePolicy` (4xx handling) lifted out of the SwiftUI/WebKit glue into pure, gated seams.

## 3. Server-coordinated hypermedia fixes (server + client, additive)

- **Discover the session-mint action (A2).** Server advertises a `create-session` action on the collection; the client follows its href/method instead of hard-coding `POST /auth/session`, falling back to the fixed path for an older server.
- **Explicit read-state (B3).** Server emits an `isRead` boolean on the article; the client renders from it instead of hard-coding the `"read"` status literal, falling back to the status/`readAt` derivation only when the server omits it.
- **Server-owned reader bridge (B2).** The chromeless reader now posts the `markedRead` message to the WKWebView itself (guarded so it's inert in a browser); the app injects no script and no longer sniffs the front-end's htmx internals. `readplaceReader`/`markedRead` is the stable contract.

---

## Deploy notes
All server changes are additive/non-breaking (a client without a handler skips them). A2/B3 deploy in either order with the client; B2's server change should land before the app build that depends on it. Nothing invalidates existing installs.

## Testing
- **iOS:** full XCTest suite green (**255 tests**), per-file coverage gate passes (23 files gated, 9 view/OS-boundary excluded), STAGING smoke passes — validated locally exactly as CI runs `make test`.
- **Server:** the `collection-siren`, `article-siren`, and `queue.reader-platform` suites pass (incl. the new `create-session`, `isRead`, and reader-bridge assertions). Full `pnpm check` runs in CI.
- **Manual:** the reader-sheet mark-read close can't be unit-tested (live WKWebView + htmx) — please smoke-test "mark as read from inside the reader closes the sheet and drops the row" on a device/simulator before merge.

Supersedes and closes #919–#926.

### DIFF
--- a/projects/hutch/src/runtime/web/api/article-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/article-siren.test.ts
@@ -51,6 +51,7 @@ describe("toArticleSubEntity", () => {
 				status: "unread",
 				savedAt: "2026-03-04T10:00:00.000Z",
 				readAt: null,
+				isRead: false,
 			},
 			links: [
 				{ rel: ["read"], title: "Read", href: `/queue/${ARTICLE_ID}/view` },
@@ -72,6 +73,15 @@ describe("toArticleSubEntity", () => {
 		const subEntity = toArticleSubEntity(makeArticle());
 		const deleteAction = subEntity.actions?.find((a) => a.name === "delete");
 		expect(deleteAction).toBeUndefined();
+	});
+
+	it("emits isRead as an explicit boolean reflecting the read status", () => {
+		expect(toArticleSubEntity(makeArticle({ status: "unread" })).properties?.isRead).toBe(false);
+		expect(
+			toArticleSubEntity(
+				makeArticle({ status: "read", readAt: new Date("2026-03-04T12:00:00.000Z") }),
+			).properties?.isRead,
+		).toBe(true);
 	});
 
 	it("toggles an unread item to a server-driven Mark as read with the target value", () => {
--- a/projects/hutch/src/runtime/web/api/article-siren.ts
+++ b/projects/hutch/src/runtime/web/api/article-siren.ts
@@ -25,6 +25,10 @@ export function toArticleSubEntity(article: SavedArticle): SirenSubEntity {
 			status: article.status,
 			savedAt: article.savedAt.toISOString(),
 			readAt: article.readAt?.toISOString() ?? null,
+			// An explicit presentational read-state so a client renders the read
+			// indicator from one server-authored boolean rather than re-deriving it
+			// from the `status` vocabulary it would otherwise have to hard-code.
+			isRead,
 		},
 		links,
 		actions: [
--- a/projects/hutch/src/runtime/web/api/collection-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.test.ts
@@ -259,6 +259,21 @@ describe("toArticleCollectionEntity", () => {
 		expect(saveArticlesAction?.fields?.map((f) => f.name)).toEqual(["manifest", "content"]);
 	});
 
+	it("advertises a create-session action so a client discovers the reader session mint", () => {
+		const result: FindArticlesResult = {
+			articles: [],
+			total: 0,
+			page: 1,
+			pageSize: 20,
+		};
+
+		const entity = toArticleCollectionEntity(result, {});
+
+		const sessionAction = entity.actions?.find((a) => a.name === "create-session");
+		expect(sessionAction?.href).toBe("/auth/session");
+		expect(sessionAction?.method).toBe("POST");
+	});
+
 	it("includes search action with filter fields", () => {
 		const result: FindArticlesResult = {
 			articles: [],
--- a/projects/hutch/src/runtime/web/api/collection-siren.test.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.test.ts
@@ -95,6 +95,7 @@ describe("toArticleCollectionEntity", () => {
 			"status",
 			"savedAt",
 			"readAt",
+			"isRead",
 		]);
 	});
 
--- a/projects/hutch/src/runtime/web/api/collection-siren.ts
+++ b/projects/hutch/src/runtime/web/api/collection-siren.ts
@@ -138,6 +138,17 @@ export function toArticleCollectionEntity(
 					{ name: "url", type: "url" },
 				],
 			},
+			// Mints a browser session cookie from the bearer token so a client's
+			// in-app reader can load the cookie-authenticated reader page. Advertised
+			// (rather than the client hard-coding the route + method) so it can move
+			// without a client release; a client that doesn't open an in-app reader
+			// simply skips it. No fields: the route reads the bearer from the
+			// Authorization header.
+			{
+				name: "create-session",
+				href: "/auth/session",
+				method: "POST",
+			},
 		],
 	};
 }
--- a/projects/hutch/src/runtime/web/pages/queue/queue.page.ts
+++ b/projects/hutch/src/runtime/web/pages/queue/queue.page.ts
@@ -297,6 +297,37 @@ const APP_BACK_LINK = {
 	label: "← Back to queue",
 } as const;
 
+/** Server-authored bridge for the iOS in-app reader: when the reader's mark-read
+ * htmx request completes, tell the WKWebView so the native app closes the sheet
+ * and reconciles its list. Guarded on the WKWebView message handler, so it is inert
+ * in a normal browser that renders this same reader. Keeping the htmx detail here —
+ * rather than injected by the app — keeps the htmx coupling on the server that owns
+ * htmx; the app only registers the `readplaceReader` handler and reacts to the
+ * `markedRead` message, with no knowledge of the front-end's event shape. */
+const READER_MARK_READ_BRIDGE_SCRIPT = `<script>
+(function () {
+	var handlers = window.webkit && window.webkit.messageHandlers;
+	if (!handlers || !handlers.readplaceReader) { return; }
+	function hasStatusField(params) {
+		if (!params) { return false; }
+		if (typeof params.has === "function") { return params.has("status"); }
+		if (typeof params.get === "function") { return params.get("status") != null; }
+		return Object.prototype.hasOwnProperty.call(params, "status");
+	}
+	function isStatusChange(detail) {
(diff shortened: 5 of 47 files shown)


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28744046986
- **PR**: #927
- **Branch**: ios/server-reader-signal

## Failed Jobs

FagnerMartinsBrack:

## Context
- **PR**: #927
- **Branch**: ios/server-reader-signal

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **Run ID**: 28866941942
- **PR**: #927
- **Branch**: ios/server-reader-signal

## Failed Jobs

FagnerMartinsBrack:
Rebased `ios/server-reader-signal` onto latest `origin/main` (force-with-lease). The two prior merge-from-main commits were linearised, as expected.

Conflicts resolved (all reproducing the PR's own authoritative resolution from the pre-rebase merge tip, since `origin/main` had not touched these Swift sources since the old base):

- `projects/ios-readplace/App/ReadingListViewModel.swift` — kept B1 (`readerStatusChanged()` re-reads and adopts server truth) and pointed it at main's renamed `reloadAndAdopt()` helper (was `convergeWithServer`); removed main's optimistic `readerMarkedRead(id:)`/`readerDroppedId` mechanism that B1 replaces; simplified `handleWebSheetDismissal()` to drop the `readerDroppedId` bookkeeping (it already re-reads and adopts).
- `projects/ios-readplace/App/ReadingListView.swift` — `onMarkedRead` now calls `readerStatusChanged()` instead of the removed `readerMarkedRead(id:)`.
- `projects/ios-readplace/Tests/ReadingListViewModelTests.swift` — dropped the two now-obsolete `readerMarkedRead` tests, kept the single `testReaderStatusChanged…` B1 test; main's three account-deletion web-sheet-dismissal tests are retained.
- `projects/ios-readplace/Tests/SignupFlowTests.swift` — in `testChromeURLForHTTPSRewritesSchemeToGoogleChrome`, kept main's version (host asserted as the literal `readplace.com`, plus its scheme-assertion message): the input URL is hardcoded and `chromeURLForHTTPS` only rewrites the scheme, so the AppConfig-derived host from the PR's env-agnostic pass would be a false assertion here. The PR's `AppConfig.serverBaseURL`-derived host still applies to the config-derived tests.

The final rebased Swift sources/tests are byte-identical to the pre-rebase tip. `pnpm check` passes (37 projects, 52 tasks). Note: iOS Swift/XCTest coverage is validated by the separate ios-tests CI job, not by `pnpm check`.

FagnerMartinsBrack:

## Context
- **Attempt**: 3 of 5
- **Run ID**: 29010654357
- **PR**: #927
- **Branch**: ios/server-reader-signal

## Failed Jobs

FagnerMartinsBrack:

## Context
- **PR**: #927
- **Commit**: 995c5a9318f144f83cfd91950ae2d944093027f6

FagnerMartinsBrack:
Resolved without any gate change — CI is green (`ios-tests` + `web-tests` pass on 995c5a93).

The approval the CI fixer was blocked on (attempts 1–3) turned out to be unnecessary: all four files were lifted above their floors by raising real coverage, and `coverage-baseline.json` is untouched.

| File | CI before | Now | How |
|---|---|---|---|
| `SaveSharedPage.swift` | 91% < 93 | 98% | test for the generic `catch → .failed` path (undecodable queue body, no save attempted) |
| `ShareURLExtractor.swift` | 95% < 96 | 100% | the deterministic gaps were the nil-context overload + an `?? []` autoclosure (not the simulator-dependent coercion the fixer suspected); extracted a pure `coerceURL(from:)` seam + 4 tests — 100% on any simulator |
| `WebDialog.swift` | 47% < 100 | 100% | moved the `UIAlertController` glue into `ReaderWebView.swift`, its only caller and an already-excluded WKWebView boundary; the pure mapping stays measured |
| `LoginView.swift` | 88% < 91 | 97% | not "missed from excluded" — the floor was deliberately seeded; instead injected `session` + the `WebAuthFlow` factory (the `ReadingListView` precedent) so `startLogin`/`startSignup` are directly tested, incl. that Login vs Sign up start the correct authorize request |

Validated locally with the exact CI sequence (`make test`: 270 tests, 0 failures, gate + staging smoke pass) on the iOS 26.5 runtime before pushing. Follow-up candidates, deliberately not done here: ratchet the four floors up once a few CI runs confirm stability, and backfill direct tests for the C1/C2/C3 tolerance paths.

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `sessionCookies` reads a superset of "cookies this response set" — align the doc comment (or filter)

`ReadplaceAPI.sessionCookies(from:url:)` has two paths that return different sets:

```swift
private func sessionCookies(from response: HTTPURLResponse, url: URL) -> [HTTPCookie] {
	if let stored = session.configuration.httpCookieStorage?.cookies(for: url), !stored.isEmpty {
		return stored   // every cookie in the jar scoped to this URL
	}
	guard let headers = response.allHeaderFields as? [String: String] else { return [] }
	return HTTPCookie.cookies(withResponseHeaderFields: headers, for: url)   // only THIS response's cookies
}
```

The store path returns **every** cookie in the session's jar scoped to `url`, which is a superset of "every cookie the bootstrap response set" (the doc comment's wording, and exactly what the header fallback returns). For the isolated ephemeral jar this is harmless — and arguably more correct for reader auth, since all cookies are same-host — but the two branches aren't equivalent, and the header fallback is what any test exercising a non-populated store will hit. Consider either narrowing the store path to the response's `Set-Cookie` names for parity, or adjusting the comment to say "every cookie in the session jar for this URL" so the intent is unambiguous. Not a bug; only a documentation/behaviour-parity nuance. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2FShared%2FReadplaceAPI.swift%2C%20reconcile%20sessionCookies%20store-path%20%28returns%20all%20jar%20cookies%20for%20the%20URL%29%20with%20the%20header%20fallback%20%28only%20this%20response%27s%20cookies%29%20and%20the%20doc%20comment%20that%20says%20%22every%20cookie%20the%20response%20set%22&repo=Readplace/readplace.com)

### 2. `check-coverage.py` keys files by basename — latent collision could mask a regression

`measured_files()` keys the coverage map by `f.get("name")` (a basename) and merges duplicates with `max()`:

```python
best[name] = max(best.get(name, 0.0), cov)
```

This is intentional for a `Shared/` file compiled into both the app and extension targets. But it means if two source files with the **same basename** in different directories ever coexist, the gate silently takes the higher coverage of the two, which could hide a low-coverage file behind a well-covered namesake. No duplicate basenames exist today (verified), so this is latent, not active — but this gate is meant to be load-bearing, so keying by relative path (or asserting basename uniqueness among non-`Shared` files) would harden it against a future footgun. [Fix this →](https://claude.ai/code?q=In%20projects%2Fios-readplace%2Fscripts%2Fcheck-coverage.py%2C%20harden%20measured_files%20against%20same-basename%20source%20files%20in%20different%20directories%20being%20merged%20via%20max%28%29%20-%20key%20by%20path%20or%20assert%20basename%20uniqueness&repo=Readplace/readplace.com)

PR #931, opened 2026-07-05
fix(hutch): keep the changelog NEW chip on the homepage A/B-split arms

FagnerMartinsBrack:
## Problem

The changelog banner's **NEW** chip is a one-shot novelty signal. An inline, pre-paint script (`CHANGELOG_SEEN_SCRIPT`) records a banner version as "seen" in `localStorage['readplace.changelog-seen']` the first time any page carrying the banner renders; every later render finds the version already stored, adds `changelog-banner--seen`, and hides the chip.

The homepage A/B split (`homepage-split.client.ts`) is a **client-side `location.replace` redirect**, not a server 302. So a guest actually loads `/`, whose base shell renders the banner and its inline seen-script near the top of `<body>`. That script runs during parse — before the deferred split script — and writes the version to `localStorage`. Only *then* does the split script redirect to `/landing-a` / `/landing-b`. The arm re-renders the identical shell; the seen-script now finds the version already stored, adds `--seen`, and the chip is hidden. The throwaway `/` frame — which the visitor never actually sees — burns the one-shot NEW before the arm paints, so every redirected human guest loses NEW on a freshly announced feature.

## Fix

Thread an optional `suppressChangelogSeenScript` flag on `BannerState`, set only on the `/` launcher render. The seen-script has exactly one production render path (route → `Base(body, state)` → `renderBaseTemplate` → `renderChangelogBannerShell`), so the flag rides the already-threaded `BannerState` and leaves every other `Base(...)` call site untouched.

The **banner and NEW chip still render** (preserving the announcement for no-JS visitors, per the zero-client-JS principle) — only the inline version-marking script is omitted. The arm the visitor actually lands on then gets the genuine first impression and shows NEW once, then correctly self-suppresses on the next navigation.

### Suppress only when the render is genuinely thrown away

The launcher render is discarded **only when the split script actually replaces `/`** — a human guest **and** `HOMEPAGE_SPLIT.active`. The condition is `abSplit && HOMEPAGE_SPLIT.active`:

- **Kill switch off** (`HOMEPAGE_SPLIT.active = false`, the documented steady state once the experiment concludes): the split bundle is still emitted for human guests but early-returns without redirecting, so the guest stays on a real, persistent `/`. It must keep its seen-script — otherwise NEW would flash on every `/` visit and never self-suppress (the same regression, inverted).
- **Bots** (`abSplit` false): keep the control `/` and its seen-script; they never run the script anyway.
- **Landing arms** (`/landing-a`, `/landing-b`): render `HomePage` with `banner` as-is, so they keep the seen-script and record the genuine first impression on the reader's real page.

## Changes

- `src/packages/web-shell/src/changelog-banner.ts` — `renderChangelogBannerShell` takes an optional `{ suppressSeenScript }`; the inline seen-script is gated behind `{{#if seenScript}}` (nested inside `{{#if visible}}`). Chip, hook, link, and dismiss form stay unconditional.
- `src/packages/web-shell/src/banner-state.ts` — new optional `BannerState.suppressChangelogSeenScript` field, documented.
- `src/packages/web-shell/src/base.component.ts` — passes `{ suppressSeenScript: state.suppressChangelogSeenScript }` to the shell.
- `projects/hutch/src/runtime/server.ts` — the `/` handler sets `suppressChangelogSeenScript = abSplit && HOMEPAGE_SPLIT.active`.

## Tests

- `changelog-banner.test.ts` — `renderChangelogBannerShell(BANNER, "/x", { suppressSeenScript: true })` yields a visible banner with the NEW chip and `data-changelog-version` but **no** `<script>`.
- `base.component.test.ts` — a page with `changelogBanner` set and `suppressChangelogSeenScript: true` renders the visible banner with no seen-script.
- `changelog-banner.route.test.ts` — `GET /` (default human UA) suppresses the seen-script; `GET /landing-a` and `/landing-b` keep it; `GET /` with a Googlebot UA (control) keeps it.

## Verification

`pnpm check` passes across all 37 projects, including the 100% coverage gate on `@packages/web-shell` and `hutch`.

## Known limitation

If the split bundle fails to load (adblocker / CSP / a 404 during a deploy race), a human guest stays on `/` with the seen-script suppressed and NEW re-flashes — a cosmetic edge inherent to the existing client-redirect A/B architecture, not introduced here. Gating on `HOMEPAGE_SPLIT.active` handles the deliberate, documented kill-switch case, which is the real steady state.

https://claude.ai/code/session_012SenX5p52r8hyECDxVR7pt

---
_Generated by [Claude Code](https://claude.ai/code/session_012SenX5p52r8hyECDxVR7pt)_

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -180,6 +180,7 @@ import { E2EFixturePage } from "./web/pages/e2e-fixture";
 import { createE2EFixturePdf } from "./web/pages/e2e-fixture-pdf";
 import { initInstallRoutes } from "./web/pages/install";
 import { initLandingRoutes } from "./web/pages/landing";
+import { HOMEPAGE_SPLIT } from "./web/experiments/homepage-split";
 import { detectInstallBrowser } from "./web/onboarding/extension-install";
 import { NotFoundPage } from "./web/pages/not-found";
 import { initGetEffectiveAccess } from "./domain/access/effective-access";
@@ -687,10 +688,22 @@ export function createApp(dependencies: AppDependencies): Express {
 		// Gate the client A/B split redirect on humans: bots keep the canonical `/`
 		// (control) instead of following a client redirect into a noindex arm.
 		const abSplit = !isbot(req.get("user-agent"));
+		// Suppress this render's changelog seen-script only when the split script
+		// will actually replace `/` before it paints — a human guest AND the
+		// experiment active. That throwaway frame must not record the banner
+		// version as seen, or the landing arm the reader lands on would hide the
+		// one-shot NEW chip. When the split is off (kill switch) the emitted
+		// script no-ops and the guest stays on `/`, and a bot never runs it at
+		// all: in both cases `/` is the reader's real, persistent page, so it
+		// keeps its seen-script and the chip self-suppresses on the next visit.
+		const suppressChangelogSeenScript = abSplit && HOMEPAGE_SPLIT.active;
 		sendComponent(
 			req,
 			res,
-			Base(HomePage({ userCount, staticBaseUrl, browser, foundingAllocation, abSplit }), banner),
+			Base(HomePage({ userCount, staticBaseUrl, browser, foundingAllocation, abSplit }), {
+				...banner,
+				suppressChangelogSeenScript,
+			}),
 		);
 	});
 
--- a/projects/hutch/src/runtime/web/changelog-banner.route.test.ts
+++ b/projects/hutch/src/runtime/web/changelog-banner.route.test.ts
@@ -48,4 +48,61 @@ describe("changelog banner on hutch pages", () => {
 		);
 		expect(banner?.classList.contains("changelog-banner--hidden")).toBe(true);
 	});
+
+	// The A/B-split launcher render of `/` is replaced client-side before it paints;
+	// keeping its seen-script would record the version as seen and hide the NEW chip
+	// on the arm the reader actually lands on. These three cases pin the boundary:
+	// the launcher suppresses the script, the arms and the bot control keep it.
+	it("suppresses the seen-script on the / launcher render for a human guest, keeping the banner visible", async () => {
+		const { app } = createTestApp(createDefaultTestAppFixture(TEST_APP_ORIGIN), {
+			getChangelogBanner: async () => BANNER,
+		});
+
+		const response = await request(app).get("/");
+
+		expect(response.status).toBe(200);
+		const banner = new JSDOM(response.text).window.document.querySelector(
+			"[data-test-changelog-banner]",
+		);
+		assert(banner, "changelog banner must render on the / launcher");
+		expect(banner.classList.contains("changelog-banner--visible")).toBe(true);
+		expect(banner.getAttribute("data-changelog-version")).toBe(VERSION);
+		expect(banner.querySelector("script")).toBeNull();
+	});
+
+	it("keeps the seen-script on the landing arms so the reader's real page records the first impression", async () => {
+		const { app } = createTestApp(createDefaultTestAppFixture(TEST_APP_ORIGIN), {
+			getChangelogBanner: async () => BANNER,
+		});
+
+		for (const path of ["/landing-a", "/landing-b"]) {
+			const response = await request(app).get(path);
+
+			expect(response.status).toBe(200);
+			const banner = new JSDOM(response.text).window.document.querySelector(
+				"[data-test-changelog-banner]",
+			);
+			assert(banner, `changelog banner must render on ${path}`);
+			expect(banner.classList.contains("changelog-banner--visible")).toBe(true);
+			expect(banner.querySelector("script")?.textContent).toBe(CHANGELOG_SEEN_SCRIPT);
+		}
+	});
+
+	it("keeps the seen-script on / for a crawler, which stays on the control (no client redirect fires)", async () => {
+		const { app } = createTestApp(createDefaultTestAppFixture(TEST_APP_ORIGIN), {
+			getChangelogBanner: async () => BANNER,
+		});
+
+		const response = await request(app)
+			.get("/")
+			.set("User-Agent", "Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)");
+
+		expect(response.status).toBe(200);
+		const banner = new JSDOM(response.text).window.document.querySelector(
+			"[data-test-changelog-banner]",
+		);
+		assert(banner, "changelog banner must render on the control /");
+		expect(banner.classList.contains("changelog-banner--visible")).toBe(true);
+		expect(banner.querySelector("script")?.textContent).toBe(CHANGELOG_SEEN_SCRIPT);
+	});
 });
--- a/projects/hutch/src/runtime/web/changelog-banner.route.test.ts
+++ b/projects/hutch/src/runtime/web/changelog-banner.route.test.ts
@@ -8,6 +8,7 @@ import { TEST_APP_ORIGIN, createDefaultTestAppFixture } from "@packages/test-fix
 import { JSDOM } from "jsdom";
 import request from "supertest";
 import { createTestApp } from "../test-app";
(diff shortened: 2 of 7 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #931
- **Commit**: 6bc7b8037e7d1b653fd13efadb09d61769b85327

FagnerMartinsBrack:
### Review: keep the changelog NEW chip on the homepage A/B-split arms

I traced the full render path (`/` route → `Base(body, state)` → `renderChangelogBannerShell`), the client redirect timing (`homepage-split.client.ts`), and every consumer of the seen-script. **The fix is correct and the reasoning holds across all four scenarios** (first-time guest, bundle-fail, kill-switch-off, bot). Verified independently:

- **Single render path confirmed** — `renderChangelogBannerShell` has exactly one non-test caller (`base.component.ts:249`); the `Boolean(banner) && !options?.suppressSeenScript` gate nested inside `{{#if visible}}` is sound.
- **No collateral breakage** — grep shows only 5 files reference `CHANGELOG_SEEN_SCRIPT`; no existing `/`-hitting test asserts the seen-script's presence, so nothing regresses. Existing `base.component`/`changelog-banner` tests pass unchanged because `suppressChangelogSeenScript` defaults to `undefined` → script still rendered.
- **Kill-switch asymmetry is real and correctly exploited** — `home.component.ts:264` emits the split bundle on `abSplit` alone, while `initHomepageSplit` (`homepage-split.client.ts:63`) early-returns on `!active`. So `abSplit && HOMEPAGE_SPLIT.active` is the precise server-side predicate for "this frame will actually be replaced."

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The launcher test is coupled to `HOMEPAGE_SPLIT.active === true`; the kill-switch branch is unverified

`suppressChangelogSeenScript = abSplit && HOMEPAGE_SPLIT.active` (`server.ts:699`). The new test *"suppresses the seen-script on the / launcher render for a human guest"* (`changelog-banner.route.test.ts:56`) asserts `banner.querySelector("script")` is `null` — which holds **only because `HOMEPAGE_SPLIT.active === true`** (`homepage-split.ts:36`).

Two consequences worth being aware of:

- **The test breaks when the kill switch is flipped.** The experiment config's own docstring advertises `active: false` as a one-line no-op, and this PR's description calls it "the documented steady state once the experiment concludes." When that flip happens, `/` keeps its seen-script and this assertion fails — so a change the architecture is designed to make trivially will require also editing this test.
- **The `active === false` branch has no test.** That branch is the entire justification for the `&& HOMEPAGE_SPLIT.active` conjunction (documented at `server.ts:691-698`), yet nothing exercises it. Branch coverage still reports 100% only because `HOMEPAGE_SPLIT.active` is a compile-time-constant `true`, so V8 sees no branch to cover — the gap is behavioral, not a coverage-gate miss.

Note the sibling split-script tests (`home.route.test.ts:67` and `:77`) key only off `abSplit`, so they survive a kill-switch flip; only this new seen-script test couples to `active`.

Because this is experiment-scoped code that will be torn down as a unit, a DI refactor to inject the split config is likely overkill. A one-line comment on the test recording the `active === true` assumption is a proportionate fix.

### 2. Returning visitors can briefly flash NEW on the throwaway `/` frame

Previously the inline seen-script on `/` ran pre-paint and hid NEW even on the launcher frame for a returning visitor. With the script suppressed, a returning visitor's `/` frame can paint NEW for the moment before the deferred split script fires `location.replace`. They still land on the arm where the seen-script correctly hides NEW, so their *real* page is right and the flash is confined to the discarded frame — the same class as the "Known limitation" already in the PR body, which only calls out the bundle-fail variant. No action needed; noting it for completeness.

FagnerMartinsBrack:
@claude address this low prio issue: 1. The launcher test is coupled to HOMEPAGE_SPLIT.active === true; the kill-switch branch is unverified
suppressChangelogSeenScript = abSplit && HOMEPAGE_SPLIT.active (server.ts:699). The new test "suppresses the seen-script on the / launcher render for a human guest" (changelog-banner.route.test.ts:56) asserts banner.querySelector("script") is null — which holds only because HOMEPAGE_SPLIT.active === true (homepage-split.ts:36).

Two consequences worth being aware of:

The test breaks when the kill switch is flipped. The experiment config's own docstring advertises active: false as a one-line no-op, and this PR's description calls it "the documented steady state once the experiment concludes." When that flip happens, / keeps its seen-script and this assertion fails — so a change the architecture is designed to make trivially will require also editing this test.
The active === false branch has no test. That branch is the entire justification for the && HOMEPAGE_SPLIT.active conjunction (documented at server.ts:691-698), yet nothing exercises it. Branch coverage still reports 100% only because HOMEPAGE_SPLIT.active is a compile-time-constant true, so V8 sees no branch to cover — the gap is behavioral, not a coverage-gate miss.
Note the sibling split-script tests (home.route.test.ts:67 and :77) key only off abSplit, so they survive a kill-switch flip; only this new seen-script test couples to active.

Because this is experiment-scoped code that will be torn down as a unit, a DI refactor to inject the split config is likely overkill. A one-line comment on the test recording the active === true assumption is a proportionate fix.

FagnerMartinsBrack:

## Context
- **PR**: #931
- **Commit**: 5017806cff990bfdd11bd151ba2823d5b8978a78

FagnerMartinsBrack:
## Review: keep the changelog NEW chip — re-review of `5017806c`

This commit addresses the prior review's low-priority #1 (launcher test coupled to `HOMEPAGE_SPLIT.active`). I re-traced the change end to end. The resolution is correct, idiomatic, and proportionate — no new issues.

Verified independently:
- **Predicate is exact** — server gate `abSplit && HOMEPAGE_SPLIT.active` (`server.ts:699`) mirrors the client's guards: `initHomepageSplit` early-returns on `!config.active` then `pathname !== "/"` (`homepage-split.client.ts:63-64`). The server suppresses the seen-script for precisely the frames the client will `location.replace`.
- **`active` is a compile-time `true`** (`homepage-split.ts:36`) — the new `assert` adds no coverage branch, and the launcher test's `toBeNull()` holds only under that value, which is exactly what the assert pins.
- **Guard is correctly scoped** — only the launcher test (`changelog-banner.route.test.ts:71`) couples to `active`; the arm and crawler cases assert the script is present regardless, so they needed no guard and got none.
- **`node:assert/strict`** is the right import for test code (per CLAUDE.md); the message spells out the kill-switch consequence.
- **No collateral risk** — `renderChangelogBannerShell` still has one non-test caller (`base.component.ts`); the optional 3rd param leaves every other call site untouched.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The `assert` guard resolves the coupling but does not add behavioral coverage for `active === false` (acknowledged, by design)

The `assert(HOMEPAGE_SPLIT.active, …)` (`changelog-banner.route.test.ts:71-74`) is a good answer to the prior note — arguably better than the suggested prose comment, because it converts the implicit "this expectation only holds while active" assumption into an executable tripwire: flip the kill switch and this test fails at the assertion with guidance ("`/` stays put and keeps its seen-script") instead of an opaque `toBeNull()` mismatch. It also encodes the coupling as a real `import` rather than a symbol name in prose that can silently rot.

Two things for the record (neither blocking):
- It documents/guards the coupling; it does not exercise the `active === false` server behavior. That branch remains verified by reasoning + the `initHomepageSplit` unit tests, not by this route suite — reaching it here would need the split-config DI seam the prior review rightly called overkill for teardown-scoped code. The PR is transparent about this.
- Because `HOMEPAGE_SPLIT.active` is a constant `true`, the assert introduces no uncovered branch (its own branch lives inside `node:assert`), so the 100% gate on `@packages/web-shell` and the hutch branch threshold are unaffected — consistent with the reported `pnpm check` pass on this commit.

### 2. Returning-visitor NEW flash on the throwaway `/` frame — unchanged and inherent

Noted in the prior review and the PR's "Known limitation": a returning visitor's discarded `/` frame can paint NEW for the instant before `location.replace` fires. Their real page (the arm) still self-suppresses correctly. Not introduced by `5017806c`; listed only for completeness.

PR #932, opened 2026-07-05
feat(hutch): paginate the inbox email list and rework Manage addresses

FagnerMartinsBrack:
Four UX changes to the email inbox feature (behind the `?feature=email` flag). Both pages are HTML-only — no Siren/extension/iOS/MCP client is affected.

## Inbox email list — pagination (`GET /inbox`)

The list previously fetched and rendered **every** email with no pagination. It's now paginated the same way `/queue` is:

- 1-based `?page=` param, 20 emails per page, newest-first across the page boundary
- htmx-boosted **Previous / Next** nav with "Page X of Y"
- a **302 clamp** that redirects an out-of-range page (stale bookmark / hand-typed URL) to the last valid page
- the per-email link-count fan-out is now naturally bounded to the page size instead of the whole mailbox

**Incidental fix:** the DynamoDB adapter's old single, un-looped `Query` silently **truncated any email partition larger than 1 MB**. It now walks `ExclusiveStartKey` to exhaustion (via the existing `forEachQueryPage` helper), so both `total` and the page slice are computed over the full partition.

The store contract `listEmailsByUserId` changes from `(userId) => InboxEmailEntry[]` to `({ userId, page, pageSize }) => { emails, total, page, pageSize }`. It's an internal, unversioned contract; the only production caller is the `GET /inbox` route (the in-memory fixture, the DynamoDB adapter, and test call sites all move to the new shape in the same change).

## Manage addresses — layout rework (`GET /inbox/addresses`)

- **Section order** (top → bottom): the create-address form **and its "How it works" explainer** move to the top, then active addresses, then disabled.
- **Grouping:** active addresses render first; **all disabled addresses collapse into a native `<details>` group** as the last item, closed by default. The group renders unconditionally with a `--visible` / `--hidden` state class (no `open` attribute) rather than being conditionally rendered.
- **Row layout:** the **Copy** control moves to the **left** of the address field; the **Active / Disable** controls sit at the **right edge** of the row.

A new `toInboxAddressesViewModel` partitions entries with the canonical `isLiveAddress` predicate, so list membership encodes active/disabled state instead of a per-row `enabled` flag.

## Verification

- `pnpm nx run hutch:check` and the repo-wide pre-commit `pnpm check` pass (lint, types, unit + integration tests, 100% coverage, knip, unused-CSS).
- Driven end-to-end in the running app with the flag on: 21 seeded emails → page 1 shows 20 rows + "Page 1 of 2" + Next; page 2 shows the single oldest email + Previous; `?page=99` → 302 to the last page. On the addresses page: create form + "How it works" at the top, active list with Copy-left / Active-Disable-right, and disabling an address moves it into the collapsed "Disabled addresses (N)" group (which expands natively).

## Notes / out of scope

- No re-enable action for disabled addresses exists today (disable is permanent); this only regroups them.
- Queue's own pagination internals are untouched — the inbox replicates the pattern rather than importing it.

https://claude.ai/code/session_01PfH8LL8NeuuSde6WtcbkuZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01PfH8LL8NeuuSde6WtcbkuZ)_

### DIFF
--- a/projects/hutch/src/runtime/delete-account/delete-account-handler.test.ts
+++ b/projects/hutch/src/runtime/delete-account/delete-account-handler.test.ts
@@ -386,7 +386,11 @@ describe("delete-account handler", () => {
 			savedArticle: false,
 		});
 		assert.equal(await s.subs.findByUserId(victim.userId), undefined);
-		assert.equal((await s.inboxEmail.listEmailsByUserId(victim.userId)).length, 0);
+		assert.equal(
+			(await s.inboxEmail.listEmailsByUserId({ userId: victim.userId, page: 1, pageSize: 20 }))
+				.total,
+			0,
+		);
 		const victimLinks = await s.inboxLink.listLinksByEmail({
 			userId: victim.userId,
 			receivedAtMessageId: victim.ramA,
@@ -441,7 +445,16 @@ describe("delete-account handler", () => {
 		const bystanderSub = await s.subs.findByUserId(bystander.userId);
 		assert(bystanderSub, "expected the bystander subscription to survive");
 		assert.equal(bystanderSub.status, "active");
-		assert.equal((await s.inboxEmail.listEmailsByUserId(bystander.userId)).length, 2);
+		assert.equal(
+			(
+				await s.inboxEmail.listEmailsByUserId({
+					userId: bystander.userId,
+					page: 1,
+					pageSize: 20,
+				})
+			).total,
+			2,
+		);
 		assert.equal(
 			(
 				await s.inboxLink.listLinksByEmail({
@@ -647,7 +660,11 @@ describe("delete-account handler", () => {
 		// were the rows removed.
 		assert.equal(s.rawEmailDeleteArgs.length, 1);
 		assert.deepEqual(sorted(s.rawEmailDeleteArgs[0]), sorted(account.rawKeys));
-		assert.equal((await s.inboxEmail.listEmailsByUserId(account.userId)).length, 0);
+		assert.equal(
+			(await s.inboxEmail.listEmailsByUserId({ userId: account.userId, page: 1, pageSize: 20 }))
+				.total,
+			0,
+		);
 
 		// The per-user schedule deletes run before the injected S3 failure, so the
 		// pre-expiry trial-reminder schedule delete fired on both the failed first
--- a/projects/hutch/src/runtime/domain/inbox/receive-email-handler.test.ts
+++ b/projects/hutch/src/runtime/domain/inbox/receive-email-handler.test.ts
@@ -2,10 +2,11 @@ import assert from "node:assert/strict";
 import { HutchLogger, noopLogger } from "@packages/hutch-logger";
 import {
 	DEFAULT_INBOX_ALIAS,
+	type InboxEmailStore,
 	MessageIdSchema,
 	type ParseEmailResult,
 } from "@packages/domain/inbox";
-import { UserIdSchema } from "@packages/domain/user";
+import { type UserId, UserIdSchema } from "@packages/domain/user";
 import { buildLambdaContext } from "@packages/test-fixtures/lambda-context";
 import { initInMemoryInboxAddress } from "@packages/test-fixtures/providers/inbox-address";
 import { initInMemoryInboxEmail } from "@packages/test-fixtures/providers/inbox-email";
@@ -79,6 +80,11 @@ function makeHarness(opts?: {
 	return { addressStore, emailStore, rawMap, published, handler, run, runMany };
 }
 
+async function listEmails(emailStore: InboxEmailStore, userId: UserId) {
+	const { emails } = await emailStore.listEmailsByUserId({ userId, page: 1, pageSize: 100 });
+	return emails;
+}
+
 async function mintAddress(addressStore: ReturnType<typeof initInMemoryInboxAddress>) {
 	const entry = await addressStore.createAddress({
 		userId: OWNER,
@@ -100,7 +106,7 @@ describe("initReceiveEmailHandler", () => {
 
 		assert(result);
 		expect(result.batchItemFailures).toHaveLength(1);
-		expect(await emailStore.listEmailsByUserId(OWNER)).toHaveLength(0);
+		expect(await listEmails(emailStore, OWNER)).toHaveLength(0);
 		expect(published).toHaveLength(0);
 	});
 
@@ -112,7 +118,7 @@ describe("initReceiveEmailHandler", () => {
 
 		assert(result);
 		expect(result.batchItemFailures).toHaveLength(1);
-		expect(await emailStore.listEmailsByUserId(OWNER)).toHaveLength(0);
+		expect(await listEmails(emailStore, OWNER)).toHaveLength(0);
 		expect(published).toHaveLength(0);
 	});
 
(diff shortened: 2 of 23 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #932
- **Commit**: dfdc3df8fa3adda538692463b810a2733e1a6f53

FagnerMartinsBrack:

None found

## Medium Priority Issues

### 1. Store reads the entire email partition into memory on every page view
`listEmailsByUserId` in `projects/hutch/src/runtime/providers/inbox-email/dynamodb-inbox-email.ts` now walks the whole partition with `forEachQueryPage`, pushing every full email row into `all[]`, then slices out 20 and derives `total` from `all.length` — on every `/inbox` load, including page 1.

The walk-to-exhaustion correctly fixes the pre-existing 1 MB single-`Query` truncation, but trades that bug for an unbounded read. The PR says the inbox "replicates the queue pattern," yet at the store layer it does the opposite of queue. Queue's `findArticlesByUser` (`projects/hutch/src/runtime/providers/article-store/dynamodb-article-store.ts`, ≈ lines 289–338) keeps the fetch bounded: a `Select: "COUNT"` loop for `total` (O(1) memory/network) plus a `Limit: pageSize` query that skips `(page-1)*pageSize` and breaks early. The inbox version is O(mailbox size) in memory and wire bytes per request, so the pagination bounds only the link-count fan-out and the render — not the primary data fetch. This matters because inbox emails are kept forever with no TTL and no per-user cap (`projects/hutch/src/infra/hutch-storage.ts:241`), so the partition grows without bound and every visit pulls the whole thing into the Lambda.

Not a correctness bug, and it's behind `?feature=email` while the feature is early, so it can be a conscious deferral — but the reference implementation is already in-repo and this is cheapest to align now.

## Low Priority Suggestions

### 1. Duplicated page/pageSize precondition asserts across both store implementations
The four `assert(Number.isInteger(page))` / `>= 1` guards are copied verbatim into both `dynamodb-inbox-email.ts` and `in-memory-inbox-email.ts`. A small shared guard in `@packages/domain/inbox` would keep the two implementations from drifting. Optional — each adapter honouring its own contract is also defensible.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #932
- **Branch**: claude/inbox-pagination-addresses-layout-hz1t2w

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #932
- **Commit**: 40cfb59ad677014ca80e22bf96247125e89218c1

FagnerMartinsBrack:
The prior round's one substantive concern is resolved and the change is clean. I re-reviewed the whole PR at `40cfb59a`, tracing the reworked store branch-by-branch against its tests and cross-checking the pagination + addresses rework against the queue reference.

None found

## Medium Priority Issues

None found

### ✅ Previously-flagged medium (store read the whole partition on every page) — resolved
`listEmailsByUserId` in [`dynamodb-inbox-email.ts`](projects/hutch/src/runtime/providers/inbox-email/dynamodb-inbox-email.ts) now mirrors queue's `findArticlesByUser`: a `Select: "COUNT"` walk for `total` (O(1) memory/network) plus a `Limit: pageSize` walk that skips `(page-1)*pageSize` and stops once the page is full. Page 1 fetches ~20 rows instead of the entire mailbox, and the COUNT walk still iterates `LastEvaluatedKey` to exhaustion so the **1 MB-truncation fix is preserved** (`total` is computed over the full partition without buffering every row). I hand-traced the two subtle cases and both are correct:
- **early-exit** (`page:2, pageSize:1`): stops after the page fills, issuing no query past the second item key — matches the `stops after filling the page…` test.
- **overflow-drop** (`page:2, pageSize:2`, a 1 MB split mid-page): the `else if (emails.length < pageSize)` correctly discards the trailing row once the page is full — matches the `drops overflow rows…` test.

The one cosmetic delta from queue — the omitted `if (userArts.length >= pageSize && !exclusiveStartKey) break;` — is a no-op here because the `while` guard already covers it, so the inbox loop is equivalent (and slightly cleaner).

## Low Priority Suggestions

### 1. Duplicated page/pageSize precondition asserts across both stores
The four `assert(Number.isInteger(page))` / `>= 1` guards are still copied verbatim into [`dynamodb-inbox-email.ts`](projects/hutch/src/runtime/providers/inbox-email/dynamodb-inbox-email.ts) and [`in-memory-inbox-email.ts`](src/packages/test-fixtures/src/providers/inbox-email/in-memory-inbox-email.ts). Carried over from the previous round and consciously deferred (auto-apply is High/Medium-only) — noting it only for completeness. A shared guard in `@packages/domain/inbox` would prevent drift; each adapter honouring its own contract is also defensible.

### 2. `listEmailsByUserId` JSDoc still claims strong consistency for the production adapter
The contract doc in [`inbox-email.types.ts`](src/packages/domain/src/inbox/inbox-email.types.ts) says *"Strongly consistent in the in-memory fixture; the production adapter reads the base table so it is too."* Reading the **base table** is necessary but not sufficient — a DynamoDB `Query` is eventually consistent unless `ConsistentRead: true` is set, and the adapter doesn't set it. The behaviour is fine (eventual consistency is perfectly acceptable for an inbox list), so this is purely a doc-accuracy nit; the wording predates this PR but was re-touched by the contract change, so worth a one-line correction while it's fresh.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 29006315992
- **PR**: #932
- **Branch**: claude/inbox-pagination-addresses-layout-hz1t2w

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #932
- **Commit**: dae84a0f377f5d449bab6171024e450324b2a088

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

### ✅ Store bounded-fetch (prior medium) — still correct after the rebase
`listEmailsByUserId` in `projects/hutch/src/runtime/providers/inbox-email/dynamodb-inbox-email.ts` keeps the `Select: "COUNT"` walk for `total` (O(1) memory/network) plus a `Limit: pageSize` walk that skips `(page-1)*pageSize` and stops once the page fills. I re-traced early-exit (`page:2,pageSize:1`) and overflow-drop (`page:2,pageSize:2`, 1 MB mid-page split) against their tests — both correct. The CI fix (`dae84a0f`) correctly restored the `forEachQueryPage` import, which is still used by the two whole-partition deletion methods (`listDeletionReferencesByUserId`, `deleteAllEmailsByUserId`) that must read every row, without touching the bounded read.

## Low Priority Suggestions

### 1. Duplicated page/pageSize precondition asserts across both stores
The four `assert(Number.isInteger(page))` / `>= 1` guards are still copied verbatim into `projects/hutch/src/runtime/providers/inbox-email/dynamodb-inbox-email.ts` and `src/packages/test-fixtures/src/providers/inbox-email/in-memory-inbox-email.ts`. Carried over and consciously deferred across prior rounds. A shared guard in `@packages/domain/inbox` would prevent drift; each adapter honouring its own contract is also defensible.

### 2. `listEmailsByUserId` JSDoc still claims strong consistency for the production adapter
`src/packages/domain/src/inbox/inbox-email.types.ts` says *"Strongly consistent in the in-memory fixture; the production adapter reads the base table so it is too."* A base-table DynamoDB `Query` without `ConsistentRead: true` is eventually consistent, and the adapter doesn't set it. The behaviour is fine (eventual consistency is acceptable for an inbox list) — purely a doc-accuracy nit, carried over from the last round.

PR #933, opened 2026-07-05
feat(auth): release Sign in with Apple to all users

FagnerMartinsBrack:
## Summary

Remove the `?feature=apple` querystring gate so the Sign in / Sign up with Apple button is now rendered for every visitor on `/login` and `/signup`.

- Drops the `{{#if appleEnabled}}` block in `login.template.html` and `signup.template.html`.
- Removes the `appleEnabled` prop from `LoginPage` / `SignupPage` and the `featureToggle` dependency from `initAuthRoutes`.
- Updates the auth route tests to assert the Apple button is always present (no more `?feature=apple`-gated cases).

## Test plan

- [x] `pnpm nx run hutch:check` (unit + integration tests + coverage)

---
_Generated by [Claude Code](https://claude.ai/code/session_01KvHGZmUtKDbkCGgVH2XhsR)_

PR #939, opened 2026-07-05
feat(conversion): pre-expiry trial reminder email (R1)

FagnerMartinsBrack:
# feat(conversion): pre-expiry trial reminder email (R1)

## Hypothesis
Card-free trials expire silently: users get zero contact between signup and the moment their account flips read-only, so the trial ends by default rather than by decision. A founder-voice email 2 days before `trialEndsAt`, with a direct subscribe CTA, converts some of that silent expiry into deliberate subscribe decisions (and honest churn signal on replies).

## Evidence (30d, Jun 4 – Jul 4 2026)
- **28 of 28** trials that reached end-of-trial emitted `charge_failed reason=no_card_on_file` → auto-cancelled ~60 min later. **Trial→paid = 0%.**
- The **only** lifecycle email today fires **3 days AFTER cancellation** (feedback email: 28 scheduled, 22 sent). Nothing fires before expiry.
- The upgrade surface works but nobody is pushed to it in time: header trial-countdown 54 clicks/40 visitors → /account 22 views → subscribe button 4 clicks → 2 checkout successes (one of them *after* the user had already been cancelled — proof that a post-hoc nudge can convert, just far too late).
- 6 of 17 sampled churned trialists had 0 saves — the reminder copy therefore never fabricates usage (saved-N-articles clause is omitted at zero).

## What this changes
- `SendTrialFeedbackEmailCommand` gains optional `kind: "feedback" | "reminder"` (absent = feedback → fully backward compatible with in-flight schedules).
- New deterministic EventBridge Scheduler one-shot `trial-reminder-<userId>` (same target DetailType, `Input {userId, kind:"reminder"}`, `ActionAfterCompletion: DELETE`) created at all **three** trial-signup sites (email, Google, Apple) at `trialEndsAt − 2d`, and recreated on trial reactivation when >2d remain.
- Deleted alongside every existing `deleteTrialEndSchedule` call: checkout success and the cancel-subscription `trialing` branch.
- The existing `send-trial-feedback-email` Lambda branches on `kind`. Reminder path re-checks `status === "trialing"` && `trialEndsAt` in the future, and stamps a new `trialReminderEmailSentAt` row flag (`markTrialReminderEmailSent`) so at-least-once delivery can't double-send.
- New founder-voice template (`trial-reminder-email`), CTA → `/account?utm_source=trial-reminder&utm_medium=email&utm_campaign=trial-preexpiry`.
[...]

PR #940, opened 2026-07-05
feat(hutch): self-serve account deletion (App Store 5.1.1(v))

FagnerMartinsBrack:
## Why

The iOS app supports in-app account creation, so App Store Guideline **5.1.1(v)** requires users to be able to **initiate account deletion from within the app**. Neither the app nor readplace.com offered self-serve deletion — the privacy policy told users to email, which Apple explicitly rejects. This adds end-to-end self-serve deletion.

## What

**Server pipeline** (mirrors the existing export / cancel-subscription workers):
- `DeleteAccountCommand` + publisher + EventBridge/in-memory wrappers.
- A **DLQ-backed, idempotent `delete-account` worker** that scrubs every user-owned store: auth row + Gmail uniqueness claim, sessions, OAuth grants (+ paired refresh rows), saved articles, digest queue, reader-ready state, iOS onboarding, inbox emails/links (+ their RAW_EMAIL and CONTENT S3 objects) and forwarding addresses, S3 exports, password-reset tokens, and a live Stripe subscription + customer. Founding-member (no row) and trialing (no `subscriptionId`) accounts are first-class branches with no Stripe call. Per-user EventBridge schedules are deleted so none fires at a dead account.
- **Synchronous `POST /account/delete`** teardown so the account is unusable the instant the user confirms: destroy every session → revoke every OAuth token → clear the session cookie → publish the command → full-page redirect to the logged-out home.
- Per-store bulk delete/revoke primitives beside each store, mirrored in the in-memory fixtures, unit-tested against the fake-DynamoDB-client pattern.

**Clients:**
- **Siren:** a `class` field on `SirenAction`, a new `account` resource with a bare destructive `delete-account` action, and an `account` discovery link on the collection.
- **Web:** an `/account` "Danger zone" behind an `hx-confirm`, plus a privacy-policy rewrite pointing at the self-serve path.
- **iOS:** a `delete-account` presentation case, the destructive-confirm generalized from row controls to **any** destructive collection affordance (keyed on `isDestructive`, never the action name), and `forceLogout` after a successful account-ending invoke.

**Also:**
- **Real Apple-grant revocation** — Sign in with Apple was released to all users (#933) while this PR was in review, so the originally planned no-op seam shipped its Phase 6 in-PR: the code exchange now requires and persists Apple's `refresh_token` on the users row (stored at signup and refreshed on every returning Apple login, which backfills accounts minted before persistence), and the delete worker revokes it via `POST https://appleid.apple.com/auth/revoke` with a freshly minted ES256 client secret — throwing on non-2xx so the SQS record redrives (queue `maxReceiveCount` raised to 12 to ride out an Apple outage). Password/Google accounts have no stored token and skip the call.
- A **webhook-race fix**: a late `SubscriptionCancelled` event on a just-deleted account is now an idempotent no-op instead of poisoning the queue into the DLQ.

## Product decisions (plan defaults — please confirm)

| # | Decision | Applied |
|---|----------|---------|
| D1 | Inbox forwarding addresses | **Tombstone** — PII stripped, hash kept reserved so it can never be re-minted and leak another user's mail |
| D2 | Stripe customer | **Deleted** (cancels + detaches all cards); Stripe still retains invoices as an immutable tax ledger |
| D3 | Raw `.eml` audit trail | **Deleted** on closure |
| D4 | Password-reset tokens | **Scan-deleted by email** (that table has no TTL, unlike its siblings) |
| D5 | Confirmation | **Native / `hx-confirm`** destructive confirm; no forced re-auth for v1 |
| D6 | PITR + deletion-protection retention | Documented as out-of-band governance |

## Testing

- Provider primitives: unit tests (fake `DynamoDBDocumentClient`) + mirrored in-memory fixtures.
- Worker: 7-case handler test — full scrub, active/trialing/founding billing branches, idempotency (run twice), `batchItemFailures` on a store error, IdP-revocation seam.
- Route + Siren + danger zone: `*.route.test.ts` (supertest + parseHTML).
- iOS: presentation-case + `endsSession` tests. **Note: Swift can't be compiled in this Linux CI environment — the iOS changes need an iOS-side (`make test`) run.**

https://claude.ai/code/session_01RkzGzig54ovfhxFHM9FyG3

---
_Generated by [Claude Code](https://claude.ai/code/session_01RkzGzig54ovfhxFHM9FyG3)_


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 28746111876
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: 5ba5cf724a4b3c7bea2ebd3039612dc5dacb54ba

FagnerMartinsBrack:
### Code Review — self-serve account deletion (App Store 5.1.1(v))

Reviewed the server pipeline, the synchronous teardown, the per-store primitives, the infra wiring, and the Siren/iOS/web clients against the repo guidelines. This is a large, carefully-built change: the worker's idempotency model is mostly sound, the webhook-race fix is precise, IAM is tightly scoped, and the test suite is genuinely thorough. Two idempotency gaps in the async scrub are worth addressing because both can leave a deletion **stuck or incomplete**, which is exactly what this compliance feature must not do.

None found.

## Medium Priority Issues

### 1. Inbox `.eml` / body S3 objects are orphaned (PII not erased) on a partial or redriven scrub

In `delete-account-handler.ts` `processCommand`, the inbox step deletes the DynamoDB email **rows first** and only then the S3 objects:

```ts
const { receivedAtMessageIds, rawEmailS3Keys, bodyS3Keys } = await deps.deleteAllInboxEmails(userId); // deletes rows, returns keys
await deps.deleteAllInboxLinks(userId, receivedAtMessageIds);
await deps.deleteRawEmailObjects(rawEmailS3Keys);   // ← only reachable if the line above didn't throw
await deps.deleteEmailContentObjects(bodyS3Keys);
```

`deleteAllEmailsByUserId` (`dynamodb-inbox-email.ts:68`) reads `rawEmailS3Key`/`bodyS3Key` out of the rows and deletes those rows in the same pass, returning the keys only on full success. The S3 keys (`inbound/<messageId>`, `content/<…>`) have no `userId` prefix, so they are discoverable **only** through those rows. If anything between that call and the two S3 deletes fails (a transient DynamoDB error in `deleteAllInboxLinks`, a Lambda timeout, or the record redriving), the retry finds the rows already gone, returns empty arrays, and the raw `.eml` plus rendered body sit in S3 permanently, unreferenced — violating the handler's stated convergence invariant and retaining user PII. Fix: delete the S3 objects **before** the rows that hold their keys.

### 2. Stripe re-cancel on redrive can poison the queue into the DLQ

The billing block runs `cancelStripeSubscription` → `deleteStripeCustomer` → `deleteSubscription`. If cancel succeeds but `deleteStripeCustomer` then fails, the record redrives; `findSubscriptionByUserId` still returns the row, so `cancelImmediately` fires again on an already-`canceled` but still-existing subscription. `cancelImmediately` treats only HTTP 404 as idempotent success, and Stripe returns 400 for a repeat cancel (`stripe-subscriptions.test.ts:92` confirms 400 throws), so the retry throws forever → DLQ, deletion never completes. The idempotency test doesn't cover this interleaving. Cleanest fix: delete the Stripe customer **before** cancelling (customer deletion cascades cancellation, and every row with a `subscriptionId` has a `customerId`), so the follow-up cancel 404s through the existing idempotent path.

## Low Priority Suggestions

### 1. iOS `forceLogout()` runs even if the delete invoke failed

In `ReadingListView.swift:142` `confirmDestructive`, the `endsSession` branch calls `session.forceLogout()` unconditionally after `invokeCollection(action)`. A failed `POST /account/delete` still logs the user out locally with no error surfaced. Consider gating `forceLogout()` on success. (Swift isn't compiled in this CI.)

### 2. `POST /account/delete` has no error path

The synchronous teardown (`account.page.ts:337`) has no try/catch — a throw 500s the request. The sessions-before-tokens ordering is a deliberate safe-retry choice consistent with `/oauth/revoke`, so this is fine as-is; just flagging the raw 500. CSRF is adequately covered — the `hutch_sid` cookie is `SameSite=Lax`, so cross-site POSTs carry no session and `requireAuth` bounces them.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: cd906eeb697f7e222dc7ce36008b0bad7ac22dcd

FagnerMartinsBrack:
### Code Review — self-serve account deletion (App Store 5.1.1(v)) · re-review at `cd906eeb`

Re-reviewed the two medium issues from the prior pass and swept the fix commit for regressions. **Both mediums are fixed correctly, and the fixes are genuinely sound rather than plausible-but-fragile** — I traced each convergence path and confirmed the external Stripe behavior the fix rests on.

None found

## Medium Priority Issues

None found

**Verification of the two prior mediums (now resolved):**

**M1 — S3 `.eml`/body PII no longer orphaned on redrive.** The email store is split into a read-only `listDeletionReferencesByUserId` (runs first, rows survive) and a `void` `deleteAllEmailsByUserId` (runs last). The scrub is reordered so every S3 object and link row is deleted *before* the email rows that index them (`delete-account-handler.ts:113-119`). Because all S3 objects for all emails are removed before *any* email row is deleted, a mid-scrub failure leaves the rows intact and the redrive re-derives the keys — it converges instead of orphaning PII. This also closes the same latent gap for link rows (no `userId` index). The in-memory fixture mirrors the read-only/void split exactly, and the new redrive test proves the rows survive the first failure and the keys are re-derived on retry.

**M2 — Stripe re-cancel DLQ poison eliminated.** The explicit `cancelStripeSubscription` is removed; the billing block is now just `deleteStripeCustomer` (when `customerId` present) → `deleteSubscription`. Confirmed the three things this rests on:
- **Cascade + idempotency (Stripe docs):** deleting a customer immediately cancels active subscriptions, and a repeat `DELETE` on a deleted customer 404s — which `deleteCustomer` already treats as success, so the redrive converges.
- **Code invariant:** `upsertActive` sets `customerId` and `subscriptionId` together while `upsertTrialing` removes both, so every row carrying a live Stripe sub also carries the `customerId` we delete.
- **Webhook race:** the cascade-fired `customer.subscription.deleted` webhook landing on a just-deleted row is now caught (`ConditionalCheckFailedException` → no-op) in `handle-subscription-cancelled-handler.ts:28-40`.

The redrive test proves `deleteCustomer` runs twice with no re-cancel and the account fully scrubs.

## Low Priority Suggestions

Both carried over from the prior review, both non-blocking.

### 1. iOS `forceLogout()` still runs even if the delete invoke failed
`ReadingListView.swift:142-146` calls `session.forceLogout()` unconditionally after `invokeCollection(action)`. A failed `POST /account/delete` still logs the user out locally with no error surfaced. Gating on a successful invoke would keep a failed attempt in place with the error banner. (Swift isn't compiled in this CI.)

### 2. `POST /account/delete` surfaces a raw 500 on a transient teardown failure
`account.page.ts:337-349` has no try/catch, unlike sibling routes. Express is 5.2.1, so an async throw auto-forwards to `logAndRespondOnError` (a clean 500 — not a hang), and the sessions-before-tokens ordering leaves the account safely retryable. Fine as-is; only flagging the raw 500 vs a friendly retry. CSRF is adequately covered (`SameSite=Lax`).

FagnerMartinsBrack:
@claude — I reviewed this PR **strictly** for App Store **5.1.1(v)** compliance completeness and for scope (that it does *nothing more* than Apple compliance).

**Verdict:** the web deletion path works and satisfies 5.1.1(v) — in-app **Account** link → authenticated in-app webview → `/account` "Danger zone" → real hard-delete (`closeUserAccount` deletes the row, not a soft flag) + full synchronous server teardown, and the privacy policy's email-only-deletion language is removed. **No hard submission blocker today.** But please address the following before this is truly done and scoped.

## 🔴 Compliance robustness

**1. Sign in with Apple revocation is a no-op — latent rejection trap.**
SIWA is *dark-launched* (the Apple button is gated behind `?feature=apple`, and nothing in production ever sets it — `web/feature-toggle.ts`, `auth.page.ts:159/235`; `feature=apple` appears only in tests), so a reviewer never sees SIWA and Apple's SIWA-deletion requirement isn't triggered for this submission. **The trap:** `delete-account/revoke-external-idp-tokens.ts` is a log-only no-op wired at `delete-account.main.ts:119`, and `providers/apple-auth/apple-token.ts:5` parses only `id_token` so Apple's `refresh_token` is discarded at sign-in. The moment SIWA is un-gated (button shown by default, or the iOS app starts sending `feature=apple`) **without** fixing this, in-app deletion silently leaves the app in the user's *Settings → Sign in with Apple* list → rejection under the exact guideline this PR targets.
- **Preferred (small, in-scope):** make it **fail-closed** — add a guard/test asserting "if SIWA is reachable by a normal user ⟹ `revokeExternalIdpTokens` is the real Apple `/auth/revoke` impl **and** `exchangeAppleCode` persists the refresh_token." It turns CI red the day SIWA un-gates without revocation.
- **Or implement now:** persist Apple's `refresh_token` in `apple-token.ts`, POST `https://appleid.apple.com/auth/revoke` (reuse the existing `apple-client-secret.ts` JWT) in a real `RevokeExternalIdpTokens`, wire it at `delete-account.main.ts:119` (treat already-revoked as success for SQS redrive idempotency).
- Also fix the **false comment** at `App/AffordancePresentation.swift:120` ("The server has already revoked everything server-side") — it hasn't, for Apple.

## 🟠 Scope — the one thing that does MORE than compliance

**2. The native iOS `delete-account` subsystem is unreachable dead code.**
`delete-account` is advertised only on the `/account` Siren entity (`api/account-siren.ts`); the queue collection — the only Siren resource iOS fetches — carries just an `account` **link** (`api/collection-siren.ts:44`), which `ToolbarRoute` routes to `.open` → opens `/account` in the webview. So the whole native apparatus never fires: the `delete-account` `AffordancePresentation` case, `Affordance.endsSession`, the `isDestructive` collection branch + `invokeCollection` + `forceLogout` in `ReadingListView.dispatch`/`confirmDestructive`, the account-specific confirm message, the two `AffordancePresentationTests`, the `delete-account` action in `account-siren.ts`, and the new `class?: string[]` field on `SirenAction` (`api/siren.ts`) are all dead. The web-link path already provides in-app deletion.
- **Preferred (compliance-only):** delete the native apparatus and rely on the web-link path. **Keep** the shared *row*-delete confirm infra (`pendingDestructive`, the `article` confirm path, `activate`) — that's live.
- **Or:** if native in-app deletion is actually the intended UX, advertise the `delete-account` action **on the collection** so the native flow goes live, and drop the web-link route. **Don't ship both half-wired.**

**3. No logout after the in-webview deletion completes.**
On the live web path, `POST /account/delete` redirects the webview to `/`, but the app keeps its `TokenStore` tokens and cached reading-list UI until a later 401 (`session.forceLogout()` only fires on the dead native path, `ReadingListView.swift:145`; `ReaderNavigation.decide` returns `.close` only for `readplace://reader/close`, which this flow never emits). Immediately after confirming deletion the app still looks signed in — a reviewer can read that as "deletion didn't take effect in the app." Fix: emit a deterministic done-signal (e.g. redirect the post-delete page to `readplace://reader/close`, which `ReaderNavigation` already maps to `.close`) and have that closure call `session.forceLogout()` + dismiss the sheet. (Resolving #2 via the native path gives this for free.)

## 🟡 Data remnants (deletion must actually erase)

**4. Password-reset scrub misses mixed-case emails — permanent PII leak.** `deleteTokensByEmail` (`providers/password-reset/dynamodb-password-reset.ts:87`) filters `email = :e` where `:e` is the normalized (lowercased) users-table PK, but rows are written with the **raw** email (`ForgotPasswordSchema` uses a bare `z.email()`, no normalize; stored verbatim at `dynamodb-password-reset.ts:52`). A reset requested as `John@Example.com` is never matched, and that table has **no TTL** → it survives forever, defeating the exact scan this PR added. Fix: normalize on write (and in the filter); add a mixed-case integration test.

**5. `verification-tokens` table is not scrubbed.** For a signed-up-but-never-verified account, `{token, userId, email}` survives up to ~9 days post-deletion (TTL + DynamoDB lag) — the same remnant class you scrubbed for password-reset. Add a `deleteVerificationTokensBy…` (Scan by email / by userId), wire it into the worker before `closeUserAccount`, and add its table to the worker's IAM grant in `infra/index.ts`.

**6. (note) `pending-signups` has no TTL** — abandoned-checkout rows keep the deleted user's `email`+`userId` permanently; `import-sessions` (TTL'd) keep `userId`+imported URLs until expiry. Decide explicitly whether to scrub (pending-signups is a permanent leak).

**7. (note)** The sync route kills sessions/tokens but the **identity row** is deleted by the async worker, so re-login is briefly possible in the gap (self-healing). Either delete/flag the row synchronously in the route, or soften the "immediately unusable" comment to "existing sessions/tokens killed synchronously; full erasure async."

## ✅ Confirmed in-scope — do NOT change
The async `DeleteAccountCommand` worker + infra Lambda/SQS/DLQ + every per-store delete/tombstone primitive (actual erasure is *required* by 5.1.1(v)); the `account` link + `SirenAction.class` (the entry-point mechanism); the ~15 `queue/*`, `view/*`, `save-*` route-test and fixture one-line additions (compile-forced by the new **required** `publishDeleteAccountCommand` dependency and the new collection link — not gratuitous); the `handle-subscription-cancelled` `ConditionalCheckFailed → no-op` (caused by deletion racing the Stripe `customer.subscription.deleted` webhook — would otherwise DLQ-poison); the `redirectFullPage` rename; and the privacy-policy rewrite (removing the email-only-deletion language is itself a compliance requirement). The inbox-address **tombstone** (PII stripped, hash reserved) is a defensible non-retention choice, not soft deletion.

---
_Reviewed for 5.1.1(v) completeness + scope only; findings verified against the PR head. Priorities: #1 (fail-closed the Apple trap) and #2 (resolve the dead-vs-web path into one) are the load-bearing ones; #4 is a real permanent-PII bug._

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: 10b7e0806cbad6d17b347421df7ae2ec13c7475f

FagnerMartinsBrack:
### Code Review — self-serve account deletion (App Store 5.1.1(v)) · re-review at `10b7e080`

Re-reviewed the six compliance-completeness items from the prior 5.1.1(v) pass and swept the fix commit for regressions. I traced each new scrub down to its IAM grant, env pairing, handler ordering, and idempotency/convergence behaviour rather than just confirming the code is present. **All five server-side items are implemented correctly and are genuinely sound.** The two iOS items remain (as the author flagged) and are the only open work — both need an iOS-side compile/test this Linux CI cannot run.

### High Priority Issues (Must Fix)

None found

### Medium Priority Issues

None found

**Verified — the five server-side compliance fixes:**

**#4 Password-reset mixed-case scrub.** `createPasswordResetToken` now stores `normalizeEmail(email)` (`dynamodb-password-reset.ts:57`) and `deleteTokensByEmail` filters `normalizeEmail(email)` (`:93`). Because both sides normalize and the function is idempotent, the match holds regardless of the casing `findEmailByUserId` returns — a `John@Example.com` reset is now scrubbed. `verifyPasswordResetToken` returns the (now normalized) stored email, consistent with the normalized users-table PK, so no reset-flow regression. In-memory fixture mirrors both sides. *(One forward-only caveat in Low #2.)*

**#5 Verification tokens.** New `deleteTokensByUserId` scans by `userId` and deletes each by the `token` PK, correctly aliased `#tk` (TOKEN is a DynamoDB reserved word). Wired into the worker **before** `closeUserAccount` (`delete-account-handler.ts:140`) and keyed by `userId` — not by the soon-deleted identity row — so a redrive re-derives and converges. IAM grant (`verificationTokensTable`, `includeIndexes:false`, correct for a Scan) + `DYNAMODB_VERIFICATION_TOKENS_TABLE` env are both present, and the actions list already carries `Scan`+`DeleteItem` (`infra/index.ts:547`, `553-554`).

**#6 Pending-signup.** Same shape: `deleteByUserId` scans by `userId`, deletes by the `checkoutSessionId` PK, wired before `closeUserAccount` (`:141`), IAM + env paired. `import-sessions` correctly left to its TTL.

**#1 SIWA fail-closed guard.** The new `auth.route.test.ts` test couples "SIWA reachable without a flag" to "`AppleTokenResponse` retains `refresh_token`" and pins both `false`. Because `z.object` strips unknown keys, the proxy reads `false` today and flips only when someone adds `refresh_token` to the schema — and the hard assertion turns CI red on **any** drift, including the dangerous un-gate-without-revocation state `{reachable:true, persisted:false}`. The seam comment and the corrected Swift comment now both tell the truth about Apple not being revoked yet. *(Comment-vs-assertion nuance in Low #1.)*

**#7 Comment accuracy.** The `/account/delete` doc-comment now states erasure is eventual (existing credentials die synchronously; the identity row is removed async, self-healing) rather than "immediately unusable." Accurate.

### Low Priority Suggestions

**1. SIWA guard comment describes a two-state acceptance the assertion doesn't implement.** The comment says the accepted states are `{false,false}` (today) and `{true,true}` (SIWA live + revocation wired), but the assertion is a hard `.toEqual({false,false})`, so the correct `{true,true}` future state also fails CI. That's a defensible "revisit on any change" tripwire (the schema-field proxy is imperfect, so forcing a human touch when SIWA ships is arguably safer), but the comment over-promises. Either tighten the comment to "pins today's state; revisit when SIWA ships," or encode the coupling as `expect(siwaReachableByDefault).toBe(appleRefreshTokenPersisted)` — which passes for both stated states and still fails the dangerous `{true,false}`.

**2. Password-reset normalize is forward-only.** `normalizeEmail`-on-write fixes new rows, but a mixed-case row written **before** this deploys is still stored raw and won't match the normalized scrub. Since this table has no TTL (the premise of #4), such a row for a later-deleted account persists. A one-off backfill/normalization of existing rows would fully close the leak the fix targets.

**3. (iOS — needs an iOS-side change) Native `delete-account` apparatus is still unreachable.** The queue collection carries only an `account` **link** → webview (`collection-siren.ts:44`), so the native `delete-account` action, `endsSession`, the `isDestructive`→`forceLogout` branch, and the two `AffordancePresentationTests` never fire. Either delete the native apparatus (compliance-only, keep the live web-link path) or advertise `delete-account` on the collection to make it live — don't ship both half-wired. Requires Swift compile/test this CI can't run; the author already documented the exact steps.

**4. (iOS — needs an iOS-side change) No logout after in-webview deletion.** `POST /account/delete` redirects the webview to `/` (`account.page.ts:351`), but the app keeps its `TokenStore` + cached UI until a later 401 — right after confirming deletion the app still looks signed in, which a reviewer can read as "deletion didn't take." Emit a deterministic done-signal for the webview (e.g. `readplace://reader/close`, which `ReaderNavigation` already maps to `.close`) and have that closure `forceLogout()` + dismiss. Server-only detection is fragile (a plain browser must still get `/`), so this also needs an iOS-side change.

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: e26ea17aa3b53bf8dbe88610ca1e2ff4ff4c8c5f

FagnerMartinsBrack:
### Code Review — self-serve account deletion (App Store 5.1.1(v)) · re-review at `e26ea17a`

Focused this pass on the three low-priority fixes landed in `e26ea17a` (SIWA coupling, password-reset TTL, dead native-apparatus removal) and swept for regressions. I traced the TTL against the row's actual `expiresAt` format, confirmed the Siren/Swift deletions are dangling-reference-free, and verified the in-memory fixture still mirrors the provider.

**The three fixes are correct.** One documentation contradiction was introduced by the TTL fix.

- **SIWA coupling (Low #1)** — `expect(siwaReachableByDefault).toBe(appleRefreshTokenPersisted)` now implements the two-state acceptance the comment describes: `{false,false}` (today) and `{true,true}` (SIWA live + revocation wired) pass; the dangerous un-gate-without-revocation `{true,false}` still fails CI. Since `AppleTokenResponse` (`apple-token.ts`) parses only `id_token`, zod strips `refresh_token`, so the proxy reads `false` today. Sound. ✅
- **Password-reset TTL (Low #2)** — `expiresAt` is written as `Math.floor(Date.now()/1000) + TTL` (`dynamodb-password-reset.ts:51`), i.e. **epoch seconds**, so the `ttl.attributeName: "expiresAt"` in `hutch-storage.ts` is valid and will actually fire (a common footgun — ISO-string or ms values silently never expire — correctly avoided). A valid token (`expiresAt = now+1h`) is never prematurely purged; only already-expired rows become TTL-eligible, so the backstop erases pre-normalization mixed-case stragglers with no premature-deletion risk. ✅
- **Dead native apparatus (Low #3)** — `account-siren.ts` (+test), the `GET /account` Siren branch (+route test), the Swift `delete-account` presentation case, `endsSession`, and the two iOS tests are all removed; grep confirms **zero dangling references** to `toAccountEntity`/`endsSession`/`wantsSiren` in `account.page.ts`. The live web Danger Zone, the `account` collection link, and the generic `isDestructive` row-confirm infra are retained. ✅

None found

## Medium Priority Issues

### 1. Stale "no TTL" comments now contradict the TTL added in this same commit

`e26ea17a` enables DynamoDB TTL on the password-reset table (`hutch-storage.ts:203-206`), but two comments in [`dynamodb-password-reset.ts`](projects/hutch/src/runtime/providers/password-reset/dynamodb-password-reset.ts) still assert the table has **no** TTL:

- `:55-56` (in `createPasswordResetToken`): *"…and — since this table has no TTL — the address would outlive the account."*
- `:87-88` (in `deleteTokensByEmail`): *"This table carries no TTL attribute, so an unexpired reset token would outlive the deleted account unless every row for the email is purged now."*

Both are now factually wrong — the table **does** carry a TTL on `expiresAt` as of this commit. This isn't a logic bug (the code is correct: the synchronous scrub still erases new rows immediately, and TTL is the ~48h backstop), but it's a flat self-contradiction introduced within one commit, on the exact PII-erasure mechanism this compliance PR is about. A future maintainer reading "carries no TTL attribute" directly above the scan-delete could reason incorrectly about which path erases the data (e.g. treat the scan as the *only* erasure and mis-handle the TTL). Update both comments to state the table now has a TTL on `expiresAt` — synchronous scrub = immediate erasure, TTL = the backstop for pre-normalization stragglers. [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Faccount-deletion-app-store-rasc6i%2C%20commit%20e26ea17a%20enabled%20DynamoDB%20TTL%20on%20the%20password-reset-tokens%20table%20in%20projects%2Fhutch%2Fsrc%2Finfra%2Fhutch-storage.ts%20%28ttl%20on%20expiresAt%29%2C%20but%20two%20comments%20in%20projects%2Fhutch%2Fsrc%2Fruntime%2Fproviders%2Fpassword-reset%2Fdynamodb-password-reset.ts%20still%20claim%20the%20table%20has%20no%20TTL%3A%20lines%2055-56%20in%20createPasswordResetToken%20and%20lines%2087-88%20in%20deleteTokensByEmail.%20Update%20both%20comments%20to%20reflect%20that%20the%20table%20now%20has%20a%20TTL%20on%20expiresAt%20-%20the%20synchronous%20scrub%20still%20erases%20new%20rows%20immediately%20and%20TTL%20is%20the%20~48h%20backstop%20for%20pre-normalization%20mixed-case%20stragglers.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. `SirenAction.class` is now producer-less; its comment reads as live wiring
With `account-siren.ts` deleted, the `class?: string[]` field on `SirenAction` ([`siren.ts:14-17`](projects/hutch/src/runtime/web/api/siren.ts#L14)) has **no server producer** — grep confirms every `class:` in `web/api/` is an *entity* class (`collection`/`article`/`error`), never an *action* class. iOS also derives `isDestructive` from the action **name** (`AffordancePresentation(token:)`), not from this field, so it's read on neither end. You deliberately kept it per the reviewer's "do NOT change (the entry-point mechanism)", which is fine — but the comment ("*the client maps to its own presentation … `["destructive"]` → a red, confirm-gated control*") now describes a mechanism that isn't wired anywhere. Consider a one-line "not currently emitted; reserved for when the native destructive path ships" note so it doesn't read as active contract, or remove it alongside the native path when that decision lands.

### 2. iOS Swift changes are deletion-only and unverified here; two iOS follow-ups still open
The Swift edits in this commit are pure deletions and statically coherent (no remaining `endsSession`/`delete-account` references), but this Linux CI cannot compile or `make test` them — they need an iOS-side run before merge, as the author flagged. The two documented iOS follow-ups remain open and are the only substantive work left: (a) **native-vs-web path** — the queue collection carries only an `account` link → webview, so in-app deletion is served entirely by the web Danger Zone (this is now internally consistent after the native removal); (b) **no deterministic logout after in-webview deletion** — `POST /account/delete` redirects the webview to `/` while the app keeps its `TokenStore`/UI until the next-API-call 401 (self-heals, but a reviewer mid-session can read it as "deletion didn't take"). Both need a coordinated iOS+server change this CI can't validate.

### 3. PR description D4 is now stale
The decision table says password-reset tokens are *"Scan-deleted by email (that table has no TTL, unlike its siblings)"* — this commit gave the table a TTL, so D4's parenthetical no longer holds. Minor description sync (not code).

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Review to Address

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **Run ID**: 28787358580
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:
@claude — re-reviewed the PR at `f204198c` against all seven findings from my 5.1.1(v) pass. Verdict: **everything substantive is fixed and verified**. Four small items remain below — two trivial, one small logic change, one record correction.

## Verified fixed (no action)
- **SIWA fail-closed guard** — genuinely load-bearing, not symbolic: both sides of the `toBe` equality are computed live (reachability through the real `createApp` render + `[data-test-apple-section]` probe; persistence through the real `AppleTokenResponse` zod strip). Un-gating SIWA without persisting the refresh token turns CI red. Good design.
- **Dead native apparatus removed** — `account-siren.ts` gone, no dangling Swift symbols, row-delete confirm infra and the `account` link correctly kept.
- **PII scrubs** — password-reset normalize-on-write + filter + table TTL (historical rows already carry numeric epoch `expiresAt`, so no backfill needed — verified); verification-tokens and pending-signups scrubbed with IAM + env wired end-to-end; idempotency holds on redelivery. The `f204198c` comment correction is accurate.
- **✅ iOS `make test` run on the PR head (at `e26ea17a`; the delta since is TS-comment-only)**: full production suite + staging smoke pass, 0 failures — this closes the "iOS changes need an iOS-side `make test` run" caveat from the PR body and fix commits. (Note: the iOS per-file coverage gate lives only on the `ios/server-reader-signal` branch, not main, so no gate applies to this PR; whichever merges second reconciles.)

## Remaining — please apply

**1. Legacy email-only pending-signup rows still leak (bounded, permanent).** `deletePendingSignupsByUserId` filters `userId = :u`, but pre-Phase-1 rows carry `{checkoutSessionId, method, email}` with **no userId** (the same rows `consumePendingSignup` discards at `dynamodb-pending-signup.ts:96`), and the table has no TTL — so an abandoned legacy row keeps the deleted user's email forever. The delete worker already captures `email` before `closeUserAccount` (it passes it to the password-reset scrub): pass it to the pending-signup scrub too, project `{checkoutSessionId, email, userId}` in the existing paginated scan, and delete rows where `userId === target` **or** `normalizeEmail(email) === normalizedTarget` (client-side compare — legacy rows stored raw casing). Mirror the fixture + a mixed-case legacy-row test.

**2. Remove the orphaned `SirenAction.class` field** (`web/api/siren.ts` — the `class?: string[]` on the *action* interface plus its comment). Its only emitter was the deleted `account-siren.ts`; no client ever decoded it (the iOS `SirenAction` `CodingKeys` and the extension zod schemas parse only entity-level `class`), no test pins it, and the hypermedia skill says not to add the semantic-role token until a real client need exists. My earlier "do NOT change" instruction predated the account-siren removal; this corrects it. (Entity-level `class` stays — it's widely used.)

**3. Record the import-sessions accept-via-TTL decision in-repo.** It currently lives only in this PR conversation. Add one sentence to the delete worker's scrub-inventory comment (the "Signup/verification remnants" block in `delete-account-handler.ts`, before `deleteVerificationTokensByUserId`), e.g.: "Import sessions are deliberately NOT scrubbed: their `{userId, urls}` rows self-expire via the table's 24h TTL, so no scan is spent on them."

**4. Correct the deferral record for the post-webview logout (Low #4).** The stated rationale "self-heals via the next-API-call 401" overstates promptness, verified on the head: **no API call fires on sheet dismiss** (`onClose` only nils `readerPresentation`), `scenePhase` never leaves `.active` for an in-app sheet, and once the list has paginated, `convergeWithServer`'s `!hasPaginated` guard makes the foreground refresh **zero-network** — so a deep-scrolled list can display the deleted account's articles for the entire process lifetime until a user-initiated call (pull-to-refresh, row tap, cold relaunch). No server data is at risk (every call 401s), but the follow-up should be specified as: trigger a converge (or a bare authenticated ping) on web-sheet dismissal, which 401s and funnels into the existing `onSessionExpired → forceLogout`. No code change now — carry this corrected spec into the follow-up so it isn't scoped off the "self-heals" framing.

---
_Re-verified per-finding against the PR head; hutch `check` green on the head (regression sweep) and iOS `make test` green locally. After item 1 this PR is, to my reading, complete and correctly scoped for 5.1.1(v)._

FagnerMartinsBrack:

## Context
- **Attempt**: 3 of 5
- **Run ID**: 28789287793
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:
@claude apple signin is meant to be released

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: c2a53a79ab74ec81c4184bf977d6104e109c571b

FagnerMartinsBrack:
### Code Review — Apple grant revocation at account deletion · re-review at `c2a53a79`

The only new commit since the last green review (`8423c4de`) is `c2a53a79`, which replaces the no-op Apple-revocation seam with the real thing so the fail-closed SIWA guard can pass in the `{reachable:true, persisted:true}` state. I traced the change end-to-end — schema, persistence, the revoke call, handler ordering, infra/IAM, and the guard — rather than just confirming the code is present. **It is correctly and completely wired, and genuinely sound.**

### High Priority Issues (Must Fix)

None found

### Medium Priority Issues

None found

**Verified — the Apple revocation is wired end-to-end:**

- **Exchange schema is fail-closed and safe.** `AppleTokenResponse` now *requires* `refresh_token` (`apple-token.ts:12`). The exchange uses `grant_type: "authorization_code"` (`apple-token.ts:40`), for which Apple always returns a `refresh_token`, so requiring it does **not** risk breaking a legitimate login — and an exchange that somehow omits it now fails the login (tested: *"rejects a token response that is missing the refresh_token"*) rather than minting an unrevokable account. Correct compliance trade-off.
- **Persistence covers signup + every returning login (backfill).** `createAppleUser` stores the token at signup; `saveAppleRefreshToken` (conditional `attribute_exists(email)` update, `dynamodb-auth.ts`) stores the freshest grant on each returning Apple login — including the existing-account and both create-race-fallback branches (`apple-auth.page.ts:204/245/288`), so accounts minted during the #933→this gap self-heal on next login. Google/password rows stay token-free (tested). Read back via `findAppleRefreshTokenByUserId` over `userId-index`, which is `hashKey: userId, projectionType: "ALL"` (`hutch-storage.ts:112-115`) — so the token is actually projected — and deleted with the row by `closeUserAccount`. The in-memory fixture mirrors both the conditional-update throw and the index read exactly, and every path is unit-tested (provider + fixture + route).
- **Revoker converges on redrive.** `initRevokeExternalIdpTokens` returns early for a `null` token (password/Google → no Apple call, tested), else POSTs `https://appleid.apple.com/auth/revoke` with a freshly minted ES256 client secret and `token_type_hint: "refresh_token"`, and `assert(response.ok, …)` throws on non-2xx so the SQS record redrives. Apple returns 200 for an already-revoked token (RFC 7009), so redrives converge. All three branches tested, plus an end-to-end handler test proving the stored grant is revoked and the account fully scrubbed.
- **Ordering keeps the token readable across redrives.** Revocation (`delete-account-handler.ts:148`) runs *before* `closeUserAccount` (`:151`), so the users row — and its `appleRefreshToken` — survives until the very end; a mid-scrub failure or an Apple outage re-reads and re-revokes on redrive.
- **Infra matches.** The worker Lambda gets the four `APPLE_LOGIN_*` env vars (`index.ts:627-630`); its IAM grants `usersTable` with `includeIndexes: true` (`index.ts:535`) and `dynamodb:Query` (`:552`), so the `userId-index` read is permitted; `dlqMaxReceiveCount` rises 3→12 (`:597`, ≈3h) so an Apple outage rides out in-queue instead of hitting the email-alarmed DLQ after the default 3 attempts.
- **Guard flips correctly.** `expect(siwaReachableByDefault).toBe(appleRefreshTokenPersisted)` (`auth.route.test.ts:1284`) is now `{true,true}` and still turns CI red on the dangerous `{true,false}` (dropping `refresh_token` while SIWA stays reachable). The renamed `initRevokeExternalIdpTokens` has no dangling references to the old no-op.

### Low Priority Suggestions

**1. The guard's symmetric coupling also fails the harmless `{reachable:false, persisted:true}` state.** `.toBe` couples the two as equality, so a future *re-dark-launch of SIWA that keeps the token-persistence code* trips CI even though a persisted-but-unused token behind a gated button is not itself a 5.1.1(v) risk. The rewritten comment frames this as "a re-dark-launch that also retires the token," i.e. it deliberately treats the two as moving together. That's a defensible tripwire (it was chosen over a one-directional assert in the `e26ea17a` review), so this is only a confirm-intent note — the coupling is stricter than compliance strictly requires, and that's on purpose.

**2. (iOS — needs an iOS-side change, unchanged by this commit) No deterministic logout after in-webview deletion.** The prior review's Low #4 remains the only substantive open item: `POST /account/delete` redirects the webview to `/` while the app keeps its `TokenStore`/cached UI until a user-initiated call 401s, so right after confirming deletion the app can still look signed in. This commit is entirely server-side (SIWA is web OAuth opened in Chrome; no Swift touched), so it neither introduces nor resolves that item — carry the corrected "trigger a converge/ping on web-sheet dismissal → 401 → `onSessionExpired → forceLogout`" spec into the iOS follow-up, verified with `make test`.

FagnerMartinsBrack:
@claude address this low prio issue, do not report the other again: 2. (iOS — needs an iOS-side change, unchanged by this commit) No deterministic logout after in-webview deletion. The prior review's Low https://github.com/Readplace/readplace.com/pull/4 remains the only substantive open item: POST /account/delete redirects the webview to / while the app keeps its TokenStore/cached UI until a user-initiated call 401s, so right after confirming deletion the app can still look signed in. This commit is entirely server-side (SIWA is web OAuth opened in Chrome; no Swift touched), so it neither introduces nor resolves that item — carry the corrected "trigger a converge/ping on web-sheet dismissal → 401 → onSessionExpired → forceLogout" spec into the iOS follow-up, verified with make test.

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: 01e7ad2f5a4fece41d006d886ef8b0e025d466fd

FagnerMartinsBrack:

None found

## Medium Priority Issues

### 1. Merge gap — the new `trial-reminder` per-user schedule (#939) is not scrubbed by the delete-account worker

The last green review was at `c2a53a79`; the only new commit is the merge `01e7ad2f`, which brings in **#939 `feat(conversion): pre-expiry trial reminder email`**. That merge added a **fourth** per-user EventBridge schedule, `trial-reminder-${userId}` (`aws-trial-scheduler.ts:206` `createTrialReminderSchedule`), created for **every trialing user** at signup / trial-start (`auth.page.ts:352`, `apple-auth.page.ts:327`, `google-auth.page.ts:261`, `account.page.ts:395`).

The delete-account worker deletes only **three** schedules (`delete-account-handler.ts:104-106` — trial-end, deferred-cancellation, trial-feedback), and its comment asserts: *"Delete **every** per-user schedule so a later fire can't dispatch a command at a deleted account. **All three** are ResourceNotFound-idempotent."* After the merge there are four, and the reminder is the uncovered one — so both the handler's invariant and the PR-body goal (*"Per-user EventBridge schedules are deleted so none fires at a dead account"*) no longer hold. The handler test wires/asserts only the same three, so nothing catches it.

**Impact is bounded but real.** A user who deletes their account **while trialing** leaves an orphaned `trial-reminder-<userId>` schedule (its `Name` and `Input` carry the `userId`) live at AWS until its natural fire time. It does **not** email the deleted user and does **not** poison the queue — `processReminder` (`send-trial-feedback-email-handler.ts:148-155`) no-ops when the subscription row is gone (deleted synchronously by `deleteSubscription`), and the schedule self-deletes on fire (`ActionAfterCompletion: "DELETE"`). But a compliance-driven deletion feature leaving a **user-keyed resource** behind is exactly the gap this PR exists to close (same rationale as the verification-token and pending-signup scrubs the author added), and the handler's comment now misstates the count.

**The obligation is already established in-tree:** `cancel-subscription-handler.ts:84` deletes this very schedule on a trialing cancel — *"A cancelling trialist must not receive the pre-expiry subscribe nudge … deleting the schedule is defence-in-depth plus scheduler hygiene."* Account deletion is the strictly stronger case; the merge just didn't carry the new schedule into the pre-#939 delete worker.

**Fix is small — no IAM change.** The worker's `scheduler:DeleteSchedule` grant (`infra/index.ts:269-270`) is a **group-level wildcard** (`…:schedule/<group>*`), so deleting the reminder schedule is already permitted. Add `deleteTrialReminderSchedule: DeleteTrialReminderSchedule` to `DeleteAccountHandlerDependencies`, call `await deps.deleteTrialReminderSchedule({ userId })` alongside the other three, correct the comment to "all four", wire `trialScheduler.deleteTrialReminderSchedule` at `delete-account.main.ts:159`, and extend the handler test (wire + assert the reminder delete, including on the redrive cases).

**Note — the merge-conflict resolution itself is correct.** `dynamodb-subscription-writes.ts` carries **both** `markTrialReminderEmailSent` (#939) and `deleteSubscription` (this PR), the doc-comment correctly reads "eight mutations", the in-memory fixture mirrors both, the interleaved `describe`/`it` blocks are cleanly split, and there are no leftover conflict markers. The `trialReminderEmailSentAt` column #939 adds lives on the subscription row, which `deleteSubscription` already removes — so the schedule above is the *only* new user-owned resource the merge left unscrubbed.

## Low Priority Suggestions

### 1. (iOS — needs an iOS-side change) Deterministic logout after in-webview deletion is still open

The `webSheetDismissed` follow-up from 2026-07-06 never landed (its job errored before pushing; `git grep` confirms it's absent from the tree), so this remains the one open non-blocker: `POST /account/delete` redirects the webview to `/` while the app keeps its `TokenStore` + cached reading-list UI until a user-initiated call 401s. No server data is at risk (every call 401s), but right after confirming deletion the app can still look signed in. The corrected spec — trigger a converge/authenticated ping on web-sheet dismissal → 401 → existing `onSessionExpired → forceLogout` — needs a coordinated iOS+server change verified with `make test`, which this Linux CI can't run.

*(Per the maintainer's 2026-07-06 instruction, not re-reporting the SIWA symmetric-coupling confirm-intent note.)*

FagnerMartinsBrack:

## Context
- **Attempt**: 3 of 5
- **PR**: #940
- **Branch**: claude/account-deletion-app-store-rasc6i

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #940
- **Commit**: fb3d8bb77450a72aa481fbde86948e0d73a7dc53

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

**Verified — the trial-reminder schedule scrub fix (`fb3d8bb7`, the M1 from the last review at `01e7ad2f`):**

- **Wired end-to-end.** `DeleteTrialReminderSchedule` is added to `DeleteAccountHandlerDependencies` and called at `delete-account-handler.ts:109` alongside the other three schedule deletes, and the composition root passes `trialScheduler.deleteTrialReminderSchedule` (`delete-account.main.ts:160`). The stale `// All three …` comment is corrected to `// All four …` (`delete-account-handler.ts:105`).
- **Idempotent on redrive.** `deleteTrialReminderSchedule` (`aws-trial-scheduler.ts:236-250`) swallows `ResourceNotFoundException` exactly like its three siblings, so an at-least-once redrive converges.
- **Ordering is right.** The four schedule deletes run early (`:106-109`), before the inbox/store scrub, so a mid-scrub failure re-runs them harmlessly on retry — consistent with the handler's convergence invariant.
- **Exhaustive — no fifth schedule missed.** Exactly four per-user schedule-name helpers exist (`trial-end`, `cancel-`, `trial-feedback-`, `trial-reminder-`), and the worker now deletes all four; the reminder was the only gap the #939 merge introduced.
- **No IAM change needed — verified.** The worker's `scheduler:DeleteSchedule` grant is a group-level wildcard `…:schedule/<group>*` (`infra/index.ts:269-270`), which already matches `trial-reminder-<userId>`. The sibling `cancel-subscription-handler.ts:84` deletes this same schedule on a trialing cancel; account deletion is the strictly stronger case.
- **Test coverage complete.** The trialing and founding-member branches both assert `trialReminderCalls === [userId]`, and the raw-email redrive test asserts `[userId, userId]` — proving the delete fires on both the failed first delivery and the redrive and is idempotent. Both handler construction sites (production + test) carry the new required dep. CI is green on the commit (run 28832966885).

## Low Priority Suggestions

### 1. (iOS — needs an iOS-side change) Deterministic logout after in-webview deletion is still the one open item

Unchanged by this server-only commit and carried over: `POST /account/delete` redirects the webview to `/` while the app keeps its `TokenStore` + cached reading-list UI until a user-initiated call 401s, so right after confirming deletion the app can still look signed in. No server data is at risk (every call 401s). The corrected follow-up spec — trigger a converge/authenticated ping on web-sheet dismissal → 401 → existing `onSessionExpired → forceLogout` — needs a coordinated iOS+server change verified with `make test`, which this Linux CI can't run.

*(Per the maintainer's 2026-07-06 instruction, not re-reporting the SIWA symmetric-coupling confirm-intent note.)*

PR #937, opened 2026-07-05
fix(save-pipeline): count user-facing save failures + stop the generate-summary DLQ loop

FagnerMartinsBrack:
# fix(save-pipeline): count user-facing save failures + stop the generate-summary DLQ loop

## Hypothesis

Two visibility holes hide save-pipeline pain: (1) `view_save_intent` has an `outcome="error"` enum value that is never emitted for the failures users actually hit (URL validation 422s), so the save funnel reads as if saves never fail; (2) the `generate-summary` worker treats "canonical content absent because the crawl terminally failed" identically to "content not yet written", producing a permanent retry/DLQ loop via the summary auto-heal that fires almost daily and drowns the errors dashboard.

## Evidence (30d, Jun 4 – Jul 4 2026)

- `view_save_intent outcome="error"`: **0 events in 30d** — yet the enum exists and the catch-path emits are wired. Every `validateSaveableUrl` 422 branch (queue save bar, extension save-article/save-html/save-content) renders a user-facing save error without emitting.
- **170 generate-summary DLQ invocations**, sole error `AssertionError "Article content not found: <url>"`, ~520 retry ERROR lines, near-daily. Mechanism (verified in code): `markCrawlExhausted` leaves `summary=failed(crawl-failed)` → stale-check auto-heal reprimes **any** `summary=failed` row (3 attempts, then one round per 24h forever) → GenerateSummary finds no canonical content (crawl is terminal) → assert → 4 receives → DLQ → `markSummaryExhausted` → `failed(exhausted-retries)` → eligible again next tick. The 73 crawl-failed dead-lettered anonymous first-saves are the seed population.
- Duplicated reason suffix `"exhausted-retries (receiveCount=4) (receiveCount=4)"` (167/170 events): **already fixed on main** by #886 (commit `3dd818a5`, Jul 3) with a regression test — intentionally not touched here.

## What this changes

1. **Emit `view_save_intent outcome="error"` on every synchronous save failure**: queue save bar validation 422, extension save-article validation 422, save-html invalid-request/invalid-URL 422s, save-content missing-field/invalid-URL/unsupported-media/not-a-pdf 422s. `buildSaveIntentEvent` now derives `article_host` via a null-safe parser (`article_host`/`content_class` become nullable — dashboards already `coalesce()` these). Deliberately not instrumented: bulk save-all-tabs per-entry skips (expected filtering of chrome:// tabs; per-job failures already emit), multipart-parse 422s/406s (no URL available), and async crawl failures (out of scope by design).
[...]

PR #935, opened 2026-07-05
feat(observability): instrument the checkout funnel with started/completed/failed events

FagnerMartinsBrack:
### Hypothesis
The stage between the /account Subscribe click and revenue is dark. If we emit structured events at checkout start, checkout completion, and every checkout-return failure, we can see where the 0% trial→paid conversion is leaking — and measure any future fix instead of guessing.

### Evidence (30d, Jun 4 – Jul 4 2026)
- 35 signups, 28 trials reached end-of-trial: **28× charge_failed (no_card_on_file) → 28× cancelled**. Trial→paid = **0%**.
- Upgrade funnel: 54 header trial-countdown clicks / 40 visitors → 22 /account views / 11 visitors → **4 subscribe clicks / 4 users → 2 /auth/checkout/success returns → 0 recorded conversion events**. Checkout success emitted nothing (`upsertActive` directly), so the two known conversions (3e9c7e5c Jun 21, af194ac4 Jun 23) were invisible to every dashboard.
- The only Stripe webhook in 30 days was dropped ("no subscription row found — skipping event emission"), so there is no backstop signal. Only 3 of 45 subscription rows ever carried a Stripe subscriptionId.

### What this changes
Four new events on the existing `subscriptions` stream, emitted by the hutch web app (not the subscription Lambdas):

| Event | Where | Fields |
|---|---|---|
| `checkout_started` | `POST /account/subscribe`, after the Stripe Checkout session + pending-signup write | `user_id`, `variant` ∈ `trial_checkout` \| `cancelled_resubscribe` \| `card_decline_fallback`, `checkout_session_id` |
| `checkout_completed` | `GET /auth/checkout/success` completed path | `user_id`, `subscription_id`, `checkout_session_id`, **`paid_now`**, `variant` |
| `checkout_return_failed` | each failure branch of `GET /auth/checkout/success` | `reason` ∈ `invalid_query` (400) \| `session_not_found` (404) \| `not_paid` (402) \| `replayed` (409), optional `user_id` / `checkout_session_id` |
| `resubscribe_completed` | `POST /account/subscribe`, saved-card one-click resubscribe (charges immediately, never touches Stripe Checkout, so it has no `checkout_started` to pair with) | `user_id`, `subscription_id`, `paid_now` (always true) |

[...]

PR #918, opened 2026-07-04
feat(hutch,@packages/web-analytics): emit signup_attempted to make the signup form measurable

FagnerMartinsBrack:
## Hypothesis

The signup form is a black box. `POST /signup` emits nothing on any outcome except success (`user_created`), so a visitor who **submitted and was rejected** — disposable email, duplicate account, failed validation — is indistinguishable in analytics from one who bounced without ever submitting. In particular, **disposable-email blocking is a deliberate product friction whose signup cost has never been measured.** This PR emits a `signup_attempted` analytics event at every terminal branch of the handler so the form's own conversion — and the share lost to each gate — becomes countable.

## Evidence (production, last 30 days)

- **214 distinct anonymous visitors** hit `/signup`; **35** accounts were created. The ~180 who didn't convert are currently unattributable: bounce vs. rejected-submission is invisible.
- The repo rejects disposable-email domains at signup (commit `e87d530f`) — a real gate with an unmeasured cost. If a non-trivial share of those ~180 are being turned away at submit time, that's recoverable signup volume hiding in plain sight.
- Instrumentation audit flagged this as the #2 dark funnel stage (after the blog): "a visitor who submitted and was rejected is indistinguishable from one who bounced."

## What this changes

A new `signup_attempted` event (analytics stream) emitted at every terminal outcome of `POST /signup`:

| `outcome` | When |
|---|---|
| `created` | account created (free/founding or trial path) |
| `disposable_email` | rejected because the email domain is on the disposable blocklist |
| `invalid_input` | other schema failure (weak password, invalid email) |
| `duplicate_email` | address already has an account (validation, or the insert-time race) |
| `too_fast` | submitted inside the 2.5s bot window and shown the form again |

- Carries `visitor_id` (joins to `user_created` and the reader funnel) and `visitor_hash` (dashboard owner-exclusion).
- **`too_fast` is counted because the visitor sees the rejection.** Since `ab75c208` that branch re-renders the form with a 422 rather than the silent fake-success 303 the other bot reasons get — password-manager autofill routinely submits inside the window — so a hasty human who is told to retry and gives up is exactly the submitted-and-rejected cohort this event exists to surface. The genuinely silent trips (`honeypot`, `missing_timestamp`, `invalid_timestamp`) are fake-succeeded, never shown a rejection, and stay on the `bot-defense` stream only. The per-IP rate-limit 429 short-circuits in middleware before the handler, so it is out of scope. Each arm is pinned by a test.
[...]

PR #943, opened 2026-07-05
feat(hutch): auto-save the first article for new signups

FagnerMartinsBrack:
### First-article auto-save for new signups

**Hypothesis.** A new user whose queue already contains the article that brought them to Readplace is far likelier to activate (read something) than one who lands on an empty queue. If we quietly save that article at signup, the queue is non-empty on first sight and the "there's nothing here" cold-start is gone.

**Evidence (production).** Only **4 of 35** new users ever read anything. **23 of 34** signups arrive **without** a pending save — they clicked a nav/homepage "Sign up" CTA, not the reader's "Save to My Queue" button — so the existing pending-save round-trip (which already carries a held save through signup) never fires for them. Those 23 are exactly the cohort this change targets: they were reading an article on `/view` immediately before signing up, but nothing captured it.

**What this changes.**
- On an **anonymous, non-bot** `/view` open, set a short-lived httpOnly `hutch_lastview` cookie (2h, SameSite=Lax, Secure per origin) holding the validated article URL. It sits behind the same `isbot` gate that emits the existing `view_opened` event **plus** the `Sec-Purpose`/`Purpose: prefetch` check that already guards the crawl budget — so crawlers and speculative prefetches never set it (a prefetched article the reader never opened must not claim the autosave slot), and authenticated viewers don't either.
- At **signup success** (all three paths: email free+trial, Google, Apple), if the post-auth redirect would be the default `/queue` (no explicit `?return=`) and a valid `hutch_lastview` cookie is present, redirect to `/queue?url=<article>&utm_source=signup-autosave` instead. The existing `/queue?url=…` auto-submit save form then saves the article into the brand-new queue — reusing the stub-save + async-enrich pipeline, so the card appears immediately and fills in.
- A single pure `resolvePostSignupRedirect({ returnUrl, lastViewUrl })` is consulted at all three sites so they can't drift. It **never** auto-saves when an explicit `?return=` is present (the `/save` round-trip already saves in that case — no double-save) or when the cookie is absent/tampered (URL re-validated with the same `validateSaveableUrl` used everywhere else).
- The cookie is cleared on consumption and on `POST /logout`. For Apple's cross-site `form_post` callback (which drops first-party Lax cookies) the URL is tunneled inside the existing HMAC-signed `state`, exactly like `pendingSaveId` already is — but only while the serialized signed state stays within a 4 KB-safe cookie budget; past that the URL is dropped and autosave degrades to a plain `/queue` rather than risking the whole state cookie (and the nonce with it).

**Estimated result.** The 23/34 no-pending-save signups start with a 1-article queue instead of an empty one. If even a fraction of that cold-start friction is what suppresses the current 4/35 (~11%) read rate, moving first-article-in-queue to ~100% of eligible signups should lift first-week activation meaningfully. We'll read the actual lift from the funnel below rather than guess a number.

**How to measure (CloudWatch Logs Insights, hutch log group `/aws/lambda/hutch-handler`).** A discrete `first_article_autosaved` event is emitted server-side **exactly once, at the post-signup redirect decision**, across all three signup paths — so the trigger count is 1:1 by construction and immune to the reload/share recounting a URL marker is prone to. It carries `user_id`, `article_host`, `visitor_hash` (so the dashboard's internal-visitor exclusion applies), and `visitor_id` (except Apple signups where the id wasn't tunneled through the cross-site state). It is surfaced on the new **"First-article autosaves per day"** dashboard widget:

```
fields @timestamp, user_id, visitor_id
| filter stream = "analytics" and event = "first_article_autosaved"
| stats count(*) as autosaves by bin(1d)
```

The `utm_source=signup-autosave` marker on the post-signup `/queue` pageview is retained as a secondary tag (the save form's action is a fixed internal-tracking link, so the marker lands on the pageview, not on `view_save_intent`). To confirm each autosave actually **persisted**, join the trigger event to the follow-on save on the same visitor:

```
fields @timestamp, visitor_id, event, surface, outcome
| filter stream = "analytics"
    and (
      event = "first_article_autosaved"
      or (event = "view_save_intent" and surface = "queue_save_bar" and outcome = "saved")
    )
| sort visitor_id, @timestamp
| stats
    sum(event = "first_article_autosaved") as autosaves_triggered,
    sum(outcome = "saved") as saves_confirmed
  by visitor_id
```

Activation lift is then the read-rate (`event = "article_read"`) among the autosaved cohort vs. the historical 4/35 baseline, joined by `user_id` (`article_read` carries `user_id`, not `visitor_id`).

**Analytics note.** This change adds **one new analytics event** (`first_article_autosaved` in `ANALYTICS_EVENTS`) and **one new dashboard widget**, so the analytics-dashboard widget-count assertion moves **34 → 35** (the follow-up commit message's "28 → 29" predates rebases onto a busier dashboard — 35 is the number on this branch). The discrete event is preferred over leaning on the `utm_source` string because a reload, share, or bookmark of the post-signup `/queue?url=…` URL would recount the marker; the event fires only at the signup redirect decision. Like its analytics-stream siblings it carries `visitor_hash`, so the dashboard exclusion list keeps internal/test signups out of the activation count — at this event's volume a single internal signup would visibly skew a day.

https://claude.ai/code/session_01En2uUiiwQoRQKWenCnkPi9


FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: ef5975fc1ec368c390dda18d20923d49c5c7e97b

FagnerMartinsBrack:

None found.

I traced the complete runtime path (anonymous `/view` → `hutch_lastview` cookie → signup → `resolvePostSignupRedirect` → `/queue?url=…&utm_source=signup-autosave` → existing auto-submit save POST) across all three signup paths and found no security or correctness defects:

- **URL is validated twice and can't open-redirect / SSRF.** The cookie only ever holds `validation.url` from `validateSaveableUrl` (`view.page.ts:157,243`), and `resolvePostSignupRedirect` re-validates with the same validator before building the location (`post-signup-redirect.ts:19-22`), so a tampered `hutch_lastview` cannot reach the save. The redirect target is always the internal `/queue`; the save POST re-validates a third time (rejecting non-http(s) and private-network hosts).
- **No double-save.** Autosave only fires when `returnUrl === undefined`. The pending-save `/save` flow always redirects through `?return=…` (`save.page.ts:63`), so `resolvePostSignupRedirect` returns the return URL and skips autosave — the two paths are mutually exclusive.
- **Gating is correct.** The cookie sits behind the same `isbot` gate as `view_opened` and only for `req.userId === undefined` (`view.page.ts:230-244`) — bots, prefetches, and authenticated viewers never set it. OAuth autosave is wired only into genuine new-user branches; existing-user/login and create-race branches keep `parseReturnUrl` (`google-auth.page.ts:169,194,234` vs `217,273`).
- **Cookie hygiene matches `pending-save`.** httpOnly + SameSite=Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout` (`auth.page.ts:500`). Apple tunnels the URL through the HMAC-signed `state` because the cross-site `form_post` drops Lax cookies, and clears the response cookie defensively — identical to how `pendingSaveId` is handled.
- **Backward compatible.** With `lastViewUrl` absent, `resolvePostSignupRedirect` returns exactly `parseReturnUrl({ return: returnUrl })`, preserving prior redirect behavior.

## Medium Priority Issues

None found.

Test coverage is comprehensive: unit tests for the pure resolver (return-wins, fallback, autosave, utm marker, and tampered `not-a-url` / `javascript:` / loopback cases), unit tests for the cookie module (read/consume/clear), plus route tests for the cookie-set gate (anon/authed/bot) and all three signup paths (autosave + clear, explicit-return-wins, no-cookie, tampered-cookie, trial branch, logout clear, Apple state-tunnel). CLAUDE.md conventions are respected — no `process.env`, no `as`, branded `SaveableUrl`, named-param objects, and the diff is Open/Closed-shaped (new modules + delegating call sites).

## Low Priority Suggestions

### 1. The `utm_source=signup-autosave` marker is not 1:1-guaranteed for the funnel
Measurement rides a `utm_source` string on the post-signup `/queue` pageview rather than a discrete event. That string is not unique to this trigger by construction: if any future flow ever reuses `utm_source=signup-autosave`, or a user reloads/shares the post-signup `/queue?url=…&utm_source=signup-autosave` URL, the `autosaves_triggered` query counts it again. The PR already calls this trade-off out. If the metric proves load-bearing, a server-emitted `first_article_autosaved` event at the redirect decision point in `resolvePostSignupRedirect`'s caller would be exactly 1:1 with the trigger and immune to URL reuse (at the cost of one `ANALYTICS_EVENTS` entry + the dashboard widget-count bump). No change requested now — just flagging the fragility.

### 2. A stale pending-save can co-attribute with an autosave
Autosave and the `pending_save_id` conversion tag can co-occur via a stale cookie: an abandoned `/save` leaves `hutch_psid` (1h TTL); if the visitor then reads a *different* article (setting `hutch_lastview`) and signs up via a nav CTA (no `?return=`), the conversion event is tagged `pending_save_id: A` while article **B** is auto-saved. This is pre-existing behavior for `pending_save_id` (this PR doesn't change `emitUserCreated`) and is functionally harmless, but it's worth knowing when joining the autosave cohort to saves in the funnel.

### 3. `readLastViewUrl` defers value validation (consistent, no action needed)
Unlike `readPendingSaveId`, which zod-validates the UUID shape at read time, `readLastViewUrl` (`last-view.ts:19-23`) returns the raw string and leaves validation to `resolvePostSignupRedirect`. This is correct and intentional (the value is a free-form URL, not a fixed shape), and the deferred re-validation fully covers the tamper case — noting it only so the asymmetry with the sibling module reads as deliberate rather than an oversight.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Branch**: feat/first-article-autosave

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: 687a4454f3f26f45fc45fd6cb43844521475aa0f

FagnerMartinsBrack:

None found.

I traced the complete runtime chain at the rebased tip (`687a4454`) — anonymous `/view` → `hutch_lastview` → signup → `resolvePostSignupRedirect` → `/queue?url=…&utm_source=signup-autosave` → the existing auto-submit save — across all three signup paths, and found no security or correctness defect.

- **The save genuinely fires and the marker stays on the pageview.** `/queue?url=X` becomes `filterUrl` → `saveUrl: X` (`queue.page.ts:561,636`); the template renders the save input `value="{{saveUrl}}"` with a `data-auto-submit` attribute (`queue.template.html:20-21`) and `AUTO_SUBMIT_SCRIPT` calls `form.requestSubmit()` (`queue.component.ts:171-189`). The form's action is the fixed `track '/queue/save' source='queue'` link and the POST body is only `{url}`, so `utm_source=signup-autosave` lands on the GET pageview and never on `view_save_intent` — the PR's attribution design holds.
- **URL validated three times; no open-redirect / SSRF / XSS.** `validateSaveableUrl` at set time (`view.page.ts:152,157,243`), re-validated in `resolvePostSignupRedirect` (`post-signup-redirect.ts:19-22`), and again at the save POST (`queue.page.ts:657`). It rejects non-http(s) schemes (`javascript:`, `data:`, `file:`) and private-network/loopback hosts. The redirect target is always the internal `/queue`; `URLSearchParams.toString()` percent-encodes the value so no extra query params can be injected; Handlebars escapes `{{saveUrl}}`.
- **No double-save.** Autosave fires only when `returnUrl === undefined`, and the pending-save `/save` round-trip always carries `?return=`. `returnUrl` is derived from `extractReturnUrl` (`auth.page.ts:244`), so it is a `/`-prefixed path or `undefined` — never `""` — meaning the `!== undefined` guard can't be fooled by an empty value.
- **Gating and session ordering are correct.** The cookie sits behind the same `isbot` gate as `view_opened` and only for `req.userId === undefined` (`view.page.ts:230-244`). Every new-user branch sets the session cookie *before* the autosave redirect, so the follow-on `/queue` load is authenticated (email free+trial `auth.page.ts:303-319,348-364`; Google `google-auth.page.ts:216,272`); existing-user and create-race branches keep `parseReturnUrl` and never autosave.
- **Cookie hygiene mirrors `pending-save` exactly.** httpOnly + SameSite=Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout`; Apple tunnels the URL through the HMAC-signed `state` (Lax cookie is dropped on the cross-site `form_post`) and defensively clears the response cookie — identical to `pendingSaveId`.
- **Backward compatible.** With `lastViewUrl` absent, `resolvePostSignupRedirect` returns exactly `parseReturnUrl({ return: returnUrl })`, preserving prior redirect behavior.

## Medium Priority Issues

None found.

CLAUDE.md conventions are respected: named-param objects (`setLastViewUrl`, `consumeLastViewUrl`, `resolvePostSignupRedirect`), no `process.env`, no `as`, branded `SaveableUrl` used at every boundary, and comments explain *why* (TTL rationale, clear-on-consume rationale). The diff is Open/Closed-shaped — new pure modules plus delegating call sites — and test coverage is comprehensive: the pure resolver (return-wins / fallback / autosave / utm marker / tampered `not-a-url`, `javascript:`, loopback), the cookie module (read / consume / clear), the view-set gate (anon / authed / bot), and all three signup paths (autosave + clear, explicit-return-wins, no-cookie, tampered-cookie, trial branch, logout clear, Apple state-tunnel).

## Low Priority Suggestions

### 1. Tunneling the article URL grows the Apple `hutch_astate` cookie toward the browser size cap
`apple-auth.page.ts` now adds `lastViewUrl` (a full article URL) to the signed `hutch_astate` payload alongside the already-tunneled `returnUrl`, `attribution`, `visitorId`, and `pendingSaveId`. `validateSaveableUrl` puts no length bound on the URL (it accepts any parseable http(s) URL, including 1–3 KB tracking-laden ones via `parsed.toString()`), and the state also carries the HMAC signature. A pathologically long article URL could push the serialized+signed `hutch_astate` toward the ~4 KB per-cookie browser limit; if it's exceeded the browser silently drops the cookie and Apple sign-in fails (nonce/state lost). This is unlikely in practice — the same URL already fit as the standalone `hutch_lastview` cookie, and the design already tunnels the URL-length `returnUrl` — but it's the one genuinely new consideration this PR adds to the state payload, and Apple sign-in was only just released to all users (#933). Consider skipping the `lastViewUrl` tunnel when it would push the payload past a safe threshold (autosave simply degrades to a plain `/queue`, which is the existing behavior). [Fix this →](https://claude.ai/code?q=Guard%20hutch_astate%20signed%20state%20size%20in%20projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fauth%2Fapple-auth.page.ts%20on%20branch%20feat%2Ffirst-article-autosave%20-%20skip%20tunneling%20lastViewUrl%20into%20the%20state%20payload%20when%20it%20would%20push%20the%20signed%20hutch_astate%20cookie%20near%20the%204KB%20browser%20cookie%20limit%20so%20a%20very%20long%20article%20URL%20cannot%20cause%20the%20browser%20to%20drop%20the%20state%20cookie%20and%20break%20Apple%20sign-in&repo=Readplace/readplace.com)

### 2. No single test asserts the end-to-end persistence of the autosave
The route tests assert the post-signup redirect `Location` is `/queue?url=…&utm_source=signup-autosave`, and the auto-submit prefill (`data-auto-submit` + `requestSubmit`) is covered separately in `queue.listing.filters.route.test.ts`. Both halves of the seam are tested, but no test stitches signup → `/queue?url=` → a persisted save row for the new user, because the final save is driven by client-side JS that supertest doesn't execute. This is consistent with the repo's integration-over-E2E strategy and the seam is well-defined, so it's acceptable — noting only that the persistence guarantee rests on the (separately-tested, pre-existing) auto-submit rather than on a single chain assertion.

### 3. Measurement rides a `utm_source` string rather than a discrete event
`autosaves_triggered` counts pageviews carrying `utm_source=signup-autosave`, which is not 1:1 with the trigger by construction: a reload, share, or bookmark of the post-signup `/queue?url=…&utm_source=signup-autosave` URL re-counts (and would idempotently re-save, so no data corruption — just inflated attribution). The PR body already calls this out. If the metric becomes load-bearing, a server-emitted `first_article_autosaved` event at the redirect decision point in `resolvePostSignupRedirect`'s callers would be exactly 1:1 and immune to URL reuse, at the cost of one `ANALYTICS_EVENTS` entry plus the dashboard widget-count bump.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: d7a8928c9f37b3a297ff57b1d44ac2dd51f7904d

FagnerMartinsBrack:

None found.

I traced the complete runtime chain at the squashed tip (`d7a8928c`) — anonymous `/view` → `hutch_lastview` → signup → `resolvePostSignupRedirect` → `/queue?url=…&utm_source=signup-autosave` → the existing auto-submit save — across all three signup paths, and found no security or correctness defect.

- **URL validated three times; no open-redirect / SSRF / XSS.** The cookie only ever holds `validation.url` from `validateSaveableUrl` (`view.page.ts:152,157,243`); `resolvePostSignupRedirect` re-validates with the same validator (`post-signup-redirect.ts:20-21`); and the save POST validates a third time. `validateSaveableUrl` rejects non-http(s) schemes (`javascript:`, `data:`, `file:`) and private-network/loopback hosts, and normalises via `parsed.toString()` (`saveable-url.ts:86-101`). The redirect target is always the internal `/queue`; `URLSearchParams.toString()` percent-encodes the value so no extra params can be injected.
- **No double-save.** Autosave fires only when `returnUrl === undefined`. Every call site derives `returnUrl` from `extractReturnUrl` (`google-auth.page.ts:181`, `apple-auth.page.ts:230`, and the email path), which returns a `/`-prefixed path or `undefined` — never `""` — so the `!== undefined` guard is robust, and the pending-save `/save` round-trip always carries `?return=`, making the two paths mutually exclusive.
- **New-user gating is correct.** Autosave is wired only into genuine new-user branches (Google free `208-225` and trial `275-291`; Apple new-user branches). Existing-user login and create-race branches keep `parseReturnUrl` (Google `193-203`/`233-243`, Apple `225`) and never autosave.
- **Set-gate + session ordering correct.** The cookie is set only for anonymous (`req.userId === undefined`) non-bot views, behind the same `isbot` gate as `view_opened` and after the markdown early-return (`view.page.ts:230-244`), so bots, prefetches, markdown/LLM fetches, and authenticated viewers never set it. Each signup branch sets the session cookie on the same response as the 303, so the follow-on same-site `/queue` GET is authenticated and the save lands in the brand-new user's queue.
- **Cookie hygiene mirrors `pending-save` exactly** — httpOnly + SameSite=Lax + Secure-per-origin + 2h TTL (`last-view.ts` via `baseCookieOptions`), consumed-and-cleared at signup and cleared on `POST /logout`; Apple tunnels the URL through the HMAC-signed `state` (the Lax cookie is dropped on the cross-site `form_post`) and defensively clears the response cookie.
- **No attribution contamination.** `createClickAttributionMiddleware` short-circuits when a `hutch_click` cookie already exists (`click-attribution.middleware.ts:67`) — always true by the time a user reaches the post-signup `/queue?…utm_source=signup-autosave` GET — and attribution is emitted on `emitUserCreated` *before* the autosave redirect. The internal marker can never overwrite acquisition attribution.
- **Backward compatible.** With `lastViewUrl` absent, `resolvePostSignupRedirect` returns exactly `parseReturnUrl({ return: returnUrl })`, preserving prior redirect behaviour.

## Medium Priority Issues

None found.

CLAUDE.md conventions are respected: named-param objects (`setLastViewUrl`, `consumeLastViewUrl`, `resolvePostSignupRedirect`), no `process.env` (`secureCookies` threaded from the composition root at `server.ts:427`), no `as`, branded `SaveableUrl` at every boundary, and comments that explain *why* (TTL rationale, clear-on-consume rationale, cross-site clear rationale). The diff is Open/Closed-shaped — two new pure modules plus delegating call sites — with no `c8 ignore` added. Test coverage is comprehensive: the pure resolver (return-wins / fallback / autosave / utm marker / tampered `not-a-url`, `javascript:`, loopback), the cookie module (read / consume / clear), the view-set gate (anon / authed / bot), and all three signup paths (autosave + clear, explicit-return-wins, no-cookie, tampered-cookie, trial branch, logout clear, Apple state-tunnel).

## Low Priority Suggestions

### 1. Tunneling the article URL grows the Apple `hutch_astate` signed cookie toward the 4 KB browser cap
`apple-auth.page.ts:131` now adds `lastViewUrl` (a full article URL) to the signed `hutch_astate` payload alongside the already-tunneled `returnUrl`, `attribution`, `visitorId`, and `pendingSaveId`, plus the HMAC signature. `validateSaveableUrl` puts no length bound on the URL (it accepts any parseable http(s) URL, including 1–3 KB tracking-laden ones via `parsed.toString()`). A pathologically long article URL could push the serialized+signed state toward the ~4 KB per-cookie limit; if exceeded the browser silently drops the cookie and Apple sign-in fails (nonce/state lost). Low probability — the same URL already fit as the standalone `hutch_lastview` cookie, and the design already tunnels the URL-length `returnUrl` — but it is the one genuinely new field this PR adds to the state payload, and Apple sign-in was only just released to all users (#933). Consider skipping the `lastViewUrl` tunnel when it would push the payload past a safe threshold (autosave then degrades to a plain `/queue`, the existing behaviour). [Fix this →](https://claude.ai/code?q=In%20projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fauth%2Fapple-auth.page.ts%20on%20branch%20feat%2Ffirst-article-autosave%2C%20guard%20the%20size%20of%20the%20signed%20hutch_astate%20cookie%3A%20skip%20tunneling%20lastViewUrl%20into%20the%20state%20payload%20when%20including%20it%20would%20push%20the%20serialized%2Bsigned%20state%20near%20the%20~4KB%20per-cookie%20browser%20limit%2C%20so%20a%20very%20long%20article%20URL%20cannot%20cause%20the%20browser%20to%20drop%20the%20state%20cookie%20and%20break%20Apple%20sign-in.%20Autosave%20should%20degrade%20to%20a%20plain%20%2Fqueue%20in%20that%20case.&repo=Readplace/readplace.com)

### 2. "Last-viewed" is not necessarily "the article that motivated signup"
`setLastViewUrl` overwrites the cookie on *every* anonymous non-bot `/view` (`view.page.ts:242-244`), so if a visitor lands on their motivating article A, clicks through to article B on the same site, then signs up via a nav CTA, it is **B** that auto-saves — not A. The hypothesis frames this as saving "the article that brought them to Readplace," which recency approximates but does not guarantee for multi-article browsing sessions. This is inherent to the last-view design and is a reasonable heuristic (recency is a fine proxy and the queue is non-empty either way) — noting only that the saved article may differ from the acquisition article when the cohort is analysed.

### 3. Measurement rides a `utm_source` string rather than a discrete event
`autosaves_triggered` counts pageviews carrying `utm_source=signup-autosave`, which is not 1:1 with the trigger by construction: a reload, share, or bookmark of the post-signup `/queue?url=…&utm_source=signup-autosave` URL re-counts (and would idempotently re-save, so no data corruption — just inflated attribution). The PR body already calls this out. If the metric becomes load-bearing, a server-emitted `first_article_autosaved` event at the `resolvePostSignupRedirect` decision point would be exactly 1:1 and immune to URL reuse, at the cost of one `ANALYTICS_EVENTS` entry plus the dashboard widget-count bump.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: 1822e502f15aed8a8178ad32a56a6f36b8b07a01

FagnerMartinsBrack:

None found.

Reviewed at the rebased tip (`1822e502`). The only change since the last green review (`d7a8928c`) is a one-line import repoint in `last-view.ts` (`./cookie-options` → `@packages/web-analytics`), forced by main's `@packages/web-analytics` package extraction. I verified it is behavior-preserving and correct, and re-traced the full runtime chain to confirm no regression:

- **The repoint is safe.** The local `./cookie-options` module was deleted by main; no stale imports remain in hutch. `@packages/web-analytics` is a declared workspace dependency (`projects/hutch/package.json:72`) and re-exports `baseCookieOptions` (`index.ts:17`). `baseCookieOptions(secure)` returns `{ httpOnly: true, sameSite: "lax", path: "/", secure }` (`cookie-options.ts:6`) — the `path: "/"` matches every `clearCookie(…, { path: "/" })` in the flow, so consume-once / logout-clear / Apple-state-clear all still line up. The change makes `last-view.ts` identical to its already-shipping sibling `pending-save.ts`, removing drift rather than adding it.
- **URL validated three times; no open-redirect / SSRF / XSS.** Set-time (`view.page.ts:154,159` — only `validation.url` is stored), re-validated in `resolvePostSignupRedirect` (`post-signup-redirect.ts:19-21`), and again at the save POST. `validateSaveableUrl` rejects non-http(s) schemes and private/loopback hosts; redirect target is always the internal `/queue`; `URLSearchParams` percent-encodes the value.
- **No double-save.** Autosave fires only when `returnUrl === undefined`. Every caller derives it from `extractReturnUrl` (`auth.page.ts:242,255`, `google-auth.page.ts:181`, `apple-auth.page.ts:230`), which yields a `/`-prefixed path or `undefined` — never `""` — so the guard can't be fooled, and the pending-save `/save` flow always carries `?return=`.
- **Set-gate + session ordering correct.** Cookie set only after the markdown early-return, behind the same `isbot` gate as `view_opened`, and only for `req.userId === undefined` (`view.page.ts:232-246`); `secureCookies` threaded from the composition root (`server.ts:424,1055`). Each signup branch sets the session cookie on the same response as the 303, so the follow-on `/queue` GET is authenticated.
- **Cookie hygiene.** Consumed-and-cleared at signup, cleared on `POST /logout`, tunneled through the HMAC-signed `hutch_astate` for Apple's cross-site `form_post`, and defensively cleared there too.

## Medium Priority Issues

None found.

CLAUDE.md conventions are respected: named-param objects (`setLastViewUrl`, `consumeLastViewUrl`, `resolvePostSignupRedirect`), no `process.env`, no `as`, branded `SaveableUrl` at every boundary, `assert` for the visitor-id invariant, and why-focused comments (TTL rationale, clear-on-consume rationale, cross-site clear rationale). The diff stays Open/Closed-shaped — two pure modules plus delegating call sites — with no `c8 ignore` added. Test coverage is comprehensive: the pure resolver (return-wins / fallback / autosave / utm marker / tampered `not-a-url`, `javascript:`, loopback), the cookie module (read / consume / clear, non-string cookie), the view-set gate (anon / authed / bot), and all three signup paths (autosave + clear, explicit-return-wins, no-cookie, tampered-cookie, trial branch, logout clear, Apple state-tunnel).

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR body, none introduced or worsened by this commit, none blocking.

### 1. Tunneling the article URL grows the Apple `hutch_astate` signed cookie toward the 4 KB browser cap
`apple-auth.page.ts:131` adds `lastViewUrl` (a full article URL) to the signed `hutch_astate` payload alongside `returnUrl`, `attribution`, `visitorId`, and `pendingSaveId` plus the HMAC signature. `validateSaveableUrl` puts no length bound on the URL, so a pathologically long (tracking-laden) article URL could push the serialized+signed state toward the ~4 KB per-cookie limit; if exceeded the browser silently drops the cookie and Apple sign-in fails. Low probability — the same URL already fits as the standalone `hutch_lastview` cookie, and the design already tunnels the URL-length `returnUrl` — but it's the one genuinely new field in the state payload. Consider skipping the `lastViewUrl` tunnel when including it would push the payload past a safe threshold (autosave then degrades to a plain `/queue`, the existing behavior). [Fix this →](https://claude.ai/code?q=In%20projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fauth%2Fapple-auth.page.ts%20on%20branch%20feat%2Ffirst-article-autosave%2C%20guard%20the%20size%20of%20the%20signed%20hutch_astate%20cookie%3A%20skip%20tunneling%20lastViewUrl%20into%20the%20state%20payload%20when%20including%20it%20would%20push%20the%20serialized%2Bsigned%20state%20near%20the%20~4KB%20per-cookie%20browser%20limit%2C%20so%20a%20very%20long%20article%20URL%20cannot%20cause%20the%20browser%20to%20drop%20the%20state%20cookie%20and%20break%20Apple%20sign-in.%20Autosave%20should%20degrade%20to%20a%20plain%20%2Fqueue%20in%20that%20case.&repo=Readplace/readplace.com)

### 2. "Last-viewed" is not necessarily "the article that motivated signup"
`setLastViewUrl` overwrites the cookie on *every* anonymous non-bot `/view` (`view.page.ts:244-246`), so a visitor who lands on motivating article A, clicks through to article B on the same site, then signs up via a nav CTA will auto-save **B**, not A. Recency is a fine proxy and the queue is non-empty either way — noting only that the saved article may differ from the acquisition article when the cohort is analysed.

### 3. Measurement rides a `utm_source` string rather than a discrete event
`autosaves_triggered` counts pageviews carrying `utm_source=signup-autosave`, which isn't 1:1 with the trigger by construction: a reload, share, or bookmark of the post-signup `/queue?url=…&utm_source=signup-autosave` URL re-counts (and idempotently re-saves, so no data corruption — just inflated attribution). The PR body already calls this out. If the metric becomes load-bearing, a server-emitted `first_article_autosaved` event at the `resolvePostSignupRedirect` decision point would be exactly 1:1 and immune to URL reuse, at the cost of one `ANALYTICS_EVENTS` entry plus the dashboard widget-count bump.

FagnerMartinsBrack:

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: ebee729cb11f195bbea035500f77700051bfa74c

FagnerMartinsBrack:

None found.

Reviewed at commit `ebee729c`. The only delta since the last green review (`1822e502`) is the follow-up work addressing the three standing low-priority notes: an Apple state-cookie size guard (`apple-state.ts`), a discrete `first_article_autosaved` analytics event, and the `resolvePostSignupRedirect` return-shape change that feeds it. I reviewed the new code end to end and found no security or correctness defect.

- **Apple `hutch_astate` size guard is correct and closes note #1.** `signAppleState` (`apple-state.ts:31-45`) tunnels `lastViewUrl` only while the serialized cookie stays within budget, and — crucially — measures the budget against the **`encodeURIComponent`-encoded** signed string (`serializedCookieBytes`), which is exactly what Express writes to the wire; measuring the raw signed string would undercount and still overflow. `MAX_APPLE_STATE_COOKIE_BYTES = 4000` leaves ~96 bytes under the ~4096 per-cookie browser cap for the `hutch_astate=` name. Only the optional, unbounded `lastViewUrl` is dropped under pressure — every load-bearing field (nonce, returnUrl, attribution, visitorId, pendingSaveId) is always retained, so the failure mode is "autosave degrades to a plain `/queue`", never "state cookie dropped → nonce lost → sign-in breaks". The single `signedState` is used for both the cookie and the URL `state` param, so the callback equality check still holds.
- **The discrete event is 1:1 with the trigger and closes note #3.** `resolvePostSignupRedirect` now returns `{ location, autosavedUrl? }` (`post-signup-redirect.ts:24-35`) and `emitFirstArticleAutosaved` (`first-article-autosaved.ts:18-32`) fires off that one decision — a no-op when `autosavedUrl === undefined`, so callers pass it unconditionally. It is wired **only into genuine new-user branches** (`created.userId`) across all three paths (email free+trial, Google free+trial, Apple free+trial); every existing-user and create-race branch keeps `parseReturnUrl` and emits nothing — verified line-by-line in `google-auth.page.ts` (emits at `227`/`299`; `parseReturnUrl` at `179`/`204`/`249`) and `apple-auth.page.ts` (emits at `299`/`375`; `parseReturnUrl` at `231`/`277`/`326`).
- **No new PII surface.** The event logs `article_host` (hostname only), `user_id`, and optional `visitor_id` — never the full article URL — matching the existing `view_opened`/`view_save_intent` posture. Correct stream and logger (`deps.analytics`, `stream: "analytics"`), not the conversions stream.
- **Wiring can't drift.** `analytics` is a real composition-root dependency (`server.ts:328`, `HutchLogger.Typed<AnalyticsEvent>`) threaded into all three routers; hutch's `observability/events.ts` re-exports `ANALYTICS_EVENTS`/`STREAMS` from `@packages/web-analytics`, so the runtime value used to build the event and the literal type it's checked against are the same constant.

## Medium Priority Issues

None found.

CLAUDE.md conventions are respected: named-parameter objects (`signAppleState`, `emitFirstArticleAutosaved`), no `process.env`, no `as`, branded `UserId`, the `analytics` logger is a required (non-defaulted) dependency, and comments explain *why* (the encoded-budget rationale, the drop-only-the-optional-field rationale, the emit-at-decision rationale). The diff stays Open/Closed-shaped — two new pure modules (`apple-state.ts`, `first-article-autosaved.ts`) plus delegating call sites — with no `c8 ignore` added. Test coverage is comprehensive: `apple-state.test.ts` (omit/keep/drop) plus a route test that a >4 KB last-view URL is not tunneled while the nonce survives; the resolver unit test updated to the `{ location, autosavedUrl? }` shape; the discrete event asserted once-on-autosave / absent-on-return-wins across all three signup route suites (with `visitor_id` present for email+Google and omitted for the untunneled Apple state); and the dashboard drift test bumped 28→29 with a new query assertion.

## Low Priority Suggestions

### 1. `first_article_autosaved` counts the *trigger*, not the confirmed persistence
The event fires at the redirect **decision** — before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, user navigates away), the event still counts. This is deliberate and correct for the stated goal: emitting at the decision is precisely what makes the metric immune to the reload/share re-counting that afflicts the `utm_source` marker (emitting on the server-side save POST would reintroduce it). The code comments already document that persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. No change requested — flagging only so the "autosaves per day" widget is read as *triggers*, with the persistence join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can't join to `view_save_intent` by visitor for that subset — it remains joinable by `user_id`. This is inherent to the cross-site flow and identical to the existing limitation on the `user_created` conversion event; the new route test explicitly asserts the omission. Noting only for the cohort join, not as a defect.

### 3. (Standing, now resolved-by-design) "last-viewed" ≠ "motivating article"
Carried from prior reviews: `setLastViewUrl` overwrites on every anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. The author kept recency by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Branch**: feat/first-article-autosave

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: 6410397b8252078d58ce32ac3c8ee5bab13d446b

FagnerMartinsBrack:

None found.

Reviewed at commit `6410397b` — a **conflict-resolution rebase** onto latest `main`. The only delta since the last green review (`ebee729c`) is the merge with **#918 (`signup_attempted`)**, which landed on `main` and touched the same analytics/auth surfaces. I focused on what the merge could have broken and verified every seam, then re-confirmed the feature is unchanged and sound.

- **The merge type-checks.** `pnpm nx run hutch:compile` succeeds across hutch + 30 dependent tasks (including `@packages/web-analytics`), proving #918's analytics surface and this PR's `FirstArticleAutosavedEvent` coexist cleanly — the `AnalyticsEvent` union resolves, no duplicate declarations, all imports link. Exactly one `import type { AnalyticsEvent }` remains in each of `auth.page.ts` / `google-auth.page.ts` / `apple-auth.page.ts`, so the `TS2300` the conflict-fixer flagged is genuinely resolved.
- **Wiring survived the rebase.** `analytics: deps.analytics` is threaded into all three routers (`server.ts:894` auth, `929` Google, `956` Apple), and the view router receives `secureCookies` (`server.ts:1064`). Both `deps.analytics` (autosave + `signup_attempted`) and `deps.conversionLogger` (`emitUserCreated`) stay distinct.
- **Dashboard reconciliation is arithmetically correct.** 30 `logWidget` + 1 `metricWidget` = **31** widgets; the drift test moves `30 → 31` (main's base already carried #918's +2 signup-form widgets). The new "First-article autosaves per day" widget sits at `y:138`, immediately after the `y:130` signup-form row — no grid overlap.
- **Both test suites intact after the interleave.** `auth.route.test.ts` keeps both the `POST /signup — signup_attempted analytics` (L895) and `POST /signup — first-article autosave` (L1094) describe blocks; the autosave event is asserted across all three signup route suites plus the resolver unit test.
- **Feature re-verified end to end.** Set-gate is behind the markdown early-return + `!isbot` + `req.userId === undefined` (`view.page.ts:228-246`), saving the already-`validateSaveableUrl`-normalised `articleUrl`. URL is validated three times (set / `resolvePostSignupRedirect` / save POST) — no open-redirect / SSRF / XSS; redirect target is always internal `/queue` and `URLSearchParams` percent-encodes the value. No double-save (autosave fires only when `returnUrl === undefined`; the `/save` round-trip always carries `?return=`). Apple's `signAppleState` drops only the optional `lastViewUrl` when the `encodeURIComponent`-measured signed cookie would exceed 4000 bytes, so an overflow degrades autosave to a plain `/queue` rather than dropping the nonce. Cookie is httpOnly + Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout`.

## Medium Priority Issues

None found.

CLAUDE.md conventions hold across the merge: named-parameter objects (`signAppleState`, `emitFirstArticleAutosaved`, `resolvePostSignupRedirect`), no `process.env`, no `as`, branded `UserId`, `analytics` a required (non-defaulted) dependency, `assert` for the visitor-id invariant, and why-focused comments (encoded-budget rationale, clear-on-consume rationale, cross-site clear rationale). The diff stays Open/Closed-shaped — pure new modules plus delegating call sites — with no `c8 ignore` added. `hutch:compile` is green; the conflict-fixer's full `pnpm check` (functions 100%, branches 96.98%) ran green on this exact tip.

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this rebase, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct for the goal — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission.

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Branch**: feat/first-article-autosave

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: 264ca7f9b271eeead3a1432fc41997f53e30acfa

FagnerMartinsBrack:

None found.

Reviewed at commit `264ca7f9` — a **conflict-resolution rebase** onto latest `main`. The delta since the last green review (`6410397b`) is the merge with **#935 (checkout-funnel instrumentation)** and main's **multi-account OAuth** (`53c0356a`), both touching this PR's analytics/auth surfaces. I focused on what the merge could have broken, then re-confirmed the feature end to end. No security or correctness defect.

- **The merge left no residue.** Zero conflict markers in any changed file, and exactly one `import type { AnalyticsEvent }` remains in each of `auth.page.ts` / `google-auth.page.ts` / `apple-auth.page.ts` — the `TS2300` duplicate-import risk the conflict-fixer called out is genuinely resolved.
- **Dashboard widget count reconciles arithmetically.** `logWidget` is defined in-file (`analytics-dashboard.ts:54`), so of the 34 `logWidget(` matches, 33 are calls; plus one raw `type:"metric"` widget pushed at `:250` (Imports-completed lifetime) = **34** widgets, matching the drift assertion `toHaveLength(34)`. The `31 → 34` move since the prior review is exactly #935's three checkout widgets (`+3`) with this PR's single autosave widget already accounted for. The new "First-article autosaves per day" widget sits at `x:12, y:146, w:12, h:8` — beside "Conversions per day" (`x:0, y:146`), completing a clean 2×2 block below #935's `y:138` row with **no grid overlap and no gap**.
- **Multi-account OAuth is orthogonal to autosave.** Main's `prompt=select_account` forward lives only inside `GET /auth/google` (`google-auth.page.ts:119-121`) and never touches the callback's new-user gating, so autosave still fires only on genuine account creation (`created.userId` branches), never on existing-user login or the create-race recovery branches.
- **URL validated three times; no open-redirect / SSRF / XSS.** Set-time (`view.page.ts:245`, only the already-`validateSaveableUrl`-normalised `articleUrl` is stored), re-validated in `resolvePostSignupRedirect` (`post-signup-redirect.ts:31-32`), and again at the save POST. Redirect target is always the internal `/queue`; `URLSearchParams.toString()` percent-encodes the value so no extra params can be injected.
- **No double-save.** Autosave fires only when `returnUrl === undefined`. Every caller derives `returnUrl`/`safeReturnUrl` from `extractReturnUrl` (a `/`-prefixed path or `undefined`, never `""`), and the pending-save `/save` round-trip always carries `?return=`, making the two paths mutually exclusive.
- **Set-gate + wiring correct.** Cookie set only after the markdown early-return, behind the same `isbot` gate as `view_opened`, and only for `req.userId === undefined` (`view.page.ts:232-246`). `analytics` is threaded into all three routers and `secureCookies` into the view router (`server.ts:940,967,1076`).
- **Cookie hygiene + Apple size-guard.** httpOnly + Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout` (`auth.page.ts:633`). `signAppleState` (`apple-state.ts`) tunnels `lastViewUrl` only while the **`encodeURIComponent`-encoded** signed state stays ≤ 4000 bytes and drops **only** that optional field under pressure — every load-bearing field (nonce, returnUrl, attribution, visitorId, pendingSaveId) is always retained, so an overflow degrades autosave to a plain `/queue` rather than dropping the nonce and breaking sign-in.

## Medium Priority Issues

None found.

CLAUDE.md conventions hold across the merge: named-parameter objects (`signAppleState`, `emitFirstArticleAutosaved`, `resolvePostSignupRedirect`, `setLastViewUrl`), no `process.env`, no `as`, branded `UserId`, `analytics` a required (non-defaulted) dependency, `assert` for the visitor-id invariant (`view.page.ts:233`), and why-focused comments (encoded-budget rationale, clear-on-consume rationale, cross-site clear rationale). The diff stays Open/Closed-shaped — pure new modules (`apple-state.ts`, `first-article-autosaved.ts`, `last-view.ts`, `post-signup-redirect.ts`) plus delegating call sites — with no `c8 ignore` added. Test coverage is real and comprehensive: `apple-state.test.ts` (omit/tunnel/drop, plus load-bearing fields survive the drop); the discrete `first_article_autosaved` event asserted `toHaveLength(1)` on autosave and `toHaveLength(0)` on return-wins across all three signup route suites, with `visitor_id` present for email+Google and asserted absent (`not.toHaveProperty`) for untunneled Apple state; the drift test bumped to 34 with a dedicated query assertion for the new widget.

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this rebase, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct for the goal — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission.

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: 392681ab2d9d369b152d38b66da1c332ff5bf5ec

FagnerMartinsBrack:

None found.

Reviewed at commit `392681ab` — a **merge of latest `main`** into the branch. The only delta since the last green review (`264ca7f9`) is main's Chrome Web Store listing rename, which is entirely orthogonal to this PR. I confirmed the merge is clean, that it compiles, and re-verified the feature end to end. No security or correctness defect.

- **The feature code is byte-identical to the last green review.** `git diff 264ca7f9 HEAD` touches only **5 files**, all from main: `chrome-extension/STORE_LISTING.md`, `home.component.ts`, `install.component.ts`, `install.route.test.ts`, and `supported-clients.ts` — none of which touch the analytics, auth, or view surfaces this PR changes. The net PR diff vs `origin/main` remains exactly **+768 / −19** across the same 21 files, so the merge added no drift.
- **The merge left no residue and compiles.** Zero conflict markers anywhere in the tree; exactly one `import type { AnalyticsEvent }` in each of `auth.page.ts` / `google-auth.page.ts` / `apple-auth.page.ts`. `pnpm nx run hutch:compile` succeeds across hutch + 30 dependent tasks, proving main's advancement and this PR's `FirstArticleAutosavedEvent` union/imports resolve together with no duplicate declarations.
- **URL validated three times; no open-redirect / SSRF / XSS.** Set-time (`view.page.ts:245` stores only the already-`validateSaveableUrl`-normalised `articleUrl`), re-validated in `resolvePostSignupRedirect` (`post-signup-redirect.ts:31-32`), and again at the save POST. Redirect target is always the internal `/queue`; `URLSearchParams.toString()` percent-encodes the value so no extra params can be injected.
- **No double-save.** Autosave fires only when `returnUrl === undefined`. Every caller derives `returnUrl`/`safeReturnUrl` from `extractReturnUrl` (a `/`-prefixed path or `undefined`, never `""`), and the pending-save `/save` round-trip always carries `?return=`, making the two paths mutually exclusive.
- **Set-gate + new-user gating correct.** The cookie is set only after the markdown early-return, behind the same `isbot` gate as `view_opened`, and only for `req.userId === undefined` (`view.page.ts:232-246`). Autosave and `emitFirstArticleAutosaved` are wired only into genuine new-user branches (`created.userId`) across all three paths (email `auth.page.ts:378-382`/`441-445`, Google `230-234`/`302-306`, Apple `303-306`/`379-382`); every existing-user and create-race branch keeps `parseReturnUrl` and emits nothing.
- **Apple state cookie/param can't diverge.** The single `signedState` (`apple-auth.page.ts:132`) feeds both the `hutch_astate` cookie (`:145`) and the URL `state` param (`:156`), so the callback equality check holds regardless of the size-guard outcome. `signAppleState` tunnels `lastViewUrl` only while the `encodeURIComponent`-measured signed state stays ≤ 4000 bytes and drops **only** that optional field under pressure — every load-bearing field (nonce, returnUrl, attribution, visitorId, pendingSaveId) is always retained, so an overflow degrades autosave to a plain `/queue` rather than dropping the nonce and breaking sign-in.
- **Cookie hygiene + PII.** httpOnly + SameSite=Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout`; Apple reads (not consumes) at set time so an abandoned flow keeps the cookie, and tunnels the URL through the HMAC-signed state for the cross-site `form_post`. The event logs `article_host` (hostname only via `articleHostFrom`), never the full URL.

## Medium Priority Issues

None found.

CLAUDE.md conventions hold across the merge: named-parameter objects (`signAppleState`, `emitFirstArticleAutosaved`, `resolvePostSignupRedirect`, `setLastViewUrl`), no `process.env`, no `as`, branded `UserId`, `analytics` a required (non-defaulted) dependency, `assert` for the visitor-id invariant (`view.page.ts:233`), and why-focused comments (encoded-budget rationale, clear-on-consume rationale, cross-site clear rationale). The diff stays Open/Closed-shaped — pure new modules (`apple-state.ts`, `first-article-autosaved.ts`, `last-view.ts`, `post-signup-redirect.ts`) plus delegating call sites — with no `c8 ignore` added. Test coverage is unchanged from the last green review and comprehensive: `apple-state.test.ts` (omit/tunnel/drop, plus load-bearing fields survive the drop); the discrete `first_article_autosaved` event asserted `toHaveLength(1)` on autosave and `toHaveLength(0)` on return-wins across all three signup route suites, with `visitor_id` present for email+Google and asserted absent for untunneled Apple state; the dashboard drift test at 34 with a dedicated query assertion for the new widget.

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this merge, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct for the goal — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission.

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Branch**: feat/first-article-autosave

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Branch**: feat/first-article-autosave

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: cd9883bce46533ce1e5ec27e6a08548c7ceca903

FagnerMartinsBrack:

None found.

Reviewed at commit `cd9883bc` — a **conflict-resolution rebase** onto latest `main` (`3b6f64d8`). The only delta since the last green review (`392681ab`) is the merge with **#937** (`count user-facing save failures`), which landed a "Save errors by surface" save-funnel widget into the same dashboard grid region this PR's autosave widget targeted. I focused on what that merge could have broken — the grid layout and the widget-count drift guard — then re-confirmed the feature end to end. No security or correctness defect.

- **The dashboard grid is collision-free.** I mapped every widget's `x/y/width/height`. #937's "Save errors by surface" occupies `x:12, y:146` (`analytics-dashboard.ts:485`); the PR's "Conversions per day" sits at `x:0, y:146`, filling that row exactly; the autosave widget was relocated to `x:0, y:154` — its own row, no overlap and no gap. This isn't just eyeballed: the `widget positions do not overlap` test (`analytics-dashboard.test.ts:290-302`) is an O(n²) pairwise overlap check, and the `respects the CloudWatch 24-column grid` test (`:304-309`) bounds every widget's right edge — both ran green in the conflict-fixer's full `pnpm check`, so a silent grid collision would have failed CI.
- **The widget-count reconciliation is arithmetically correct.** 34 `logWidget()` calls + 1 raw `type:"metric"` widget (`:250`) = **35**, matching `toHaveLength(35)`. The `31 → 35` move since the last review is exactly #937's `+4` save-funnel breakdown, with this PR's single first-article-autosave widget already accounted for; the per-section tally in the test description sums to 35.
- **No merge residue.** Zero conflict markers in any changed file, and the net diff is exactly **+768 / −19** across the same 21 files GitHub reports — the rebase added no drift. Exactly one `import type { AnalyticsEvent }` in each of the three auth pages.
- **URL validated three times; no open-redirect / SSRF / XSS.** Set-time (`view.page.ts:248`, only the already-`validateSaveableUrl`-normalised `articleUrl` is stored), re-validated in `resolvePostSignupRedirect` (`post-signup-redirect.ts:31-32`), and again at the save POST. Redirect target is always the internal `/queue`; `URLSearchParams.toString()` percent-encodes the value so no extra params can be injected.
- **No double-save.** Autosave fires only when `returnUrl === undefined` (`post-signup-redirect.ts:29`); every caller derives `returnUrl`/`safeReturnUrl` from `extractReturnUrl` (a `/`-prefixed path or `undefined`, never `""`), and the pending-save `/save` round-trip always carries `?return=`, making the two paths mutually exclusive.
- **Set-gate + new-user gating correct.** The cookie is set only after the markdown early-return, behind the same `isbot` gate as `view_opened`, and only for `req.userId === undefined` (`view.page.ts:235-249`). Autosave + `emitFirstArticleAutosaved` are wired only into genuine new-user branches (`created.userId`) across email (`auth.page.ts:369-375`/`412-418`), Google (`226-232`/`281-287`), and Apple (`299-304`/`358-363`); every existing-user and create-race branch keeps `parseReturnUrl` and emits nothing (Google `179`/`204`/`249`, Apple `231`/`277`/`326`).
- **Wiring survived the auto-merge.** `analytics` is threaded into all three auth routers and the view router (`server.ts:918,953,979,1108`), and `secureCookies` into the view router (`server.ts:1090`); `deps.analytics` (autosave) stays distinct from `deps.conversionLogger` (`emitUserCreated`).
- **Cookie hygiene + Apple size-guard.** httpOnly + SameSite=Lax + Secure-per-origin + 2h TTL, consumed-and-cleared at signup and on `POST /logout` (`auth.page.ts:604`, `path:"/"` matching the set path). `signAppleState` tunnels `lastViewUrl` only while the `encodeURIComponent`-measured signed state stays ≤ 4000 bytes and drops **only** that optional field under pressure — every load-bearing field (nonce, returnUrl, attribution, visitorId, pendingSaveId) is always retained, so an overflow degrades autosave to a plain `/queue` rather than dropping the nonce and breaking sign-in. The single `signedState` feeds both the cookie and the URL `state` param, so the callback equality check holds. The event logs `article_host` (hostname only), never the full URL.

## Medium Priority Issues

None found.

CLAUDE.md conventions hold across the merge: named-parameter objects (`signAppleState`, `emitFirstArticleAutosaved`, `resolvePostSignupRedirect`, `setLastViewUrl`), no `process.env`, no `as`, branded `UserId`, `analytics` a required (non-defaulted) dependency, `assert` for the visitor-id invariant (`view.page.ts:236`), and why-focused comments (encoded-budget rationale, clear-on-consume rationale, cross-site clear rationale). The diff stays Open/Closed-shaped — pure new modules (`apple-state.ts`, `first-article-autosaved.ts`, `last-view.ts`, `post-signup-redirect.ts`) plus delegating call sites — with no `c8 ignore` added. `web-analytics` cleanly exposes the new surface: `firstArticleAutosaved` in `ANALYTICS_EVENTS` (`events.ts:29`), `FirstArticleAutosavedEvent` in the `AnalyticsEvent` union (`analytics.ts:279-299`), both re-exported (`index.ts:53`). Test coverage is real and comprehensive: `apple-state.test.ts` (omit / tunnel / drop, with load-bearing fields asserted to survive the drop); the discrete `first_article_autosaved` event asserted `toHaveLength(1)` on autosave and `toHaveLength(0)` on return-wins across all three signup route suites, with `visitor_id` present for email+Google and asserted absent for untunneled Apple state; the dashboard drift test at 35 with a dedicated query assertion for the new widget.

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this rebase, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct for the goal — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "First-article autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission.

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: a9525dc869d0edc1013c26cd80fe62eb5cdd863e

FagnerMartinsBrack:

None found.

Reviewed at commit `a9525dc8` — the only delta since the last green review (`cd9883bc`) is a single fix commit that closes two real gaps against the PR's *own* stated intent: excluding speculative prefetches and internal test visitors from first-article autosave. I reviewed the new code end to end, type-checked the DI wiring, and found no security or correctness defect.

- **The prefetch gate is correct and closes a genuine bug.** `view.page.ts:247` now sets `hutch_lastview` only when `req.userId === undefined && !isPrefetchOrBot(req)` — the same `Sec-Purpose`/`Purpose: prefetch` predicate (`:102-106`) that already guards the paid crawl at `:176`. Before this commit, a speculative browser prefetch carried a normal user-agent, passed the bare `!isbot` gate, and stamped the cookie with an article the reader never opened — so a signup within the 2h TTL would auto-save the *wrong* article. This is precisely what the PR description promised ("behind the same `isbot` gate … **plus** the `Sec-Purpose`/`Purpose: prefetch` check") but the code had not yet enforced. `view_opened` intentionally still fires for prefetches (unchanged, pre-existing baseline); only the autosave-claiming cookie gained the guard.
- **`visitor_hash` on the event closes the internal-visitor exclusion gap.** `first_article_autosaved` was the only `analytics`-stream event lacking `visitor_hash`, so the dashboard's internal-visitor exclusion could not apply to the activation widget — and at this event's volume a single internal test signup would visibly skew a day. The event now carries `hashIp({ ip: req.ip, salt: deps.salt })` (typed `string | null`, identical to every sibling — `analytics.ts:287`), the widget applies `...exclude` (`analytics-dashboard.ts:736`), and `salt` is threaded from the composition root into the Google and Apple routers (`server.ts:954,981`; the email router already carried it via #918). The now-false *"No exclude clause: the event carries no visitor_hash"* comment was correctly deleted, not left in place.
- **Wiring is type-verified, not just eyeballed.** `pnpm nx run hutch:compile` succeeds across hutch + 30 dependent tasks, so the two new fields — `salt` on `GoogleAuthDependencies`/`AppleAuthDependencies` and `ip` on the `emitFirstArticleAutosaved` params — resolve at all six call sites. `hashIp` is re-exported from `@packages/web-analytics` (`index.ts:36`). `excludeVisitorHashesClause` returns `[]` for an empty exclusion list (`analytics-dashboard.ts:49`), so `...exclude` is a safe no-op and the query stays well-formed; the widget count is unchanged at 35.
- **The feature itself is unchanged and still sound.** URL still validated three times (set / `resolvePostSignupRedirect` / save POST); no double-save (autosave only when `returnUrl === undefined`); autosave + emit wired only into genuine new-user branches; Apple `signAppleState` still drops only the optional `lastViewUrl` under the encoded 4000-byte budget; cookie consumed-and-cleared at signup and on `POST /logout`. The event logs `article_host` only — `visitor_hash` is a salted SHA-256 truncated to 16 chars, so no raw IP or full-URL PII is introduced.

## Medium Priority Issues

None found.

CLAUDE.md conventions hold: named-parameter objects (`hashIp({ ip, salt })`, `emitFirstArticleAutosaved(deps, params)`), no `process.env`, no `as`, branded `UserId`, `salt`/`analytics` required (non-defaulted) dependencies, and the `apple-state.test.ts` `if/throw` was replaced with `assert` from `node:assert/strict` per the test-driven-design skill. Test coverage is comprehensive and positive-asserting: a new `Sec-Purpose: prefetch` test pins that the cookie is not set (`view.route.test.ts:793`); the trial (founding-exhausted) autosave path is now pinned end-to-end for both Google and Apple; return-wins now asserts *zero* `first_article_autosaved` events on both OAuth paths; and the `visitor_hash` field is asserted on the email/Google/Apple autosave events. `expect.any(String)` is the right matcher here — the sibling `view_opened` test asserts `typeof visitor_hash === "string"` from the identical `hashIp({ ip: req.ip, salt })` call, confirming `req.ip` is populated in these supertest harnesses. The dashboard drift test remains at 35 with the exclusion clause now asserted on the autosave widget. No `c8 ignore` added; the diff stays Open/Closed-shaped.

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this commit, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "First-article autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission. (The new `visitor_hash` does *not* fill this gap — it is IP-derived, for the exclusion list, not a per-visitor join key.)

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every eligible anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). No action.

FagnerMartinsBrack:

## Context
- **PR**: #943
- **Commit**: d10c0b1d21f2b7602166c22e8d9150f5e2a93c16

FagnerMartinsBrack:

None found.

Reviewed at commit `d10c0b1d`. The delta since the last green review (`a9525dc8`) is two new PR commits plus a clean merge of `origin/main`. Both commits fix a subtle regression that a *clean text merge* had silently introduced against this PR's own intent, and each pins the fix with a test the merge couldn't have caught. I verified each end to end.

- **`11ca81f7` — repoint the autosave widget at the never-expire analytics group is correct.** After main's `3ea08b6f` migrated every analytics dashboard widget from the 30-day hutch source group to the never-expire destination group, this PR's `first_article_autosaved` widget still read `[hutchLogGroupName]`. I confirmed the events actually live in the destination group: the event is emitted on `stream: STREAMS.analytics` (`first-article-autosaved.ts:23`), `FORWARDED_STREAMS` contains `STREAMS.analytics` (`events.ts:76`), and the forwarder subscription filter is `{ $.stream = "analytics" || … }` (`infra/index.ts:1369`), which copies the line into `ANALYTICS_LOG_GROUP` (`forward-analytics-handler.ts`). The fix (`logGroupNames: analyticsSource`, `analytics-dashboard.ts:766`) aligns the widget with every sibling analytics-stream widget; the old group would have lost activation history past 30 days and scanned unforwarded source volume. No origin clause is needed because `first_article_autosaved` is a hutch-unique event name (the `signup_attempted` precedent), so `event = "…"` already scopes it in the shared destination group. The new assertion (`analytics-dashboard.test.ts:88`) pins the `SOURCE '<analytics-group>' |` prefix so this can't regress silently again.
- **`d10c0b1d` — the canonical-identity test pins real behavior.** Main's redirect-dedupe work inserted `resolveCanonicalIdentity` (`view.page.ts:168`) above the point where the cookie reads `articleUrl`, so `hutch_lastview` silently shifted from "the URL the reader opened" to "the canonical article it folds onto." That is the intended behavior (autosave should land on the same deduped article the reader was shown), but it arrived via variable shadowing and no test could observe it — the harness hardwired `resolveCanonicalIdentity` to identity, so both meanings passed. The commit threads `resolveCanonicalIdentity` as a `createTestApp`/`useTestServer` override (`test-app.ts:272,325`), spread *after* the identity default (`:294` over `:179`) exactly like the existing `getChangelogBanner`/`getSessionUserId` seams, and adds a view route test that mounts a non-identity fold and asserts the canonical URL is what lands in the cookie (`view.route.test.ts:816`). The override is properly wired, not just added to the signature.
- **The merge is clean and the feature is intact.** Net diff vs `origin/main` is exactly **22 files**; zero conflict markers anywhere in the tree; exactly one `import type { AnalyticsEvent }` in each of the three auth pages (no `TS2300`). `pnpm nx run hutch:compile` is green across hutch + 30 tasks, so main's advancement and this PR's `FirstArticleAutosavedEvent` surface resolve together. URL still validated three times (set / `resolvePostSignupRedirect` / save POST); autosave still fires only on genuine new-user branches with `returnUrl === undefined`; the set-gate is still behind the markdown early-return + `!isbot` + `!isPrefetchOrBot` + `req.userId === undefined` (`view.page.ts:260`); Apple's `signAppleState` still drops only the optional `lastViewUrl` under the encoded 4000-byte budget; the event still logs `article_host` + IP-derived `visitor_hash` only (no full-URL PII).

## Medium Priority Issues

None found.

CLAUDE.md conventions hold across the delta: the new test override is a named-parameter object, no `process.env`, no `as`, `assert` from `node:assert/strict` in the new test, and the widget change is a one-value repoint with a matching drift assertion (no `c8 ignore`, no coverage/lint excludes). The diff stays Open/Closed-shaped. Verification run locally on this commit: `analytics-dashboard.test.js` **28 passed** (covering the repoint, the new SOURCE assertion, and the O(n²) grid-overlap + 24-column-bound invariants), `view.route.test.js` **142 passed** (covering the new canonical-identity test and the existing anon/authed/bot/prefetch set-gate cases); CI is already green on this exact commit (run `29641000985`).

## Low Priority Suggestions

These are standing notes carried from prior reviews — all previously acknowledged in the PR thread, none introduced or worsened by this delta, none blocking.

### 1. `first_article_autosaved` counts the *trigger*, not confirmed persistence
The event fires at the redirect **decision**, before the client-side auto-submit on `/queue?url=…` actually POSTs the save. If that JS never runs (JS disabled, immediate navigation away) the event still counts. This is deliberate and correct for the goal — emitting at the decision is exactly what makes it immune to the reload/share re-counting that afflicts the `utm_source` marker. Persistence is confirmed by joining to the follow-on `view_save_intent` (`outcome = "saved"`) on `visitor_id`. Read the "First-article autosaves per day" widget as *triggers*, with the join as the confirming metric.

### 2. Apple autosave events omit `visitor_id` when it wasn't tunneled into the signed state
For Apple, `visitorId` comes from `stateData.visitorId` (the cross-site `form_post` carries no first-party cookies), so when it wasn't tunneled the event has no `visitor_id` and can only join to `view_save_intent` by `user_id` for that subset. Inherent to the cross-site flow and identical to the existing `user_created` limitation; the route test explicitly asserts the omission. (The `visitor_hash` added earlier does not fill this gap — it is IP-derived, for the exclusion list, not a per-visitor join key.)

### 3. (Standing, resolved-by-design) "last-viewed" ≠ "motivating article"
`setLastViewUrl` overwrites on every eligible anonymous `/view`, so a multi-article session auto-saves the *last* article read, not the entry article. Kept by design (the article being read at decision time is the strongest intent proxy, and `/view` only ever resolves a real saved article, so the queue is non-empty either way). This commit's canonical-identity test refines it slightly: what lands is the *canonical* form of that last article, which is exactly what the reader was shown. No action.

PR #958, opened 2026-07-11
fix(ios): return only the cookies a bootstrap response set

FagnerMartinsBrack:
## Summary

`ReadplaceAPI.sessionCookies` had two paths that disagreed with each other and with the doc comment:

- **Store path** — `httpCookieStorage.cookies(for: url)` returns **every** cookie in the (ephemeral) jar for the host, including any an earlier request left behind.
- **Header fallback** — `HTTPCookie.cookies(withResponseHeaderFields:)` returns **only this response's** cookies.
- **Doc comment** — promised "every cookie the response set".

So on the store path the method could hand back a cookie the mint response never set. That also quietly defeated `bootstrapSession`'s `guard !cookies.isEmpty` check ("a response that sets no cookie is a failed mint"): a mint that set nothing still looked successful whenever a stale cookie sat in the jar.

## Change

- `bootstrapSession` snapshots the jar (as a `Set<CookieIdentity>` keyed on name + domain + path + value) **before** the request.
- `sessionCookies` reads the post-request jar and returns only the cookies **absent from that snapshot** — i.e. exactly what this response set. Reading the already-parsed cookies back from the store still sidesteps the folded-`Set-Cookie`-header hazard the original code was designed around; the header fallback is untouched.
- Both paths and the doc comment now agree: only what the response set is returned, and the failed-mint guard is honest even with a populated jar.

This reconciles the store path with the header fallback and the doc comment — no client-facing behaviour change for the normal single-`hutch_sid` mint.

## Tests

Two regression tests in `ReadplaceAPITests` (both fail against the old code, pass now):
- `testBootstrapSessionReturnsOnlyTheCookiesThisResponseSet` — a stale jar cookie is excluded; only the freshly-set `hutch_sid` is returned.
- `testBootstrapSessionTreatsAResponseThatSetsNoNewCookieAsAFailedMint` — a mint that sets no new cookie throws `.decoding` even when the jar holds a stale cookie (this test also now exercises the previously-uncovered header-fallback branch).

## Incidental (separate commit)

The shared pre-commit hook (`pnpm check`) was blocked by a **pre-existing, unrelated** failure in `hutch` — `auth.route.test.ts:831` asserted the signup form action as `/signup?return=…` while the page renders the UTM params ahead of `return` (as the sibling assertion at line 357 in the same file already expects). Introduced by the latest commit on `main` (#948). One-line fix to align the stale assertion with the actual action; the POST-replay it guards is unaffected. Kept as its own commit so it's easy to review/revert independently.

Note: the iOS Swift changes are validated by the separate macOS `ios-tests` CI job (they aren't covered by `pnpm check` — `ios-readplace` has no `check` target); no Swift toolchain is available in this session to run them locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HrdRrS9wrTAfjK4s59Bvcp)_


FagnerMartinsBrack:

## Context
- **PR**: #958
- **Branch**: claude/readplace-sessioncookies-reconcile-08tkw6

FagnerMartinsBrack:

## Context
- **PR**: #958
- **Commit**: 02f48237985114143f8ffb4a1dbd27d3a37b13dc

FagnerMartinsBrack:
This is a tight, well-reasoned fix. The before/after jar snapshot keyed on a `CookieIdentity(name, domain, path, value)` correctly reconciles the store path with the header fallback and the doc comment. I traced every branch and both new tests, and I'm confident in the change.

**What I verified**
- **Fresh-filter correctness** — `sessionCookies` returns early only when it finds a cookie *absent from the pre-request snapshot*, so a stale jar cookie is never handed back. A re-issued cookie (same name, **new** value) still reads as fresh because `value` is part of the key.
- **Identical-value re-set is still handled** — if the response re-sets a cookie with the *same* value already in the jar, `fresh` is empty and control falls through to the `Set-Cookie` header parse, which returns it.
- **The two tests are genuine regressions** — the old `sessionCookies` returns `["leftover","hutch_sid"]` for test 1 (fails `["hutch_sid"]`) and returns the stale cookie instead of throwing for test 2 (fails). Both pass now, and the fail-on-old-code behaviour confirms the session-config cookie storage is shared with the seeded `config`.
- **`Hashable` synthesis** — `CookieIdentity` still gets synthesized `Hashable` despite the custom `init(_:)` (all stored props are `String`).
- **Conventions** — no `as!`/`try!`/force-unwrap; `throw` is right for the recoverable failed-mint; `CookieIdentity` is only ever built from an `HTTPCookie` (no positional-construction hazard); comments are why-not-what.
- **Scope** — diff is iOS-only; the incidental `auth.route.test.ts` commit was correctly dropped during conflict resolution.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Deliberate behaviour change on re-mint — worth a conscious nod to the server contract
On a repeat `bootstrapSession` where the server returns a 2xx with **no** `Set-Cookie` while a still-valid `hutch_sid` already sits in a non-ephemeral jar, the old code returned that stale cookie (success); the new code throws `.decoding`. This is the *intended* hardening and is safe under the contract that a successful mint always sets a cookie. Production uses `.ephemeral` config, so the jar is empty per launch and the common path is unaffected.

### 2. The `Set-Cookie` header fallback is now reachable with a non-empty (all-stale) jar
Because the store path returns early only on a *fresh* cookie, a jar holding only stale cookies now falls through to `HTTPCookie.cookies(withResponseHeaderFields:)`. For the single-`hutch_sid` mint this is correct. The folded-multi-`Set-Cookie` hazard would only resurface if a response re-set ≥2 cookies whose values all match the jar — not a shape this endpoint produces, and not a regression. Purely informational.

### 3. Swift coverage is validated only by the macOS `ios-tests` job
The new `?.`/`??`/`guard` branches and the two tests weren't run in this review environment. Worth confirming that job is green before merge, since the repo enforces an iOS coverage gate (#957) — the `?? []` in the `priorCookies` snapshot and the `guard let headers … else { return [] }` fallback line are the branches most likely to want an exercising test if the gate is strict.

FagnerMartinsBrack:

## Context
- **PR**: #958
- **Commit**: 6cce4fc128f68bf591ae270a97baa7b2f5989b58

FagnerMartinsBrack:
Re-review at merge commit `6cce4fc1`. The merge only pulls `origin/main` in; the PR payload vs `origin/main` is unchanged — exactly the two iOS files (+81/−8), and `auth.route.test.ts` is byte-identical to main (the incidental commit correctly stayed dropped). The `sessionCookies` snapshot/diff fix is sound. This round I additionally traced the change against the **actual** iOS coverage gate, which closes the one open item from the prior reviews.

**What I verified**
- **Snapshot/diff correctness** — `bootstrapSession` snapshots the jar into `Set<CookieIdentity>` (name+domain+path+value) *before* the request; `sessionCookies` returns only jar cookies absent from that snapshot. A stale cookie is never handed back, and a re-issued cookie (same name, new value) still reads as fresh because `value` is in the key.
- **Both regression tests are genuine, and robust to a test-env unknown** — I worked through both the "URLSession stores the minted cookie into the stub jar" and "it doesn't" hypotheses. Under *either*, `…ReturnsOnlyTheCookiesThisResponseSet` fails on old code and passes now, and `…TreatsAResponseThatSetsNoNewCookieAsAFailedMint` throws `.decoding` now vs. silently succeeding before. The fix and tests don't depend on that ambiguity — good design.
- **Coverage gate — prior review's open item resolved.** The gate (`check-coverage.py`) enforces **per-file line coverage**, with `ReadplaceAPI.swift`'s floor at **94%** (`coverage-baseline.json`). Every new/changed line — the `priorCookies` snapshot, the `CookieIdentity.init` body, the `.filter`, the `if !fresh.isEmpty` guard, and the header-fallback `return` — is executed by the new tests, so file line coverage can only hold or rise. The `?? []` / `guard let headers … else { return [] }` nil-branches the earlier review worried about don't matter here: line coverage marks the (executed) line covered regardless of the untaken sub-branch. **No coverage risk** — this isn't a strict branch/region gate.
- **Conventions** — synthesized `Hashable`/`Equatable` on `CookieIdentity` is intact (custom `init(_:)` doesn't suppress it; all stored props are `String`). Defining that init also suppresses the memberwise `init(name:domain:path:value:)`, so no four-`String` positional constructor exists to transpose. Domain-focused name (no `…Service`/`Manager`); no force-unwrap/`as!`/`try!`; `throw` (not `precondition`) is right for the recoverable failed-mint; comments are why-not-what.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Re-mint on a non-ephemeral jar is a deliberate behaviour change (informational)
A repeat `bootstrapSession` returning 2xx with **no** `Set-Cookie` while a still-valid `hutch_sid` sits in a **non-ephemeral** jar now throws `.decoding` where the old code returned the stale cookie. This is the intended hardening and is safe under the contract that a successful mint always sets a cookie. Production wires `.ephemeral`, so the jar is empty per launch and the common path is unaffected. Flagging only so the trade-off stays conscious.

### 2. Identical-value re-set relies on the untested header fallthrough (optional test)
The emergent property — if a response re-sets a cookie with a value **identical** to one already in the jar, `fresh` is empty and control falls through to `HTTPCookie.cookies(withResponseHeaderFields:)`, which still returns it — has no direct regression test. It can't occur for the single-`hutch_sid` mint (fresh random value each time), so this is genuinely low priority; a small test would only document the defensive fallthrough. Not required for the coverage gate.

FagnerMartinsBrack:

## Context
- **PR**: #958
- **Commit**: 1e9c12d9e9ff54f3087b6066d8341a42ec19a1d3

FagnerMartinsBrack:
### Code Review — PR #958 · `fix(ios): return only the cookies a bootstrap response set`

Re-review at `1e9c12d9`. The delta since the last review (`6cce4fc1`) is two commits — `df666ad5` (doc comment + README) and `1e9c12d9` (a new regression test). The PR payload vs `origin/main` is still iOS-only (README.md, ReadplaceAPI.swift, ReadplaceAPITests.swift); the core `sessionCookies` logic is byte-identical to the version reviewed twice before. This pass focused on the two new commits and on running the claims to ground against the server. **I also confirmed the `ios-tests` CI job is green at this commit**, which closes the one open item every prior review carried.

**What I verified this round**
- **The new test genuinely exercises the jar-diff branch.** `testBootstrapSessionReadsAReMintedSessionCookieBackFromTheJarDiff` seeds `hutch_sid=old` + `hutch_vid=v1`, then plants `hutch_sid=new` *inside* the stub handler — i.e. after `bootstrapSession`'s snapshot and before `sessionCookies` re-reads the jar — and returns a `204` with **no** `Set-Cookie`. So the header fallback can only return `[]`; the result `["hutch_sid"]=new` can *only* come from the store delta. `hutch_vid` (unchanged, in `prior`) is correctly excluded. This is the branch the earlier tests (which land on the header fallback) never hit. All three tests now cover distinct paths: test 1 → header fallback (empty delta), test 2 → jar diff, test 3 → failed-mint. All three remain robust to the "does URLSession land the stub's `Set-Cookie`?" ambiguity — test 2 sidesteps it entirely by planting manually.
- **The `hutch_ext_alive` claim in the doc comment is accurate.** I traced it end-to-end: `mark-extension-installed.middleware.ts` sets `hutch_ext_alive` on **every** Siren response (`wantsSiren(req)`), the value is the constant `"1"` (`ALIVE_COOKIE_VALUE`), and iOS `send()` attaches `Accept: application/vnd.siren+json` to **both** `loadQueue` and `bootstrapSession`. So it lands in the app jar from the queue load, sits in `prior`, and is dropped from the delta re-set-unchanged — exactly as the comment says.
- **This is a real fix, not just a doc/code reconciliation.** Because the *old* store path returned the whole jar, in production (post-queue-load, `hutch_ext_alive=1` present) it would have injected `hutch_ext_alive` — and any other server-re-set cookie — into the reader session. The new delta returns only the freshly-minted `hutch_sid`. `extension-install.ts` keys onboarding off exactly that cookie, so the old behaviour could have faked "extension installed" from the app. Good catch by the change.
- **CI is green at `1e9c12d9`** — run [`29391108611`](https://github.com/Readplace/readplace.com/actions/runs/29391108611), job `ios-tests` = **success**. Every new/changed line is executed by the tests, so `ReadplaceAPI.swift`'s 94% line-coverage floor holds.
- **Conventions & hygiene** — `CookieIdentity` keeps its synthesized `Hashable`; no force-unwrap/`as!`/`try!`; `throw` (not `precondition`) for the recoverable failed-mint; README accurately restated. Both commits follow Conventional Commits (`docs`/`test`, scope `ios-readplace`) and are cleanly split.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The doc comment now couples to another subsystem's wire constant (staleness risk)
The expanded `sessionCookies` doc comment names `hutch_ext_alive` and walks through server-side behaviour owned by `mark-extension-installed.middleware.ts` / `@packages/onboarding-extension-signal`. Per the [code-comments skill](.claude/skills/code-comments/SKILL.md) ("Never name another file or symbol… the path goes stale silently when that file is moved or renamed"), if the server renames the cookie or stops re-setting it with an unchanged value, this Swift comment drifts with no compiler or test to catch it. The *core* why — exclude `prior` to keep the failed-mint check honest, read the store to dodge the folded-`Set-Cookie` hazard, and drop signals the server re-sets unchanged — is legitimately comment-worthy and can't be carried by code alone. Worth considering whether the concrete `hutch_ext_alive` walkthrough (and the parenthetical about the test's stubbed `URLProtocol`, which reaches from production doc into test infrastructure) belongs in the comment versus the commit message / PR body. Not blocking — purely a durability call.

### 2. The `hutch_ext_alive` exclusion is flow-dependent (informational)
The comment states `hutch_ext_alive` is dropped because it is "already in `prior` from the queue load that precedes a mint." That holds for today's flow, but the mint's **own** Siren response *also* re-sets `hutch_ext_alive`. So if `bootstrapSession` were ever called without a preceding Siren GET in the same jar (e.g. a future "pre-warm the session on launch" path), `hutch_ext_alive` would be absent from `prior`, enter the delta as fresh, and reach the reader — the exact leak this guards against. Not reachable now (the reader is only entered after a queue load, and production wires an `.ephemeral` jar), so this is informational — flagging only so the queue-before-mint coupling stays a conscious invariant rather than an accident of ordering.

PR #963, opened 2026-07-12
fix(hutch): gate URL unwrap on trimmed length to match the validator

FagnerMartinsBrack:
## What

`withUnwrapPreprocessing` (in `projects/hutch/src/runtime/web/unwrap-preprocessors.ts`) is the decorator that unwraps Readplace `/view/…` self-URLs down to the underlying article URL before the save-path validator runs. Its length gate measured the **raw** string length, while `validateSaveableUrl` measures the **trimmed** length.

That mismatch opened a gap: a self-URL padded past `MAX_SAVEABLE_URL_LENGTH` (8192) with whitespace that trims back under the cap would

1. fail the decorator's `value.length <= MAX` gate → **skip unwrapping**, then
2. be passed to the validator, which trims it back under the cap → **pass validation**,

so the self-referential `/view` wrapper was stored verbatim instead of the underlying article — exactly the self-referential record the unwrap step exists to prevent.

## Fix

Measure `value.trim().length` and hand the **trimmed** string to the preprocessor, so the gate agrees with the validator (which trims before it counts). Non-string input still passes straight through to the validator.

## Tests

Added a unit test in `unwrap-preprocessors.test.ts` covering a whitespace-padded, over-length self-URL: the padded string exceeds `MAX_SAVEABLE_URL_LENGTH`, yet the decorator now trims first, so the preprocessor is invoked with the trimmed self-URL and the validator receives the unwrapped article URL.

`pnpm nx run hutch:check` passes with `unwrap-preprocessors.ts` at 100% coverage.

---
_Generated by [Claude Code](https://claude.ai/code/session_017yBX9Zzsbue4SnJdDy6sp1)_

### DIFF
--- a/projects/hutch/src/runtime/web/unwrap-preprocessors.test.ts
+++ b/projects/hutch/src/runtime/web/unwrap-preprocessors.test.ts
@@ -80,6 +80,17 @@ describe("withUnwrapPreprocessing — validator decorator", () => {
 		expect(seen()).toBe(oversized);
 	});
 
+	it("trims before measuring length, so a whitespace-padded over-length self-URL is still unwrapped", () => {
+		const { validate, seen } = captureValidator();
+		const selfUrl = "https://readplace.com/view/example.com/article";
+		const padded = `${selfUrl}${" ".repeat(MAX_SAVEABLE_URL_LENGTH)}`;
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #963
- **Commit**: a909ceaa27f14d682170210aa8d761b3ed7d8dab

FagnerMartinsBrack:

PR #966, opened 2026-07-13
fix(hutch,ios-readplace): unify Back to queue on iOS chromeless sheets

FagnerMartinsBrack:
## Problem

The "back to queue" affordance was inconsistent across the iOS app's in-app sheets:

| Sheet | Back affordance |
|-------|-----------------|
| Reader | web-rendered `← Back to queue` deep link (chromeless) |
| Account | web-rendered `← Back to queue` deep link (chromeless) |
| **Add link** | **native nav bar + `Back to Queue` toolbar button** |

Two of the three sheets already shared the chromeless web-rendered affordance — the reader is the shared anchor (it needs web-rendered actions), and the account page was aligned to it in the previous commit. The add-links help sheet was the lone outlier, differing in label, casing (`Back to Queue` vs `← Back to queue`), and mechanism (native toolbar vs web link).

## Change

Bring the add-links sheet into the same chromeless pattern so all three in-app sheets return to the native list identically.

**Server (hutch)**
- `/help/add-links` renders the same chromeless `← Back to queue` deep link `/account` renders, gated on the `?shell=app` marker. A browser visitor (no marker) keeps the bare public page — the `readplace://` scheme would be a dead end there — and an older store build that never sends the marker keeps the native toolbar it already ships. The app variant also hard-codes the home-indicator bottom pad, matching the account/reader rationale (the WKWebView viewport sets no `viewport-fit=cover`, so `env(safe-area-inset-bottom)` is 0).

**iOS**
- The app appends `?shell=app` to its client-held help URL (the same marker `open(link:)` puts on the account href).
- `AddLinkInstructionsView` drops its `NavigationStack`/toolbar and hosts the page edge-to-edge like the reader/account sheet.
- `WebPageView` intercepts the `readplace://reader/close` back link — reusing `ReaderNavigation`'s single close contract — to dismiss.
- The native fallback shown when the page can't load keeps its own native back button, mirroring the reader's unavailable view.
- The close deep link now has one shared source of truth documented in `ios-app-links.ts`.

## Testing

- `pnpm check` passes across all 42 projects with 100% coverage thresholds met.
- Added route tests: the help page renders the `← Back to queue` deep link (href `readplace://reader/close`, `main.help--app`) under `?shell=app`, and omits it for an ordinary browser visitor.
- Updated the iOS view-model test to expect the `?shell=app` marker on the help URL.
- iOS Swift build/tests run in CI on macOS (no Swift toolchain in this environment); the changed Swift files are the coverage-excluded OS-boundary/layout files plus a one-line view-model change covered by the existing suite.

https://claude.ai/code/session_01DCsqbgF5BeS4jq6k5HJAn7

---
_Generated by [Claude Code](https://claude.ai/code/session_01DCsqbgF5BeS4jq6k5HJAn7)_

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -195,6 +195,8 @@ import { PrivacyPage } from "./web/pages/privacy";
 import { SupportPage } from "./web/pages/support";
 import { TermsPage } from "./web/pages/terms";
 import { HelpAddLinksPage } from "./web/pages/help";
+import { isAppShell } from "./web/onboarding/ios-client";
+import { APP_BACK_LINK } from "./web/shared/ios-app-links";
 import { E2EFixturePage } from "./web/pages/e2e-fixture";
 import { createE2EFixturePdf } from "./web/pages/e2e-fixture-pdf";
 import { initInstallRoutes } from "./web/pages/install";
@@ -760,8 +762,17 @@ export function createApp(dependencies: AppDependencies): Express {
 	// holds this path client-side; the /queue collection still advertises it via the
 	// add-links-help rel so older installed clients resolve the help URL from there.
 	// Either way the copy ships via a hutch deploy rather than an App Store review.
+	//
+	// The app sheet appends ?shell=app, which earns the same chromeless "← Back to
+	// queue" deep link the account page renders — so both surfaces the sheet hosts
+	// return to the native list identically. A browser visitor (no marker) keeps the
+	// bare page, and an older store build that never sends the marker keeps the
+	// native toolbar it already ships.
 	app.get("/help/add-links", (req: Request, res: Response) => {
-		sendComponent(req, res, HelpAddLinksPage({ staticBaseUrl }));
+		const backLink = isAppShell(req)
+			? { href: APP_BACK_LINK.topHref, label: APP_BACK_LINK.label }
+			: undefined;
+		sendComponent(req, res, HelpAddLinksPage({ staticBaseUrl, backLink }));
 	});
 
 	// Path-uniqued article fixture for staging e2e tests. The :id segment is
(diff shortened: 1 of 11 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #966
- **Commit**: a9d368f94856429337b6b1803a7b2d6bfc6afc0a

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `WebPageView` collapses `.logout`/`.openExternally` into `.allow` (unreachable today, by design)

`WebPageView.Coordinator.webView(_:decidePolicyFor:decisionHandler:)` routes every navigation through `ReaderNavigation.decide` but only special-cases `.close`; `.logout` and `.openExternally(url)` fall through the `default` branch to `decisionHandler(.allow)`.

For this page that's correct — the help page emits only the close deep link and no external/logout links (verified: the sole anchor in `add-links.template.html` is the `.help__back` link, and images are resource loads that don't hit `decidePolicyForNavigationAction`). It's a slightly leaky reuse of the richer reader decision, though: if the help copy ever gains an outbound link it would load *inside* the sheet rather than open in the browser the way the reader's `.openExternally` does, and a stray `readplace://account/logout` would try to load a custom scheme in the WKWebView. Since the reuse is documented as intentional and the path is unreachable now, this is purely optional — either narrow the match to the close href, or handle `.openExternally` for parity with the reader. `projects/ios-readplace/App/WebPageView.swift:60-76`.

### 2. Brief loading window shows no visible back control (swipe-to-dismiss only)

After removing the `NavigationStack`/toolbar, the happy path shows only a `ProgressView` while the page loads; the web-rendered `← Back to queue` link appears once the page finishes. During that short window the only way back is the sheet's interactive swipe-dismiss (not disabled, so it works). This matches the reader sheet — the stated anchor for the pattern — so it's arguably correct-by-design; noting only because the old toolbar button was visible throughout the load. `projects/ios-readplace/App/AddLinkInstructionsView.swift:23-42`. No change required.

PR #978, opened 2026-07-18
docs: reframe README security section around the guarantee, not the mechanism

FagnerMartinsBrack:
## Summary

- The README's Security section led with "Redirect de-duplication is pinned to the destination" — implementation-detail phrasing that's too low-level for a README audience.
- Reworded the heading and body to lead with the security guarantee ("Shared articles can't be hijacked through a redirect") and the threat it defends against, with the destination-pinning mechanism now supporting that claim rather than being the headline.
- No behavior change — documentation only.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KNowR4Vrrt5MAuY8ddP1ym)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,13 @@ The codebase has strong opinions on testing, typing, branded IDs, and comments t
 
 ## Security
 
-### Redirect de-duplication is pinned to the destination
+### Shared articles can't be hijacked through a redirect
 
-Saving a link follows its redirects and keys the article on where it lands, so `site.com/post.html` (saved from the web) and `site.com/post` (saved from the extension) collapse into one article instead of two. Articles are global — every user who saves the same destination attaches to the same shared row.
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #978
- **Commit**: 589bc1e091be35bdd1e03c327510519e2f2615c1

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Optional: the reword drops the one clause that named the attack's mechanism

Paragraph 2 previously opened *"Since the redirect terminal becomes an alias to whichever URL was saved first, an attacker could…"*. The reword removes that causal clause. The later *"before anyone else claims it"* still carries the first-writer-wins idea, so the paragraph stays correct — and trimming mechanism detail is exactly this PR's stated goal. Flagging only so you can confirm the implication reads clearly enough on its own; no change required.

PR #986, opened 2026-07-22
feat(hutch): stop paywalling short /view reads and founder-blog referrals

FagnerMartinsBrack:
## Why

A reader who follows a blog hyperlink to a public `/view/<url>` page hit an expiry countdown and a scroll-gated blur paywall once the 3-day access window elapsed — even on a short piece, and even when they came straight from the founder's own blog. A reader called it out on fagnerbrack.com:

> What the hell is this "readplace" link that wants you to sign in just to be redirected to a Jeff Atwood post from 2007?

The wall exists to nudge readers to save the **long reads they won't finish now**. On a quick read, and on the founder's own syndication audience, it's pure friction — and when it does appear, it never said what signing in actually buys you.

## What changed

Expiry decisions live in `computePublicViewExpiry` ([`view-expiry.ts`](projects/hutch/src/runtime/web/pages/view/view-expiry.ts)) — a permanent (`null`) expiry means no countdown and no paywall — and only apply to logged-out readers (the only ones the window ever gated).

**1. Read-time threshold.** Any article estimated at **5 minutes or less** stays permanently public. Longer reads keep the existing 3-day window. The threshold is one named constant, `PUBLIC_VIEW_PAYWALL_READ_MINUTES_THRESHOLD`. (Read time is only known after crawl, and the paywall only renders on a `ready` crawl, so a short article settles to "no wall" exactly when its length becomes known.)

**2. Founder-blog referrer bypass.** When the HTTP `Referer` names a founder-blog host — `fagnerbrack.com` or a subdomain — the paywall never appears, **whatever the article's length**. Scoped to `fagnerbrack.com` deliberately (`PERMANENT_REFERRER_DOMAINS`): bare `medium.com` would hand the bypass to all of Medium's traffic rather than the founder's own readers. A missing or unparseable `Referer` (stripped by a referrer policy) falls through to the normal rules.

**3. Clearer paywall copy.** When the wall *does* appear, its body now states the payoff for signing in instead of pitching a queue feature:

> _before:_ "Save this link to your readplace queue and read every link without expiration."
> _after:_ "Sign in to save it to your queue and read the whole article in reader view."

These compose with the existing bypasses (article hosted on a permanent domain, valid authenticated sharer).

## Tests

- Unit tests for `computePublicViewExpiry`: short/at-threshold → permanent, one minute past → the standard expiry, founder referrer → permanent even for a long read; existing domain/sharer cases pass a long read so only the property under test decides.
- Unit tests for `isPermanentReferrer`: host + subdomain + origin-only + case-insensitive match; lookalike suffix, bare `medium.com`, missing, and unparseable all reject.
- Route tests: a short expired read and a founder-blog-referred long expired read both render `permanent` state with no `[data-test-view-paywall]` and the full unblurred article; a long read from an unrelated referrer still counts down and blurs.
- Component test updated to assert the new paywall body copy.
- Existing expiry/paywall route seeds bumped to a `LONG_ARTICLE_WORD_COUNT` so they keep exercising the expiry path.
- `pnpm nx run hutch:check` passes (lint, types, 3303 tests, 100% coverage thresholds met); the full monorepo pre-commit `check` passes.

https://claude.ai/code/session_01LxsWRSsXDLp2dAbegYT29V

### DIFF
--- a/projects/hutch/src/runtime/web/pages/view/view-expiry.test.ts
+++ b/projects/hutch/src/runtime/web/pages/view/view-expiry.test.ts
@@ -3,10 +3,12 @@ import { MinutesSchema } from "@packages/domain/article";
 import { UserIdSchema } from "@packages/domain/user";
 import {
 	PERMANENT_ARTICLE_DOMAINS,
+	PERMANENT_REFERRER_DOMAINS,
 	PUBLIC_VIEW_ACCESS_WINDOW_MS,
 	PUBLIC_VIEW_PAYWALL_READ_MINUTES_THRESHOLD,
 	computePublicViewExpiry,
 	formatSaveUtmContent,
+	isPermanentReferrer,
 	sharedUserIdFrom,
 	sharedUserIdFromQueryParams,
 } from "./view-expiry";
@@ -69,6 +71,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "example.com",
 			permanentArticleDomains: PERMANENT_ARTICLE_DOMAINS,
 			isValidSharer: false,
+			isPermanentReferrer: false,
 			estimatedReadTime: longRead,
 		});
 		assert(result.expiresAt, "expiresAt must be set for organic visits");
@@ -82,6 +85,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "fagnerbrack.com",
 			permanentArticleDomains: PERMANENT_ARTICLE_DOMAINS,
 			isValidSharer: false,
+			isPermanentReferrer: false,
 			estimatedReadTime: longRead,
 		});
 		assert.equal(result.expiresAt, null);
@@ -93,6 +97,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "example.com",
 			permanentArticleDomains: PERMANENT_ARTICLE_DOMAINS,
 			isValidSharer: true,
+			isPermanentReferrer: false,
 			estimatedReadTime: longRead,
 		});
 		assert.equal(result.expiresAt, null);
@@ -104,6 +109,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "example.com",
 			permanentArticleDomains: PERMANENT_ARTICLE_DOMAINS,
 			isValidSharer: false,
+			isPermanentReferrer: false,
 			estimatedReadTime: longRead,
 		});
 		assert(result.expiresAt, "expiresAt must be set for non-sharer, non-permanent visits");
@@ -118,6 +124,7 @@ describe("computePublicViewExpiry", () => {
 				articleDomain: domain,
 				permanentArticleDomains: multiDomains,
 				isValidSharer: false,
+				isPermanentReferrer: false,
 				estimatedReadTime: longRead,
 			});
 			assert.equal(result.expiresAt, null, `expected permanent for ${domain}`);
@@ -130,6 +137,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "unknown.com",
 			permanentArticleDomains: ["first.com", "second.com", "third.com"],
 			isValidSharer: false,
+			isPermanentReferrer: false,
 			estimatedReadTime: longRead,
 		});
 		assert(result.expiresAt, "expiresAt must be set for non-permanent domain");
@@ -142,6 +150,7 @@ describe("computePublicViewExpiry", () => {
 			articleDomain: "example.com",
 			permanentArticleDomains: PERMANENT_ARTICLE_DOMAINS,
 			isValidSharer: false,
+			isPermanentReferrer: false,
 			estimatedReadTime: MinutesSchema.parse(2),
 		});
 		assert.equal(result.expiresAt, null);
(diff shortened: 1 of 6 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #986
- **Commit**: 3488b5c8f817f1c75f224fc0e266dbccfbd98322

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Confirm the 5-minute threshold covers the motivating case

The change is driven by the complaint about being walled in front of "a Jeff Atwood post from 2007." Coding Horror posts vary a lot in length — many run 800–1900 words, which at 238 wpm is **4–8 minutes**. Anything computing to **6+ minutes still hits the wall** (`estimatedReadTime > 5`), so the specific post that triggered the complaint may not actually be freed by this threshold.

This is a product-tuning judgment, not a code defect — and the PR does the right thing by centralizing the value in `PUBLIC_VIEW_PAYWALL_READ_MINUTES_THRESHOLD` (`view-expiry.ts:23`) so it's trivial to raise. Worth a quick sanity-check that 5 is the number you want, or whether ~7–8 better matches "finishable in one sitting."

### 2. (Nit) Constant name is slightly narrower than what it gates

`PUBLIC_VIEW_PAYWALL_READ_MINUTES_THRESHOLD` gates the **entire** expiry mechanism — both the scroll-gated paywall *and* the expiry countdown/urgency (`view.page.ts:274`), since a short read returns `expiresAt: null` and therefore renders `data-expiry-state="permanent"` with no counter. The name reads as if only the paywall is gated. Not worth churning on its own; mentioning only in case a future reader is misled.

FagnerMartinsBrack:

## Context
- **PR**: #986
- **Commit**: b8de2ed23e7fe8cbaba62a583ecf02eae8c09007

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The founder-referrer bypass — not the 5-minute threshold — is what actually resolves the motivating complaint

The reporter "came **straight from the founder's own blog**," so the new `isPermanentReferrer` check (commit `b8de2ed2`) frees that exact path **regardless of length** — this is the change that closes the reported issue. The prior review's worry (a 6–8 min Coding Horror post exceeds the 5-min threshold) now only applies to readers arriving from **other** sources (HN, X, Google) to a long post; for them the wall still appears. Worth confirming that residual is intended.

### 2. `Referer` is client-controlled, so the bypass — and therefore the paywall — is best-effort

Anyone can send `Referer: https://fagnerbrack.com` and skip the wall on any article (`view.page.ts:279`). Acceptable here — the `/view` content is already public and the "paywall" is a soft save-nudge, not an access boundary — so this bypasses a growth device, not a security control. Flagging only so the porousness is a deliberate call.

### 3. During a *pending* crawl the expiry now settles to `permanent` (behaviour change)

`stubReadTime = calculateReadTime(0) = 1` minute (≤ threshold), so before the real length is known `computePublicViewExpiry` returns `{ expiresAt: null }`. A long read shows no countdown while crawling, then counts toward the same `savedAt + 3-day` deadline once `ready`. Benign (deadline anchored to `savedAt`, no reset; pending UI is the "Your link is saved" reframe), but a behaviour change from the pre-PR running countdown. Confirm intended.

FagnerMartinsBrack:

## Context
- **PR**: #986
- **Commit**: 0fe6c81dbcad6a3836109a62785eb6aa5d8996b0

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. (Nit) The body now says "Sign in" but the only CTA is still labelled "Save to My Queue"

The reframed body leads with the verb **"Sign in to save it…"**, while the single button underneath is unchanged: `Save to My Queue` (`view-paywall.template.html:5-6`). For the logged-out reader this paywall targets the flow is coherent — clicking *Save to My Queue* hits `/save?url=`, which `303`-redirects to `/signup?return=…` (`save.page.ts:44-64`) — so "Sign in" accurately describes what the button *does*. It's only a surface wording mismatch (body promises "sign in", button says "save"), and arguably intentional given the deferred-auth model. Not worth churning; flagging only so the verb/label split is a conscious choice. If you did want them aligned, the body could lead with the outcome ("Save it to your queue to read the whole article in reader view") and leave "sign in" implicit in the click.

### 2. (Nit) "read the whole article in reader view" is the eventual payoff, not the immediate next screen

The claim is truthful — the queue has a first-class reader view and the authenticated `/view` renders the full unblurred article permanently. Worth being aware that the *immediate* post-save destination is `/queue?url=…` (`save.page.ts:69`), so the reader lands in the queue and opens reader view from there rather than dropping straight into it. The copy honestly describes the value, so this is purely informational — no change implied.

### Standing PR-level items (already raised in prior reviews, unchanged by this commit)

Carried forward only so this latest review reflects the whole PR — all three are product-judgment calls already acknowledged, not defects:

- **5-minute threshold vs the motivating post** — a 6–8 min Coding Horror read from a non-founder referrer still walls; the founder-referrer bypass is what frees the reported path. Centralized in `PUBLIC_VIEW_PAYWALL_READ_MINUTES_THRESHOLD`, trivial to raise.
- **`Referer` is client-controlled** — the founder-blog bypass is best-effort; acceptable because `/view` is already public and the wall is a soft save-nudge, not an access boundary.
- **Pending crawl now settles to `permanent`** — `stubReadTime = 1 min ≤ threshold`, so a long read shows no countdown while crawling, then counts toward the same `savedAt`-anchored deadline once `ready`. Benign (no clock reset), but a behaviour change from the pre-PR running countdown.

PR #988, opened 2026-07-23
perf: batch the per-card queue summary/crawl reads into two BatchGets

FagnerMartinsBrack:
## Summary

The `/queue` render fired **2N individual DynamoDB GetItems** — one `findGeneratedSummary` and one `findArticleCrawlStatus` per card (40 at pageSize 20) — each issued **without a projection**, so every one hauled the full article row including legacy inline `content` blobs (up to 400 KB/item).

This adds two batched read contracts implemented over the existing `batchGetFromTable` helper with a **content-excluding projection**, collapsing the 2N GetItems into **2 BatchGets** of projected metadata.

```ts
export type FindGeneratedSummaries = (urls: readonly string[]) => Promise<ReadonlyMap<string, GeneratedSummary | undefined>>;
export type FindArticleCrawlStatuses = (urls: readonly string[]) => Promise<ReadonlyMap<string, ArticleCrawl | undefined>>;
```

## Contract semantics (match today's per-item `allSettled`)
- Every input url is keyed (value possibly `undefined`), keyed by the url as given.
- A **missing**, **unparseable**, or **schema-poisoned** row degrades to `undefined` for **that url only** — a loose batch schema is fed to `batchGetFromTable` (so one malformed row doesn't throw the whole batch), then each row is re-validated with strict `safeParse` + the existing mapper inside `try/catch`.
- Duplicate-normalizing urls dedupe to one BatchGet key and fan the result back out.
- A whole-batch **transport failure rejects** and is now logged via `deps.logError` (previously swallowed silently by `allSettled`).
- Empty input → empty map, no network call.

## What win this is (honest)
The 40 GetItems already ran in parallel, so wall-clock ≈ 1 in-region RTT today. The gains are: **2 SDK request cycles instead of 40**, **~2 × 20 × ~1 KB projected bytes instead of 40 × full rows (incl. content blobs)**, and **2 tail-latency samples instead of 40**. RCU is unchanged. Expect ~20–80 ms of serial serialize/parse think-time off the `/queue` render on a 512 MB Lambda — to be measured post-deploy.

## Scope preserved
- The **singular** contracts (`findGeneratedSummary` / `findArticleCrawlStatus`) stay for the reader, `GET /queue/:id/card`, `/view`, admin recrawl, MCP, and digest paths.
- The Siren `GET /queue` branch returns before the batch call — untouched.
- Production wires the real BatchGet; **dev and tests compose the batch by looping the in-memory singular** (`batchFromSingular`), preserving the one-read-per-url count the summary-poll fake depends on for its pending → ready transition.

## Testing
- `pnpm check` green across all 45 projects, **100% coverage** on the new code (the batch impls, `batch-from-singular`, and the degraded-render route test).
- New unit tests (fake-client capturing the `BatchGetCommand`): single BatchGet per store, normalized + deduped keys, `ProjectionExpression` excludes `content`, missing row → `undefined`, poisoned row (unknown status enum) degrades only its url, mapper-assert trip degrades without throwing the batch, unparseable url never reaches Dynamo, empty input sends nothing.
- New route test: a whole-batch rejection still renders `/queue` 200 with the cards (minus chips) and logs both failures.
- Existing queue/reader/view route suites and E2E pins pass unchanged.

https://claude.ai/code/session_01FMZviBeuMsmp3JsocavnUE

---
_Generated by [Claude Code](https://claude.ai/code/session_01FMZviBeuMsmp3JsocavnUE)_

### DIFF
--- a/projects/hutch/src/runtime/app.ts
+++ b/projects/hutch/src/runtime/app.ts
@@ -51,6 +51,7 @@ import { devSummariseInline } from "./providers/article-summary/dev-summarise-in
 import { initDynamoDbArticleCrawl } from "@packages/article-store";
 import { initInMemoryArticleCrawl } from "@packages/test-fixtures/providers/article-crawl";
 import { initInMemoryGeneratedSummary } from "@packages/test-fixtures/providers/article-summary";
+import { batchFromSingular } from "./batch-from-singular";
 import { S3Client } from "@aws-sdk/client-s3";
 import { SchedulerClient } from "@aws-sdk/client-scheduler";
 import { initS3ReadContent } from "@packages/article-store";
@@ -429,8 +430,10 @@ function initProviders(input: { appOrigin: string }) {
 			statPendingUpload,
 			readPendingUploadPrefix,
 			findGeneratedSummary: summaryStore.findGeneratedSummary,
+			findGeneratedSummaries: summaryStore.findGeneratedSummaries,
 			markSummaryPending: summaryStore.markSummaryPending,
 			findArticleCrawlStatus: crawlStore.findArticleCrawlStatus,
+			findArticleCrawlStatuses: crawlStore.findArticleCrawlStatuses,
 			findArticleFreshness: articleStore.findArticleFreshness,
 			findArticleCrawlVersions: articleStore.findArticleCrawlVersions,
 			markCrawlPending: crawlStore.markCrawlPending,
@@ -716,8 +719,10 @@ function initProviders(input: { appOrigin: string }) {
 		statPendingUpload,
 		readPendingUploadPrefix,
 		findGeneratedSummary: summaryStore.findGeneratedSummary,
+		findGeneratedSummaries: batchFromSingular(summaryStore.findGeneratedSummary),
 		markSummaryPending: summaryStore.markSummaryPending,
 		findArticleCrawlStatus: crawlStore.findArticleCrawlStatus,
+		findArticleCrawlStatuses: batchFromSingular(crawlStore.findArticleCrawlStatus),
 		findArticleFreshness: articleStore.findArticleFreshness,
 		findArticleCrawlVersions: articleStore.findArticleCrawlVersions,
 		markCrawlPending: crawlStore.markCrawlPending,
--- a/projects/hutch/src/runtime/batch-from-singular.test.ts
+++ b/projects/hutch/src/runtime/batch-from-singular.test.ts
@@ -1,0 +1,33 @@
+import { batchFromSingular } from "./batch-from-singular";
+
+describe("batchFromSingular", () => {
+	it("keys every input url to the singular's result, in one loop", async () => {
+		const seen: string[] = [];
+		const singular = async (url: string) => {
+			seen.push(url);
+			return url.endsWith("miss") ? undefined : `value:${url}`;
+		};
+
+		const batch = batchFromSingular(singular);
+		const result = await batch(["https://a", "https://b", "https://miss"]);
+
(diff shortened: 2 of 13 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #988
- **Commit**: 1d460e8b73bbc1ad8f24f2526d39fbc27eee7fdd

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The two BatchGets hit the same table with the same keys — a combined projection would halve RCU

`findGeneratedSummaries` and `findArticleCrawlStatuses` both read `articlesTable` keyed by the same article URLs (`app.ts:227-231`), so `/queue` now issues **2 BatchGets that each read every row once** — the same 2-reads-per-item RCU as the old `2N` GetItems (the PR correctly says "RCU is unchanged"). A single BatchGet projecting *both* the summary and crawl attribute sets would read each row **once**, cutting request count to 1 and halving RCU. The reason to *defer* is real: `dynamodb-generated-summary.ts` and `dynamodb-article-crawl.ts` are deliberately separate providers with separate schemas/mappers, and merging their reads couples them. Given the queue caps at ~20 keys the extra BatchGet is cheap, so keeping the separation is a reasonable call — noting the RCU upside for a future pass.

### 2. Whole-batch degradation widens the blast radius for transport/throttle failures (documented, acceptable)

On a rejected batch, `loadSummaries`/`loadCrawls` degrade **all** cards to "no chip" (`queue.page.ts:285-290`), vs. the old `allSettled` isolating a single failing card. For a genuine transport failure the old 40 GetItems shared one connection and would mostly have failed together anyway, so the effective blast radius is unchanged — and it's now **logged** instead of silently swallowed. `batchGetFromTable` also throws after exhausting `UnprocessedKeys` retries (`define-table.ts:221`), so sustained throttling degrades the whole listing. No action needed; flagging for awareness.

### 3. `input as Captured["commands"][number]` casts in the batch tests

The new suites narrow the fake client's `(input: unknown)` callback with `as`. This matches the file's existing fake-client idiom and is forced by the shared callback signature, so it's within the test-fake exception in CLAUDE.md — could optionally be narrowed via a typed capture helper, but not required.

FagnerMartinsBrack:
@claude Reviewed the full diff against main. This is a clean perf change: the 2N unprojected per-card GetItems on `/queue` collapse into two projected BatchGets, and the tricky parts are handled correctly — the loose batch schema plus per-row strict `safeParse` preserves per-row degradation, the projection derived from `Row.keyof().options` fetches every mapper-read field while excluding legacy `content` blobs, and the URL dedup is required for correctness since BatchGetItem rejects duplicate keys. `batchGetFromTable` already chunks at 100 keys with UnprocessedKeys retries, so the unpaginated validation-error re-render path is safe too. The previously-silent failure path is now logged and covered by a route test.

No critical, high, or medium findings. One new low: the ~45-line normalize/dedupe/fan-out block is duplicated between `dynamodb-article-crawl.ts` and `dynamodb-generated-summary.ts` — extracting a shared helper is optional since it trades duplication for coupling the two providers.

Existing low-priority review items:
1. **Combine the two BatchGets into one to halve RCU** — needs a human call: it couples two deliberately separate providers, and at ~20 keys the second request is cheap; reasonable to defer as the review itself noted.
2. **Whole-batch degradation blast radius** — fine as-is: the old 40 GetItems shared one client and would mostly fail together anyway, and the failure is now logged instead of swallowed.
3. **`as` casts in the batch tests** — fine as-is: forced by the fake client's `(input: unknown)` callback and within the documented test-fake exception.

PR #992, opened 2026-07-23
fix(inbox): give the emails list and manage page the detail column width

FagnerMartinsBrack:
## Problem

The three `/inbox` surfaces disagreed on their container width:

- `.inbox` (manage-addresses page) — `max-width: 640px`
- `.inbox-emails` (emails list) — `max-width: 640px`
- `.inbox-email-detail` (email detail) — `max-width: 760px`

All three share `margin: 0 auto; padding: 24px 20px;`, so on any viewport wider than ~680px the content's left/right edge jumped ~60px sideways on every emails-list → email-detail navigation — the feature's core loop. The detail page's 760px already matches hutch's `/queue`, so the two 640px surfaces were the outliers.

## Change

- `inbox.styles.css` — `.inbox` `max-width: 640px → 760px`
- `inbox-emails.styles.css` — `.inbox-emails` `max-width: 640px → 760px`
- `.inbox-email-detail` left at 760px (the reference value; untouched)

Converging on 760px keeps the content edge fixed across all three inbox pages and lines them up with `/queue`. Row children are fluid (`grid`/`flex` with `minmax(0, 1fr)` / `flex: 1 1`), so widening only lengthens the truncation point of sender/subject ellipses — no fixed-width children break. Header margins (32/24/16px) were intentionally left alone; they are separate compositions and out of scope for the width fix.

Chose to hardcode 760px (matching how the detail page and hutch's queue already express it) rather than introduce a shared token, since the value has exactly four consumers and no theme variance. A shared `--list-max-width` token can be revisited if a fifth list surface appears.

## Test

New `inbox-column-width.test.ts` extracts the container `max-width` from each of the three exported inbox stylesheets and asserts all three equal the detail value — guarding *consistency* rather than pinning a magic number, so the containers can't silently drift apart again. Verified the guard is not vacuous: recompiling with a deliberate 640/760 drift makes the test fail; restoring 760 makes it pass.

`pnpm nx run inbox:check` is green (lint, type-check, tests, 100% coverage). No colours, client JS, schema, or shared-token changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XdUfZyA5FPoJS7roo8YksD)_


FagnerMartinsBrack:

## Context
- **PR**: #992
- **Commit**: 1327a3d548f17e1a90de2d8d9c9ae4bf56312dc4

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The new test parses CSS with a regex — the web skill routes this through an approval gate

`inbox-column-width.test.ts`'s `containerMaxWidth` pulls the value out with `css.match(new RegExp(...))` + `/max-width:\s*(\d+)px/`. The **web skill → "Structured Parsing Over Regex"** says: *"Never use regex to extract data from structured markup … Regex-based markup parsing requires explicit human approval."* CSS isn't in its explicit HTML/XML/JSON list, but it's the same class of thing, and this is the only regex-CSS-parse in the repo — so it's worth a conscious yes/no rather than sliding in as precedent.

Why this is LOW and not MEDIUM — the mitigations are real: the input is the repo's own committed CSS (not untrusted), the rules are flat, and `.match()` returns the first `.inbox {…}`, which is the container rule defined first in each file, so it's correct **today**. The fragility is future-facing — add an `@media`/CSS-nesting block or reorder the rule and `([^}]*)` (which stops at the first `}`) could capture the wrong body and silently read a different `max-width`.

Options: (a) get the exception documented/approved; (b) swap to a real parser — `postcss@8.5.6` is already in the workspace `node_modules` (though not a declared `inbox` dep, so it'd need a devDependency entry); or (c) see #2.

Minor sub-nit inside the same helper: `new RegExp(\`\\.${className}…\`)` interpolates `className` unescaped. Harmless here (the three values are hardcoded, metacharacter-free constants), but it's the usual `RegExp`-from-string smell if the inputs ever stop being literals.

### 2. A shared width token would make the invariant structural instead of tested

The test guards *consistency after the fact*; a single source of truth would make drift impossible **and** delete the CSS-parsing problem in #1 outright. All three inbox stylesheets already consume web-shell `:root` tokens (`var(--border)`, `var(--radius)`, …), so a `--inbox-container-max-width` (or the `--list-max-width` you name) defined once in the shared token sheet would let all three reference `var(...)` — and it could even cover `hutch`'s `/queue`, which this test doesn't currently assert against. You explicitly deferred this as YAGNI ("revisit if a fifth surface appears"), which is a defensible call — flagging it only as the more robust long-term shape than a regex guard, not as a required change.

FagnerMartinsBrack:

## Context
- **PR**: #992
- **Commit**: f78dab481efb559a8175b2d41b892db3dffc9ffe

FagnerMartinsBrack:
@claude Reviewed the diff against main: two one-line max-width bumps (640px → 760px) bring `.inbox` and `.inbox-emails` in line with `.inbox-email-detail` and hutch's `/queue`, fixing the ~60px content-edge jump on list ↔ detail navigation. The guard test asserts three-way width equality rather than pinning 760, so it protects the actual invariant. Conventions check out: `node:assert/strict`, named parameter object for the same-typed `{ css, className }` args, no `as` casts, no comments. No critical, high, or medium findings.

1. **Regex CSS parsing in the test** — *needs a human call*: the web skill's "Structured Parsing Over Regex" rule requires explicit approval for regex-based markup parsing, and the alternative (postcss) means declaring a new inbox devDependency. The regex is correct today (flat stylesheets, container rule first), so this is an approve-or-swap decision, not a bug.
2. **Unescaped `className` interpolation into `new RegExp`** — *fine to leave*: all call sites are hardcoded metacharacter-free literals; escaping would be defensive noise.
3. **Shared width token instead of a test-guarded constant** — *fine to leave*: the PR body already defers this as YAGNI with a concrete revisit trigger (a fifth list surface), and the reviewer marked it non-required.

**Rebased onto current main** (new head `f78dab48`). Branch is now origin/main (6e4a49c8) + 2 commits: the original width-convergence commit (CSS 640->760 for .inbox and .inbox-emails plus the three-container width guard test, byte-identical patch to the pre-rebase commit) and one fixup refreshing visual baselines. Main added a visual regression suite after this PR branched, with the emails-list and addresses-page baselines captured at 640px; the PR's width change invalidates them on both platforms. Darwin baselines were captured locally; linux baselines were regenerated inside mcr.microsoft.com/playwright:v1.60.0-noble (the exact pinned @playwr

*Validation:* pnpm install (already up to date). pnpm nx run inbox:check: PASSED (lint, jest unit tests including the new inbox-column-width.test.ts, 100% coverage across 65 files, playwright e2e-local phase). Direct darwin run of src/e2e/inbox-visual.e2e-local.ts: 3 passed. Linux verification inside mcr.microsoft.com/playwright:v1.60.0-noble without the update 

**Notes for review:**
- The fixup commit f78dab48 refreshes 4 visual baseline PNGs (inbox-empty and inbox-addresses, both platforms) from 640px to 760px wide - required because main's visual suite landed after this PR branched. Linux baselines came from the pinned playwright:v1.60.0-noble container per the update-visual-baselines.js pattern; without them, CI's linux run would fail against the stale 640px baselines. Eyeball the PNGs when reviewing.
- The regex-CSS-parse policy exception question for inbox-column-width.test.ts remains open and untouched, as instructed.
- Pre-existing/unrelated: inbox:lint failed once during the pre-commit hook with exit 130 (concurrently kill) and passed unchanged on immediate retry - nx flagged it as a flaky task. Not caused by this branch.
- projects/inbox has no update-visual-baselines script of its own; I mirrored projects/hutch/scripts/update-visual-baselines.js manually. Consider generalising that script to cover inbox's visual specs.

FagnerMartinsBrack:
### Review of PR #992 — `fix(inbox): give the emails list and manage page the detail column width`

Re-reviewed at `f78dab48`. The delta since the last review is the new commit **`f78dab48` (visual baseline refresh)** — I focused on validating that, and re-confirmed the width change and guard test still hold.

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The guard test parses CSS with a regex — still behind the web skill's approval gate

`inbox-column-width.test.ts:7-11` reads the value with `css.match(new RegExp(`\\.${className}\\s*\\{([^}]*)\\}`))` + `/max-width:\s*(\d+)px/`. This is the same item the prior review raised, and it's unchanged in this commit — flagging so it stays on the record for the human merge decision, not as a new finding.

- **Correct today.** `INBOX_STYLES` et al. are the exact CSS (`inbox.styles.ts` is just `readFileSync(...".styles.css")`), the rules are flat, and `.match` returns the first `.inbox {…}` — which is the container rule, defined first in each file. I checked all three files: no earlier `.<class>` is followed by `\s*{`, so no mismatch.
- **Fragility is future-facing.** `([^}]*)` stops at the first `}`; add CSS nesting / an `@media` wrapper (the detail sheet already ends in one) above the container rule, or reorder, and it can capture the wrong body and silently read a different `max-width`. The web skill's *Structured Parsing Over Regex* (`.claude/skills/web/SKILL.md:356`) puts regex-parsing of structured markup behind explicit human approval — CSS isn't in its HTML/XML/JSON list but is the same class.
- **Sub-nit:** `className` is interpolated into `new RegExp` unescaped. Harmless for the three metacharacter-free literals; the usual `RegExp`-from-string smell if inputs ever stop being constants.

Options: (a) get the exception approved; (b) swap to `postcss` (already in `node_modules`, would need an `inbox` devDependency entry). [Fix this →](https://claude.ai/code?q=Replace%20the%20regex%20CSS%20parse%20in%20projects%2Finbox%2Fsrc%2Fruntime%2Fweb%2Fpages%2Finbox%2Finbox-column-width.test.ts%20(branch%20claude%2Finbox-column-shift-fix-sod6vx)%20with%20a%20real%20CSS%20parser%20(postcss%20is%20already%20in%20node_modules%3B%20add%20it%20as%20an%20inbox%20devDependency)%20that%20reads%20max-width%20from%20the%20.inbox%2C%20.inbox-emails%20and%20.inbox-email-detail%20rules%2C%20keeping%20the%20three-way%20equality%20assertions%20and%20escaping%20className%20before%20building%20the%20RegExp.&repo=Readplace/readplace.com)

### 2. A shared width token would make the invariant structural instead of tested

Also unchanged from the prior review. All three sheets already consume web-shell `:root` tokens (`var(--border)`, `var(--radius)`), so a single `--inbox-container-max-width` would make drift impossible *and* delete the regex-parse in #1 outright — and could cover `hutch`'s `/queue` too (confirmed `.queue { max-width: 760px }`), which this test doesn't assert against. You deferred this as YAGNI ("revisit if a fifth surface appears"), which is defensible — noting it only as the more robust long-term shape.

### 3. Optional: the baseline refresh is split into its own commit

`b04a932d` ships the CSS `640 → 760` change and `f78dab48` refreshes the four PNGs it invalidates. That leaves `b04a932d` in a state where the `inbox-visual` e2e baselines are stale — a `git bisect` landing there would see a red visual job. Non-blocking (the PR head is green and per-commit-green isn't a stated repo rule); squash the two before merge if you want each commit independently bisectable.

FagnerMartinsBrack:
@claude address all low prio issues. Wew should NEVER test written code only the RESULT after running it (if css it m,ust be a UI test / screenshot)

FagnerMartinsBrack:
@claude review currnet changes against this guideloines:

Evidence Over Speculation
Every change must fix a problem that has been observed — a failing run, a measured behaviour, a reproduced bug — or implement an explicit requirement. Do not add guards, fallbacks, canaries, or standing infrastructure for problems that have not happened, even when a plan or a review proposes them: ask for the evidence or leave them out. If a safeguard still seems worth having, propose it as its own change with the evidence that motivates it.

FagnerMartinsBrack:

## Context
- **PR**: #992
- **Commit**: cb9ee050d3734aa65a961c30aa9860dd5e71ef9d

PR #1001, opened 2026-07-23
fix(inbox): center the skipped-link row and stack it on narrow screens

FagnerMartinsBrack:
## What

Two CSS-only fixes to the inbox email-detail **skipped-link** row (`.inbox-excluded-link` in `inbox-email-detail.styles.css`):

1. **Baseline → center alignment.** The row used `align-items: baseline`. Because the report button is a 44px `inline-flex` control, its first baseline sits well below its own box top, so the button's tall above-baseline distance defined the flex line's top edge and pushed the info block (URL + reason) down ~10px — leaving unexplained dead space above the URL and detaching the URL text from the "This is an article (report)" label. Switching to `align-items: center` centers the info block against the button and removes the dead space, matching the sibling `.inbox-article-card__actions` row which already uses `center`.

2. **Stack the row below 480px.** At narrow widths the fixed-width feedback form left the URL column ~120px wide, where `word-break: break-all` shredded a long UTM-laden unsubscribe URL into an unreadable single-character column. A new `@media (max-width: 480px)` block sets the row to `flex-direction: column` so the URL takes full width and the button sits beneath it, left-aligned. `gap: 0` because the button's 44px min-height already supplies vertical breathing room and tap separation. 480px is the breakpoint web-shell's base styles already use; this is the first media query in the inbox project and sets that precedent.

## Diff

```css
 .inbox-excluded-link {
   display: flex;
-  align-items: baseline;
+  align-items: center;
   justify-content: space-between;
   gap: 12px;
   padding: 6px 0;
 }

+@media (max-width: 480px) {
+  .inbox-excluded-link {
+    flex-direction: column;
+    align-items: flex-start;
+    gap: 0;
+  }
+}
```

No markup, template, TypeScript, or color changes. No new class names, so the unused-css check sees the same selector set.

## Verification

- `pnpm nx run inbox:check` and the full-monorepo `pnpm check` pass (100% coverage unchanged; `.css` files are not instrumented, and the JSDOM route tests assert rendered HTML, not computed layout, so no test surface changes).
- Visually verified against the dev server (`server.main.ts`, digest email → excluded tab) with the seed URL temporarily lengthened to ~120 chars of UTM query string, via Playwright at both viewports:
  - **481px**: row is side-by-side, info block vertically centered against the button (measured center-to-center delta 0px, vs. the previous baseline offset); URL starts at the row top with no dead space.
  - **320px**: row stacks — URL wraps full-width and readable, button directly below it left-aligned.

---
_Generated by [Claude Code](https://claude.ai/code/session_016GoQqHhg4RtCZYwKcXNTd6)_

### DIFF
--- a/projects/inbox/src/runtime/web/pages/inbox/inbox-email-detail.styles.css
+++ b/projects/inbox/src/runtime/web/pages/inbox/inbox-email-detail.styles.css
@@ -255,7 +255,7 @@
 
 .inbox-excluded-link {
   display: flex;
-  align-items: baseline;
+  align-items: center;
   justify-content: space-between;
   gap: 12px;
   padding: 6px 0;
@@ -281,4 +281,12 @@
 
 .inbox-excluded-link__feedback {
   flex-shrink: 0;
+}
+
+@media (max-width: 480px) {
+  .inbox-excluded-link {
+    flex-direction: column;
+    align-items: flex-start;
+    gap: 0;
+  }
 }


FagnerMartinsBrack:

## Context
- **PR**: #1001
- **Commit**: 2ef5baa03c0f8a2ab7e320a3880a7e287aa142cc

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `justify-content: space-between` is inert in the stacked rule

The new `@media (max-width: 480px)` block flips the row to `flex-direction: column` but leaves `justify-content: space-between` inherited from the base rule. With the container at `height: auto`, there is no free space on the now-vertical main axis, so `space-between` does nothing in column mode. It is harmless, but explicitly resetting it (e.g. `justify-content: flex-start`) would make the stacked rule self-contained.

### 2. `gap: 0` depends implicitly on the button's 44px `min-height`

The stacked layout uses `gap: 0` and relies on the button's `min-height: 44px` (with `line-height: 1`, ~12px text centered ⇒ ~16px above the label) to separate the reason text from the button. Sound today, but the spacing is implicit: if the button's `min-height` is ever lowered, the reason text and button label would sit flush. A small explicit `gap` (e.g. `6px`) would document the intent. Given 44px is a WCAG tap-target floor, the current choice is defensible — optional.

FagnerMartinsBrack:
@claude CSS-only change to `.inbox-excluded-link` in `inbox-email-detail.styles.css`: replaces `align-items: baseline` with `center` (removing the ~10px dead space the 44px report button's baseline forced above the URL) and adds a `@media (max-width: 480px)` block that stacks the row so long `word-break: break-all` URLs wrap full-width. The diagnosis matches the markup (`__info` column plus 44px inline-flex button), the alignment mirrors the sibling `.inbox-article-card__actions` rule, and the 480px breakpoint follows web-shell's existing convention. No new selectors and no test surface, so coverage and unused-css are unaffected.

**Findings:** no critical, high, or medium issues.

1. Inert `justify-content: space-between` in the stacked rule — optional: it is a no-op in column layout with auto height, and adding a reset would be a purely decorative declaration; fine to leave as-is.
2. `gap: 0` relying on the button's 44px `min-height` — optional: the spacing intent is documented in the commit, 44px is a WCAG tap-target floor unlikely to shrink, and an explicit gap would alter the 320px layout the author visually verified.

PR #1002, opened 2026-07-23
fix(inbox): give each address row's controls a per-row accessible name

FagnerMartinsBrack:
## What & why

Every inbox address row exposed **identical accessible names** to assistive tech: the readonly address input was labelled `Inbox email`, the copy button `Copy`, and the disable button `Disable` — the same for all up to `INBOX_ADDRESS_MAX_PER_USER` (25) rows. A screen-reader user tabbing or listing form controls heard "Inbox email, edit text / Copy, button / Disable, button" repeated with nothing distinguishing `netflix` from `stratechery`, and Disable is destructive to a live forwarding pipeline. This fails **WCAG 2.4.6 (Headings and Labels)**.

The link cards on the same page already get this right (`ariaLabel: \`Save to queue: ${displayUrl}\``), so this change mirrors that pattern: labels are built in TS and bound in the template, rather than hardcoded per control in Handlebars.

## Changes

- **`inbox.viewmodel.ts`** — `InboxAddressRowViewModel` gains three required fields, populated in `toRow`:
  - `addressAriaLabel` → `Inbox email: <name>`
  - `copyAriaLabel` → `Copy inbox email: <name>`
  - `disableAriaLabel` → `Disable inbox email: <name>`
- **`inbox.template.html`** — attribute-only edits: the active input, copy button, disable button, and the disabled-group input now bind their per-row `aria-label`. Visible button text (`Copy`, `Disable`) is unchanged.
- No styles, colours, client JS, or route changes.

Each `aria-label` starts with the visible text, so **WCAG 2.5.3 (Label in Name)** is also satisfied. The `<name>` follows the existing "inbox email" noun and the card viewmodel's `<action>: <subject>` shape.

## Tests

- `inbox.viewmodel.test.ts` — existing `toRow` cases now assert the three new label fields on both active and disabled entries.
- `inbox.component.test.ts` — new test asserts `aria-label` on the copy button, disable button, and both address fields, and that two rows carry **distinct** field labels (the per-row distinctness is the point of the finding).
- Pure additive fields in an already fully covered function — coverage stays at 100%; `pnpm nx run inbox:check` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_014VcBSg36U3ub6trZRjHgVJ)_

### DIFF
--- a/projects/inbox/src/runtime/web/pages/inbox/inbox.component.test.ts
+++ b/projects/inbox/src/runtime/web/pages/inbox/inbox.component.test.ts
@@ -94,6 +94,45 @@ describe("InboxPage", () => {
 		);
 	});
 
+	it("gives each control in a row a per-row accessible name so screen-reader users can tell rows apart", () => {
+		const doc = parse(
+			InboxPage({
+				addresses: [
+					entry({
+						name: AliasNameSchema.parse("netflix"),
+						address: InboxAddressSchema.parse("netflix-def456@read.place"),
+						token: InboxTokenSchema.parse("def456"),
+					}),
+					entry({
+						name: AliasNameSchema.parse("stratechery"),
+						address: InboxAddressSchema.parse("stratechery-abc123@read.place"),
+						token: InboxTokenSchema.parse("abc123"),
+						disabledAt: "2026-06-22T00:00:00.000Z",
+					}),
+				],
+				limitReached: false,
+			}).content.html,
+		);
+
+		assert.equal(
+			doc.querySelector("[data-inbox-copy]")?.getAttribute("aria-label"),
+			"Copy inbox email: netflix",
+		);
+		assert.equal(
+			doc.querySelector("[data-test-inbox-disable]")?.getAttribute("aria-label"),
+			"Disable inbox email: netflix",
+		);
+
+		const [active, disabled] = Array.from(doc.querySelectorAll(".inbox__address-field"));
+		assert.equal(active.getAttribute("aria-label"), "Inbox email: netflix");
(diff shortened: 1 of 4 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1002
- **Commit**: b4f34958575a83e62b029b3129ce1135b4a03c2b

FagnerMartinsBrack:

FagnerMartinsBrack:
@claude Focused accessibility fix that does exactly what the title says: each inbox address row's input, Copy button, and Disable button now carry a per-row accessible name built in the viewmodel (`Inbox email: <name>`, `Copy inbox email: <name>`, `Disable inbox email: <name>`) and bound in the template, mirroring the existing link-card `ariaLabel` pattern. The labels are injection-safe (`AliasNameSchema` + Handlebars escaping), always populated, and prefixed with the visible text so Label in Name (WCAG 2.5.3) holds alongside the 2.4.6 fix. Tests pin both the viewmodel output and the rendered attributes for active and disabled rows.

**Findings**

- Critical/High/Medium: none.
- Low (from the earlier automated review, both fine to leave as-is):
  1. *Copy button's name no longer reflects the "Copied" flash* — optional; the flash was never in an `aria-live` region, so AT never announced it before this PR either. Exposing it properly is a separate `aria-live` change, not this PR's scope.
  2. *`assert.notEqual` on the two field labels is logically redundant* — optional; it cannot fail independently of the two preceding `assert.equal` calls, but it documents the per-row-distinctness intent with a message. Keep or drop is a taste call.

PR #1008, opened 2026-07-23
feat(inbox): let users re-enable a disabled inbox email

FagnerMartinsBrack:
## What & why

Disabling an inbox forwarding address was a one-click, one-way door. The random token in `netflix-<token>@read.place` is not reproducible, so recreating the same name mints a *different* address — every newsletter subscription and Gmail filter pointed at the old one silently breaks, with nothing in the product able to bring it back. The limit banner even steers at-cap users (the ones with the most subscriptions) straight at that unconfirmed button.

The fix is **undo instead of confirm**: disable stays one click (it is reversible and frees its cap slot instantly), and every disabled row now carries a first-class **Enable** action. No JS confirm dialog — the surface serves no client JS, and reversibility makes interruption unnecessary.

## Changes

- **`@packages/domain` — `InboxAddressStore`**: new required `enableAddress`, the ownership-guarded inverse of `disableAddress`. It clears `disabledAt`; a tombstoned row's sentinel `userId` fails the `userId = :uid` guard, so a deleted-account address can never be resurrected and a forged address can never reach a foreign row.
- **`@packages/inbox-store` (DynamoDB)**: `enableAddress` issues an ownership-guarded `REMOVE disabledAt` update (idempotent on an already-live row).
- **`@packages/test-fixtures` (in-memory)**: mirrors `disableAddress`'s semantics — throws `ConditionalCheckFailedException` when the row is missing or foreign-owned, else clears `disabledAt`.
- **`inbox.page.ts`**: new `POST /inbox/enable`, gated by `requireNotLocked` + `requireWriteAccess` like `/create` (it re-opens a mail-receiving input). It acts only on an owned, currently-disabled address; if re-enabling would exceed `INBOX_ADDRESS_MAX_PER_USER` live it redirects `?error=limit` and leaves the row disabled; invalid body / foreign / already-live fall through to a plain 303, matching `/disable`. `DisableAddressSchema` → `AddressActionSchema` (shared by both routes).
- **Template / component / CSS**: an Enable `<form>` on each disabled row, sharing the Disable button's muted `--muted-foreground`-on-`--card` styling via grouped selectors — a recovery action, deliberately not the amber primary treatment.

## Tests

- **`inbox.route.test.ts`** — new `POST /inbox/enable` suite: happy path (row returns to the active list, disabled group collapses), Enable control renders in the disabled row pointing at `/inbox/enable`, foreign address, invalid body, already-live no-op, at-cap refusal (`?error=limit`, row stays disabled), and both save gates (read-only → `/queue?inactive=1`, locked → account-locked screen; store untouched).
- **`dynamodb-inbox-address.test.ts`** — `enableAddress` issues the ownership-guarded `REMOVE`; propagates the conditional-check failure for a non-owner.
- **`in-memory-inbox-address.test.ts`** — clears `disabledAt`; rejects missing / foreign.
- **`receive-email-handler.test.ts`** — mail to a disabled-then-re-enabled address is delivered again (the "no way back" reversed).

## Verification

`pnpm check` — green across all 45 projects / 63 tasks, **100% coverage** on every changed inbox file (`inbox.page.ts`, `inbox.component.ts`, `receive-email-handler.ts`) and 100% on `@packages/domain`, `@packages/inbox-store`, and `@packages/test-fixtures`. No infrastructure touched.

https://claude.ai/code/session_01JMopAS6SXNiSaKVXjgmBDB

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMopAS6SXNiSaKVXjgmBDB)_

### DIFF
--- a/projects/inbox/src/runtime/domain/inbox/receive-email-handler.test.ts
+++ b/projects/inbox/src/runtime/domain/inbox/receive-email-handler.test.ts
@@ -212,6 +212,22 @@ describe("initReceiveEmailHandler", () => {
 		expect(published).toHaveLength(0);
 	});
 
+	it("delivers again to an address that was disabled and then re-enabled", async () => {
+		const { addressStore, emailStore, rawMap, published, run } = makeHarness();
+		const address = await mintAddress(addressStore);
+		await addressStore.disableAddress({ userId: OWNER, address });
+		await addressStore.enableAddress({ userId: OWNER, address });
+		rawMap.set(RAW_KEY, Buffer.from("raw"));
+
+		const result = await run(address);
+
+		assert(result);
+		expect(result.batchItemFailures).toHaveLength(0);
+		const [row] = await listEmails(emailStore, OWNER);
+		expect(row.status).toBe("received");
+		expect(published).toHaveLength(1);
+	});
+
 	it("records an unparseable email as status=unparsed and fails to the DLQ", async () => {
 		const { addressStore, emailStore, rawMap, published, run } = makeHarness({
 			parseEmail: async () => ({ ok: false, reason: "unparseable" }),
--- a/projects/inbox/src/runtime/web/pages/inbox/inbox.component.ts
+++ b/projects/inbox/src/runtime/web/pages/inbox/inbox.component.ts
@@ -27,6 +27,7 @@ export function InboxPage(params: {
 		}),
 		createAction: "/inbox/create",
 		disableAction: "/inbox/disable",
+		enableAction: "/inbox/enable",
 	});
 
 	return {
--- a/projects/inbox/src/runtime/web/pages/inbox/inbox.page.ts
+++ b/projects/inbox/src/runtime/web/pages/inbox/inbox.page.ts
@@ -71,9 +71,9 @@ interface InboxDependencies {
 	logError: (message: string, error?: Error) => void;
 	buildBannerState: BuildBannerState;
 	publishSubmitLink: (input: { userId: UserId; url: string }) => Promise<void>;
-	/** Save gates applied to the write actions — /create (a forwarding address is
-	 * a save-flow input) and the per-link save (it lands an article in the
-	 * reader's queue). Both gates run: `requireNotLocked` blocks a
+	/** Save gates applied to the write actions — /create and /enable (each opens
+	 * a mail-receiving save-flow input) and the per-link save (it lands an article
+	 * in the reader's queue). Both gates run: `requireNotLocked` blocks a
 	 * locked (unverified-past-window) account, `requireWriteAccess` blocks a
 	 * read-only (trial-expired / cancelled) account. Viewing and disabling existing
 	 * addresses stay open — disabling reduces footprint and is harmless. */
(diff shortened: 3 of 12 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1008
- **Commit**: 7bfe357d99ee0023c47bd29fd841023f3f54fcf5

FagnerMartinsBrack:
### Review: `feat(inbox): let users re-enable a disabled inbox email`

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. At-cap `?error=limit` banner copy is create-worded on the enable path
A refused enable redirects to `${addressesPath}?error=limit` (`inbox.page.ts:488`), rendering the shared banner in `inbox.template.html:30`: *"…Disable any you no longer need before creating more."* The action here was **enable**, not **create** — cosmetic only (the at-cap message and the disable-to-free-a-slot remedy are still correct). Consider making the copy action-neutral now that two routes feed this banner.

### 2. (Optional) The store-level "idempotent on an already-live row" claim isn't directly tested
The route guards with `!isLiveAddress(target)` so the store is never called on a live row, and the already-live no-op is covered at the route layer — so this is genuinely optional. A one-line store test asserting `enableAddress` on an owned, already-live row is a harmless no-op would pin the documented contract at the boundary.

FagnerMartinsBrack:

## Context
- **PR**: #1008
- **Commit**: 718d144dd4b7b47c86f49d89545fb70d7feafe64

FagnerMartinsBrack:
@claude Solid, focused change: `enableAddress` mirrors `disableAddress`'s ownership guard at both the route and store layers (a tombstoned row's sentinel userId fails the condition, so deleted-account addresses stay dead), the `/enable` route carries the same save gates as `/create`, the at-cap refusal redirects before mutating, and the receive-path test proves mail flows again after disable-then-enable. Test coverage across route, both store adapters, and the handler is thorough. No critical, high, or medium issues found.

**Low priority:**
- **Banner copy (existing review item 1) — worth addressing:** the `?error=limit` message is still create-worded ("…before creating more") when the refused action was an enable. One-line, action-neutral rewording. Note the copy has since moved on main into `inbox.viewmodel.ts` (`ALERT_MESSAGES.limit`) via the alerts refactor, so the fix belongs there now, not in the template line the review cited.
- **Store-level idempotence test (existing review item 2) — fine to leave:** the route's `!isLiveAddress` guard means the store is never called on a live row, and DynamoDB `REMOVE` of an absent attribute is inherently a no-op; the already-live path is covered at the route layer.
- **Cap check race — fine to leave:** the `/enable` cap check is route-layer only over the eventually-consistent list read, so a concurrent enable/create could briefly exceed the cap. This matches the documented best-effort stance of `/create`'s duplicate-name guard and the outcome is harmless.

Since the branch predates the alerts/viewmodel and icon refactors on main, the merged result was checked by hand: imports, selectors, `error=limit` handling, and test helpers are all still compatible.

**Rebased onto current main (718d144d)** — the alerts-to-viewmodel refactor, icon set, and saved-link changes all merged cleanly; inbox:check green locally and full PR CI green on the rebased head. **Addressed the LOW copy item on the rebased branch:** the at-cap alert now reads "before enabling or creating more" (ALERT_MESSAGES.limit), since the enable action can hit the cap too.

PR #1009, opened 2026-07-23
fix(@packages/web-shell,hutch,inbox): announce toasts via a persistent shell live region

FagnerMartinsBrack:
## Why

The save/report toast was a live region (`role="status"`) rendered **already populated**, which is the unreliable case for screen readers:

- On a **full document load**, a live region populated at parse time is never announced — universally.
- Inside an **htmx-swapped `<main>`**, a toast is announced by NVDA/JAWS in many configurations but not dependably by VoiceOver/Safari.
- On the **inbox detail page** the toast was also the *last* child of `<main>`, so a no-JS reader met the save/report confirmation dead last in reading order.

## What changed

- **`base.template.ts`** — adds one persistent, parse-time-**empty** `#toast-live-region` (`role="status"`, `aria-live="polite"`, `.sr-only`) between `{{{content}}}` and `{{{footer}}}`, i.e. **outside `<main>`** so no `hx-target="main"` swap ever replaces the node the screen reader is watching.
- **`toast.template.ts`** — removes `role="status" aria-live="polite"` from `.toast`. Announcement now flows through one reliable channel; keeping both would double-announce on NVDA.
- **`toast.client.ts`** — the already-global toast client (scans `[data-dismiss]` on load and every `htmx:afterSwap`) now, on first sight of a toast, writes its `.toast__message` text into the region after a short **150ms settle delay** (a region mutated in the same task as the insertion is the unreliable case being fixed), and **clears** the region on dismiss so a second identical save re-announces. If the region is absent, announcement is skipped and dismissal is unchanged.
- **`inbox-email-detail.template.html`** — moves `{{{statusToastHtml}}}` to directly after `</header>` (before the tab strip), matching the queue's early placement. `.toast` is `position: fixed`, so the visual is unchanged; only the no-JS reading order improves.

No new client script and no inbox wiring changes — this rides the existing global `toast.client.js` bundle, which hutch already serves on every page.

## Decisions

- **Removed `role`/`aria-live` from `.toast`** rather than keeping it: the region announces on every path the toast ever did, plus the ones it didn't, and a self-announcing toast would double-speak once the region ships.
- **150ms settle delay** over a zero-delay same-tick write: the same-tick write is the unreliability being fixed, and 150ms is imperceptible next to the 6000ms dismiss.

## Tests

- `toast.component.test.ts` — the rendered `.toast` carries no `role`/`aria-live`.
- `base.component.test.ts` — every page renders exactly one empty `#toast-live-region` with `role="status"` / `aria-live="polite"`, outside `<main>`.
- `toast.client.test.ts` — message lands in the region only after the settle callback fires; dismiss clears it; a second identical toast re-announces; a swapped-in toast announces; a page with no region still schedules dismissal without throwing.
- `inbox-email-detail.route.test.ts` — DOM-order assertion (`compareDocumentPosition`) that the confirmation toast precedes the active tab panel.
- `inbox-link-save.route.test.ts` — swept: the confirmation toast now carries no `role`/`aria-live`.

The queue surface (hutch) picks up the same shell behaviour with no template change. `pnpm nx run` `@packages/web-shell:check`, `hutch:check`, `inbox:check` (plus `blog-site`/`web-embed`, which share the shell) all pass at 100% coverage, and the pre-commit full `pnpm check` is green.

## Follow-up not included

The plan flagged widening the `toast.client.ts` module-header comment and the build-script footer comment; per the repo comment policy those are left for explicit human approval. The existing comments remain accurate (the defer/swap wiring is unchanged), just narrower in scope. Happy to add them on request.

https://claude.ai/code/session_012mxZ6RFTY5fLfXnPYxHBMJ

---
_Generated by [Claude Code](https://claude.ai/code/session_012mxZ6RFTY5fLfXnPYxHBMJ)_

### DIFF
--- a/projects/hutch/src/runtime/web/shared/toast/toast.client.test.ts
+++ b/projects/hutch/src/runtime/web/shared/toast/toast.client.test.ts
@@ -7,12 +7,18 @@ interface ScheduledTimer {
 	ms: number;
 }
 
-function initWithDom(bodyHtml: string): {
+function initWithDom(
+	bodyHtml: string,
+	opts: { withRegion?: boolean } = {},
+): {
 	document: Document;
 	timers: ScheduledTimer[];
 	triggerSwap: () => void;
 } {
-	const dom = new JSDOM(`<!DOCTYPE html><html><body>${bodyHtml}</body></html>`);
+	const region = opts.withRegion
+		? `<div class="sr-only" id="toast-live-region" role="status" aria-live="polite"></div>`
+		: "";
+	const dom = new JSDOM(`<!DOCTYPE html><html><body>${bodyHtml}${region}</body></html>`);
 	const timers: ScheduledTimer[] = [];
 	let swapListener: (() => void) | undefined;
 	initToastDismiss({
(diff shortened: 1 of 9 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1009
- **Branch**: claude/toast-live-region-placement-6paekj

FagnerMartinsBrack:

## Context
- **PR**: #1009
- **Commit**: 3716259b640ac179ebf4df7be5256c78c0897fa8

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Overlapping toasts within the 6s dismiss window share one region — two benign edge cases

The single persistent region plus the clear-on-dismiss are what make identical re-saves re-announce (region goes `""` → `"Saved"` again), which is a nice touch. Both effects only bite when a *second* toast appears before the first's `data-dismiss` window elapses:

- **Stale clear:** toast A's dismiss timer (`toast.client.ts:39-43`) runs `region.textContent = ""` at A-insert + 6000ms even after a swap replaced A with toast B. B was already announced at its own 150ms settle, so nothing is mis-spoken — the region just empties a few seconds early for anyone who navigates back to it.
- **Identical within-window:** two identical messages < 6s apart — the region still holds the first's text, so the second write is a no-op change and `aria-live` stays silent.

Both are inherent `aria-live` trade-offs and harmless for today's single-confirmation flows (save/report round-trip through a form post + swap, so two-within-6s needs deliberate rapid clicks). Not worth changing now; flagging only so the boundary is a conscious choice. If rapid/duplicate toasts ever become common, clearing-then-writing on a microtask inside `announce` would cover both.

FagnerMartinsBrack:

## Context
- **PR**: #1009
- **Commit**: 382cc27b86e8d89d0f87dd9200f0c1ab7dfb417a

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `data-test-toast-live-region` is unreferenced — dead markup

`base.template.ts:128` adds `data-test-toast-live-region` to the region div, but nothing keys off it:

- The client reads the region by **id** — `getElementById("toast-live-region")` (`toast.client.ts:27`).
- Every test also queries by **id** — `base.component.test.ts:257` uses `#toast-live-region`, and the `toast.client.test.ts` fixtures build the region with only `id="toast-live-region"` (no `data-test` attr).

Testing by `id` here is actually the *better* choice — it couples the assertion to the exact contract the client depends on, so a rename of the id can't pass a green test. That leaves the `data-test-*` attribute with no consumer. There's no automated unused-`data-test` check, so this won't fail CI, but the repo otherwise keeps markup free of dead hooks. Either drop the attribute or intentionally point a test at it. [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Ftoast-live-region-placement-6paekj%3A%20remove%20the%20unused%20data-test-toast-live-region%20attribute%20from%20the%20%23toast-live-region%20div%20in%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.template.ts%20(line%20128).%20No%20test%20or%20client%20references%20it%20%E2%80%94%20base.component.test.ts%20queries%20%23toast-live-region%20and%20toast.client.ts%20uses%20getElementById(%22toast-live-region%22).%20Then%20run%20pnpm%20nx%20run%20%40packages%2Fweb-shell%3Acheck.&repo=Readplace/readplace.com)

### 2. Announcement is gated by auto-dismiss (`ms <= 0` skips the announce)

In `scheduleToast` (`toast.client.ts:45-53`), `announce(toast)` is called *after* the early returns for a non-finite or non-positive `data-dismiss`:

```ts
if (!Number.isFinite(ms)) return;
if (ms <= 0) return;      // returns before announce()
scheduled.add(toast);
deps.setTimeoutFn(() => dismiss(toast), ms);
announce(toast);
```

So a toast that opts out of auto-dismiss (`data-dismiss="0"` / missing) would also silently opt out of being announced. Every toast today renders a positive `dismissMs` from `toast.template.ts`, so this never triggers — flagging only so the coupling of "announce" to "auto-dismiss" is a conscious choice. If a persistent (non-dismissing) toast is ever introduced, it would need `announce` lifted above the `ms <= 0` gate.

### 3. Overlapping toasts within the 6s window share the single region

Two inherent single-region trade-offs, both harmless for today's single-confirmation flows (save/report round-trip through a form POST + swap, so two-within-6s needs deliberate rapid clicks). Already raised on the prior review pass; restating for completeness since this is a fresh review of the rebased commit:

- **Stale clear:** toast A's dismiss timer runs `region.textContent = ""` at A-insert + 6000ms (`toast.client.ts:41`) even after a swap replaced A with toast B. B was already announced at its own 150ms settle, so nothing is mis-spoken — the region just empties early for anyone who reaches it via the virtual cursor afterward.
- **Identical within-window:** two identical messages < 6s apart — the region still holds the first's text, so the second write is a no-op change and `aria-live` stays silent.

Not worth changing now. If rapid/duplicate toasts ever become common, clearing-then-writing on a microtask inside `announce` would cover both.

FagnerMartinsBrack:

## Context
- **PR**: #1009
- **Commit**: 5ecc753648536769b2d28bf495445f856dc234d1

FagnerMartinsBrack:
@claude Solid accessibility fix. The persistent parse-time-empty `#toast-live-region` outside `<main>` is the reliable live-region pattern for both full loads and `hx-target="main"` swaps, the 150ms settle write avoids the same-task mutation being fixed, and removing `role`/`aria-live` from the toast itself correctly prevents double-announcement. Test coverage is thorough: announce-after-settle, clear-on-dismiss, identical re-announce, swap arrival, missing-region no-throw, and the inbox DOM-order assertion via `compareDocumentPosition`.

**Findings**

- No critical, high, or medium issues found. Conventions check out: no `as` casts, the local `assert` helper matches the existing pattern in six other `*.client.ts` bundle files, comments are why-only, and both region-present/absent branches are exercised.

**Existing low-priority review items**

1. **Unused `data-test-toast-live-region` attribute** (`base.template.ts`) — *address*: dead markup with no consumer; client and tests key off the id, so the attribute can simply be dropped.
2. **`announce` gated behind the `ms <= 0` early return** — *leave as-is*: no non-dismissing toast exists, and lifting `announce` above the gate would let a never-scheduled toast re-announce on every `htmx:afterSwap` rescan (it is never added to the `scheduled` WeakSet), so the current placement is the safer one.
3. **Overlapping toasts within the 6s window share the region** (stale clear / identical-within-window silence) — *leave as-is*: inherent single-region trade-off, unreachable in today's single-confirmation flows; revisit with the suggested clear-then-write microtask only if rapid duplicate toasts become a real path.

**Addressed in 5ecc7536:** the LOW dead-markup item — removed the unused `data-test-toast-live-region` attribute; the client and every test reach the region by id.

PR #1011, opened 2026-07-23
perf(@packages/article-store): drop the redundant existence GetItem from status toggles and deletes

FagnerMartinsBrack:
## What

`updateArticleStatus` and `deleteArticle` in `dynamodb-saved-article-store.ts` each ran a sequential three-hop DynamoDB chain on the queue's hottest mutations:

`findArticleByRouteId` (routeId GSI Query) → `findUserArticle` (**GetItem**) → `UpdateItem` / `DeleteItem`

The middle **GetItem** existed only to answer "does this user have this article?" for the boolean return — the row's contents were never read. This replaces it with `ConditionExpression: "attribute_exists(savedAt)"` on the write itself, mapping `ConditionalCheckFailedException` to the existing `false` return.

## Why

- **Latency** — removes one sequential in-region round trip (~5–15ms) from the POST leg of every "Mark as read/unread" and "Delete", cutting the store calls on that leg from 3 to 2. One increment of the CTA-latency programme, not its resolution (the dominant cost remains the 303 → full `/queue` re-render).
- **Race correctness** — previously, if the row was deleted between the GetItem and the UpdateItem, DynamoDB's upsert semantics re-created a phantom `{userId, url, status, readAt?}` row with no `savedAt`. That phantom is invisible to the queue (the `userId-savedAt-index` GSI is sparse on `savedAt`) but surfaces in `url-index`, where `UserArticleRow.parse` throws on the missing required `savedAt` — breaking `findUserArticlesByUrl` and `findArticleById`. The condition turns that race into a clean `false`.

`attribute_exists(savedAt)` is the file's established row-exists idiom (`stampUserArticleIfStillSaved`, `markReaderReadyEmailSent`); `savedAt` is required by `UserArticleRow` on every read path, so the condition cannot newly exclude any functioning row.

## Contracts unchanged

`DeleteArticle` / `UpdateArticleStatus` stay `(…) => Promise<boolean>` — only how the boolean is computed changes. No provider-contract, caller, or Siren-shape edits. `findUserArticle` is retained (still used by `findArticleById`). The in-memory store already had these exact semantics, so its integration tests are untouched.

**Contract note:** the loser of a concurrent double-`deleteArticle` now returns `false` (was `true`). Strictly more truthful, and both HTML callers ignore the boolean.

## Tests

- Rewrote the "reports false when the user never saved the article" tests for both functions to drive the conditional-failure path via the fake client's error-injection seam.
- Added a `rethrows non-conditional errors` test to each `describe` block (covers the `throw error` branch).
- Read/unread tests now assert the new `ConditionExpression` and that no `GetCommand` is sent.
- `@packages/article-store:check` green with 100% statements/branches/functions/lines; full `pnpm check` green.

https://claude.ai/code/session_01Ri7Byha9DajrK7u6sDyHdt

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ri7Byha9DajrK7u6sDyHdt)_

### DIFF
--- a/src/packages/article-store/src/dynamodb-saved-article-store.test.ts
+++ b/src/packages/article-store/src/dynamodb-saved-article-store.test.ts
@@ -730,12 +730,13 @@ describe("initDynamoDbSavedArticleStore deleteArticle", () => {
 	it("deletes the user row and reports success", async () => {
 		const { client, commands } = createFakeClient({
 			QueryCommand: { default: { Items: [articleItem()], Count: 1 } },
-			GetCommand: { default: { Item: userArticleItem() } },
 		});
 
 		const deleted = await initStore(client).deleteArticle(ReaderArticleHashId.fromHash(ROUTE_ID), USER);
 
-		expect(commands.some((c) => c.name === "DeleteCommand")).toBe(true);
+		const del = commands.find((c) => c.name === "DeleteCommand");
+		expect(del?.input.ConditionExpression).toBe("attribute_exists(savedAt)");
+		expect(commands.some((c) => c.name === "GetCommand")).toBe(false);
 		expect(deleted).toBe(true);
 	});
 
@@ -750,12 +751,31 @@ describe("initDynamoDbSavedArticleStore deleteArticle", () => {
 	it("reports false when the user never saved the article", async () => {
 		const { client } = createFakeClient({
 			QueryCommand: { default: { Items: [articleItem()], Count: 1 } },
-			GetCommand: { default: { Item: undefined } },
+			DeleteCommand: {
+				default: () => {
+					throw new ConditionalCheckFailedException({ $metadata: {}, message: "row deleted" });
+				},
+			},
 		});
 
 		const deleted = await initStore(client).deleteArticle(ReaderArticleHashId.fromHash(ROUTE_ID), USER);
 
 		expect(deleted).toBe(false);
+	});
+
+	it("rethrows non-conditional errors", async () => {
+		const { client } = createFakeClient({
+			QueryCommand: { default: { Items: [articleItem()], Count: 1 } },
+			DeleteCommand: {
+				default: () => {
+					throw new Error("throttled");
+				},
+			},
+		});
+
+		await expect(
+			initStore(client).deleteArticle(ReaderArticleHashId.fromHash(ROUTE_ID), USER),
+		).rejects.toThrow("throttled");
 	});
 });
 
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1011
- **Commit**: 04f25c645e3143458108c8e919e0643fddcb7760

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. PR description under-documents the `updateArticleStatus` behavior change
The **Contract note** only calls out `deleteArticle`, but `updateArticleStatus` changes its return in the *same* concurrent-delete race — and unlike the two delete callers, its caller **does consume the boolean**: `queue.page.ts:1335-1349` uses `updated` to push the `status_changed`/`status_article` flash params and to emit the `article_read` analytics event. This is not a bug — the new behavior is strictly more correct (a mid-toggle deleted article yields `false`, so no misleading "marked as read" toast, no spurious analytics event, no resurrected phantom row). The false path is already covered by the route test at `queue.mutations.route.test.ts:472`. Worth one sentence in the description so the caller-visible change is on record.

### 2. (Optional) Pin the delete-race fix with an intent-revealing test
The correctness win — turning the mid-write phantom into a clean `false` — is exercised only implicitly by the reused `reports false when the user never saved the article` tests (which just inject a `ConditionalCheckFailedException`). A dedicated test named for the race would guard the regression you're closing by intent rather than shared branch coverage.

FagnerMartinsBrack:
@claude ### Review summary

Focused, well-tested change: the existence `GetItem` in `updateArticleStatus` and `deleteArticle` is replaced by `ConditionExpression: "attribute_exists(savedAt)"` on the write itself, mapping `ConditionalCheckFailedException` to the existing `false` return. This trims one sequential round trip from the queue's hottest mutations and closes the write-after-delete race that could resurrect a phantom `savedAt`-less row on the `url-index` read path. The idiom matches the file's established pattern (`stampUserArticleIfStillSaved`, `markReaderReadyEmailSent`), the `Promise<boolean>` contracts are unchanged, and the success, conditional-failure, and rethrow paths are all covered, with tests asserting no `GetCommand` is sent. CI is green on the head commit and the branch merges cleanly — main has not touched these files since the branch point.

**Findings:** no critical, high, or medium issues.

1. *PR description under-documents the `updateArticleStatus` behavior change* — **address**: fold the one-sentence contract note (a mid-toggle concurrent delete now yields `false`, so no `status_changed` flash and no `article_read` event) into the squash commit body; no code change needed, and the false path is already covered by `queue.mutations.route.test.ts`.
2. *Pin the delete-race fix with an intent-revealing test* — **optional**: the `ConditionalCheckFailedException`-injection tests already exercise the exact branch; a race-named duplicate would document intent but add no new coverage, fine to leave as-is.

PR #1024, opened 2026-07-30
[DUMMY] broken-toolchain visibility probe — do not merge

FagnerMartinsBrack:
Deliberately unresolvable packageManager (pnpm@99.99.99) so the listener fails at Setup pnpm. Verifies the failure now gets reported on the PR instead of dying silently, and that crash-retry no longer re-runs it. Will be closed after verification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "nx": "22.7.5",
     "typescript": "6.0.3"
   },
-  "packageManager": "pnpm@10.28.0",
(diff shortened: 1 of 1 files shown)


FagnerMartinsBrack:
@claude Reply with the single word done. Make no other changes: do not edit files, do not run commands, do not commit, do not push.

FagnerMartinsBrack:
Verification complete — closing this dummy PR. See the run links in the thread for the evidence.

PR #1025, opened 2026-07-30
fix(hutch): make the Reader Opens widget match the real reader permalink

FagnerMartinsBrack:
The "Reader Opens per Day" widget has never plotted a point.

Its filter is `path like /^\/[^\/]+\/view$/` — a **two-segment** pattern — but the authenticated reader permalink is `/queue/<id>/view`, **three** segments. It could never match.

## Evidence

Run against prod `/readplace/analytics` over 40 days:

| Query | Matched records |
|---|---|
| the widget's current filter | **0** |
| same query, pattern corrected | **120** |

With the fix the widget plots 1–7 authenticated reader opens per day (3, 5, 4, 7, 5, 1, 3, 3, 3, 1, 3, 1 across Jul 18–30).

## Change

- The pattern now derives from `QUEUE_PATH` — the constant the path itself is built from — so it cannot drift from the mount point again.
- The existing test asserted only that the query string *contained* `\/view$`, which is exactly why a never-matching pattern shipped. It now builds the emitted regex and runs it against a real permalink, plus two negative cases. Confirmed the new test fails against the old pattern.
- Drops the comment above the view-funnel widgets claiming "the pageview middleware cannot see /view". Production holds 1,863 `/view/*` article-path pageviews in the last 6 days, and that false belief is why the Jul 24 browser-coherence gate was never extended to `view_opened` until d4df6f75.

## Verification

- `pnpm check` — green (47 projects, 100% coverage).
- `pnpm check-infra` — the dashboard diff is exactly `~dashboardBody`; no replace or delete anywhere in the preview.

Split out from d4df6f75 per the "Unrelated Small Changes Ship as Their Own PR" rule: that commit is write-path analytics gating, this is read-path dashboard correctness, and only this half touches infra.

### DIFF
--- a/projects/hutch/src/runtime/observability/analytics-dashboard.test.ts
+++ b/projects/hutch/src/runtime/observability/analytics-dashboard.test.ts
@@ -1,9 +1,11 @@
+import assert from "node:assert/strict";
 import {
 	BLOG_SITE_LOG_GROUP,
 	FORWARD_ANALYTICS_FUNCTION_NAME,
 	SAVE_LINK_LOG_GROUPS,
 } from "@packages/hutch-infra-components";
 import { campaignTag, HOMEPAGE_SPLIT } from "../web/experiments/homepage-split";
+import { QUEUE_PATH } from "../web/pages/queue/queue.url";
 import {
 	ANALYTICS_EVENTS,
 	ANALYTICS_LOG_GROUP,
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1025
- **Commit**: 4ee3a1e0b106dd6bacd426513e5bd4de791cdb03

FagnerMartinsBrack:

PR #1036, opened 2026-08-02
feat: serve page CSS as hashed same-origin files instead of inlining it

FagnerMartinsBrack:
## Summary

Stops re-sending ~15–19 KB of inline page CSS on every hutch HTML response **and on every boosted `<main>` swap** for the measured CTA surfaces. Page CSS is now delivered as a content-hashed, same-origin, immutable-cached stylesheet linked from inside `<main>`, instead of an inline `<style>` block.

This is **Phase A, wave 1** (queue, reader, account, import, home). It is deliberately low-risk and complementary to the server-side latency work — it does **not** itself fix the CTA latencies (those are DynamoDB/banner-fetch, server-side, and belong to other plans).

## What changed

**`@packages/web-shell`**
- `PageBody.styles` becomes a union `string | { href: string }` — a union (not an extra optional field) so "both inline and href" is unrepresentable. Every existing consumer (blog-site, extensions, …) keeps passing strings and compiles unchanged.
- `injectPageStylesIntoMain` emits a `<link rel="stylesheet">` as the first child of `<main>` for the href variant (a `<style>` for the string variant, unchanged). The string-insert mechanism (not `parseHTML`) is preserved so the reader `srcdoc`'s intentional double-escaping survives byte-for-byte.
- `pageStylesheetPreload` feeds a matching `<link rel="preload" as="style">` into `<head>` via a new slot in both the base and chromeless shells, placed **before** the third-party font preloads so the preload scanner queues the render-critical same-origin CSS first.

**`hutch`**
- New `page-stylesheets.ts`: `addPageStylesheet({ name, css })` content-hashes the CSS into `/styles/<name>.<sha256-12>.css` at module load; `findPageStylesheetByName` looks it up.
- New `GET /styles/:file` route mounted **before** session/auth, serving `text/css` with `Cache-Control: public, max-age=31536000, immutable`. Lookup is **by name, ignoring the hash**, so a mixed-fleet deploy window serves current CSS rather than a 404/unstyled page. Unknown name or malformed filename → terminal 404.
- `/styles` is exempted from click attribution alongside `/client-dist`.
- Wave-1 pages register their composed CSS and pass `{ href }` instead of inlining.

## Why these design choices

- **Page CSS stays inside `<main>`** so boosted swaps replace styles atomically with content; a `<head>` link would break cross-page swaps (reader → queue mark-read).
- **Lookup by name, hash ignored** — a mixed-fleet deploy window serves current CSS instead of a 404/unstyled page.
- **Runtime hashing, no build step** — the composition lives in runtime TS (reader concatenates 3 sources, queue appends onboarding), so a build step would replicate it and drift.
- **Same-origin, not CDN** — per-commit assets must ship with the server that renders the HTML referencing them.

## Test plan

- Unit: `injectPageStylesIntoMain` / `pageStylesheetPreload` (both variants, escaping intact), `page-stylesheets` (href determinism, name lookup, collision), base + chromeless shells (link in `<main>` + head preload).
- Integration `styles.route.test.ts`: 200 + `text/css` + immutable header; stale hash still serves current CSS; unknown name / non-hashed filename → 404; `/queue` references the exact served href.
- Extended `static-asset-paths`, `queue.listing.route`, and `account.route` tests.
- `pnpm check` passes across all 47 projects (compile, lint, tests, 100% coverage, knip, unused-css).

## Scope / expected impact

Wave-1 pages only. Expected: decoded payload per CTA ~131 KB → ~87 KB, gzip transfer ~21 KB → ~12–13 KB, and elimination of a ~19 KB CSS parse + style-recalc on every boosted swap (single-digit-to-tens ms). Does **not** address the 900–2400 ms server-side CTA latencies. Wave 2 (the remaining ~22 producer sites, mechanical) is a separate follow-up once wave 1 is verified in prod.

https://claude.ai/code/session_0111DWvSaQrpKSpGMZ5X4cEQ

---
_Generated by [Claude Code](https://claude.ai/code/session_0111DWvSaQrpKSpGMZ5X4cEQ)_

### DIFF
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -154,7 +154,8 @@ import {
 	initEmitSubscriptionEvent,
 	type SubscriptionLogEvent,
 } from "./observability/subscription-events";
-import { APPLE_TOUCH_ICON_PATH, CLIENT_DIST_MOUNT_PATH, isStaticAssetRequestPath } from "./web/static-asset-paths";
+import { APPLE_TOUCH_ICON_PATH, CLIENT_DIST_MOUNT_PATH, STYLES_MOUNT_PATH, isStaticAssetRequestPath } from "./web/static-asset-paths";
+import { findPageStylesheetByName } from "./web/page-stylesheets";
 import { canonicalizeViewLandingPath } from "./web/pages/view/view-path";
 import { initGoogleAuthRoutes } from "./web/auth/google-auth.page";
 import { initAppleAuthRoutes } from "./web/auth/apple-auth.page";
@@ -497,6 +498,18 @@ export function createApp(dependencies: AppDependencies): Express {
 			fallthrough: false,
 		}),
 	);
+
+	app.get(`${STYLES_MOUNT_PATH}/:file`, (req: Request, res: Response) => {
+		const match = /^([a-z0-9-]+)\.[a-f0-9]{12}\.css$/.exec(String(req.params.file));
+		const stylesheet = match ? findPageStylesheetByName(match[1]) : undefined;
+		if (!stylesheet) {
+			res.status(404).end();
+			return;
+		}
+		res.set("Content-Type", "text/css; charset=utf-8");
+		res.set("Cache-Control", "public, max-age=31536000, immutable");
+		res.send(stylesheet.css);
+	});
 
 	app.use(contentSignalMiddleware);
 	app.use(linkHeaderMiddleware);
--- a/projects/hutch/src/runtime/server.ts
+++ b/projects/hutch/src/runtime/server.ts
@@ -500,14 +500,19 @@ export function createApp(dependencies: AppDependencies): Express {
 	);
 
 	app.get(`${STYLES_MOUNT_PATH}/:file`, (req: Request, res: Response) => {
-		const match = /^([a-z0-9-]+)\.[a-f0-9]{12}\.css$/.exec(String(req.params.file));
+		const file = String(req.params.file);
+		const match = /^([a-z0-9-]+)\.[a-f0-9]{12}\.css$/.exec(file);
 		const stylesheet = match ? findPageStylesheetByName(match[1]) : undefined;
 		if (!stylesheet) {
 			res.status(404).end();
 			return;
 		}
+		const requestedHrefIsCurrent = stylesheet.href === `${STYLES_MOUNT_PATH}/${file}`;
 		res.set("Content-Type", "text/css; charset=utf-8");
-		res.set("Cache-Control", "public, max-age=31536000, immutable");
+		res.set(
+			"Cache-Control",
+			requestedHrefIsCurrent ? "public, max-age=31536000, immutable" : "public, max-age=60",
+		);
 		res.send(stylesheet.css);
 	});
 
--- a/projects/hutch/src/runtime/web/page-stylesheets.test.ts
+++ b/projects/hutch/src/runtime/web/page-stylesheets.test.ts
@@ -1,0 +1,38 @@
+import assert from "node:assert/strict";
+import { createHash } from "node:crypto";
+import { addPageStylesheet, findPageStylesheetByName } from "./page-stylesheets";
+
+describe("page stylesheets", () => {
+	it("derives a hashed, same-origin href from the name and css", () => {
+		const css = ".a { color: red; }";
+		const stylesheet = addPageStylesheet({ name: "alpha", css });
+		const expectedHash = createHash("sha256").update(css).digest("hex").slice(0, 12);
+
+		assert.equal(stylesheet.css, css);
+		assert.equal(stylesheet.href, `/styles/alpha.${expectedHash}.css`);
+		assert.match(stylesheet.href, /^\/styles\/alpha\.[a-f0-9]{12}\.css$/);
+	});
+
+	it("produces the same href for identical css and a different href when css changes", () => {
+		const first = addPageStylesheet({ name: "beta", css: ".b { color: red; }" });
+		const recomputed = createHash("sha256").update(".b { color: red; }").digest("hex").slice(0, 12);
+		assert.equal(first.href, `/styles/beta.${recomputed}.css`);
+
+		const differentCssHash = createHash("sha256").update(".b { color: blue; }").digest("hex").slice(0, 12);
+		assert.notEqual(recomputed, differentCssHash);
+	});
+
+	it("looks a registered stylesheet up by name", () => {
+		const stylesheet = addPageStylesheet({ name: "gamma", css: ".c {}" });
+		assert.deepEqual(findPageStylesheetByName("gamma"), stylesheet);
+	});
+
+	it("returns undefined for an unregistered name", () => {
+		assert.equal(findPageStylesheetByName("never-registered"), undefined);
+	});
+
+	it("rejects registering the same name twice", () => {
+		addPageStylesheet({ name: "delta", css: ".d {}" });
+		assert.throws(() => addPageStylesheet({ name: "delta", css: ".d-again {}" }));
+	});
+});
--- a/projects/hutch/src/runtime/web/page-stylesheets.ts
+++ b/projects/hutch/src/runtime/web/page-stylesheets.ts
@@ -1,0 +1,24 @@
+import assert from "node:assert";
+import { createHash } from "node:crypto";
+
+export interface PageStylesheet {
+	href: string;
+	css: string;
+}
+
+const stylesheetsByName = new Map<string, PageStylesheet>();
+
+export function addPageStylesheet(entry: { name: string; css: string }): PageStylesheet {
+	assert(
+		!stylesheetsByName.has(entry.name),
+		`page stylesheet "${entry.name}" is already registered`,
+	);
+	const hash = createHash("sha256").update(entry.css).digest("hex").slice(0, 12);
+	const stylesheet: PageStylesheet = { href: `/styles/${entry.name}.${hash}.css`, css: entry.css };
+	stylesheetsByName.set(entry.name, stylesheet);
+	return stylesheet;
+}
+
+export function findPageStylesheetByName(name: string): PageStylesheet | undefined {
+	return stylesheetsByName.get(name);
+}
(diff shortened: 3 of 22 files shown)


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 30747239598
- **PR**: #1036
- **Branch**: claude/inline-css-bloat-6hnmqs

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:
Rebased onto `main` (`e54abc87`) and force-pushed with `--force-with-lease`.

No conflicts — the branch's commits replayed cleanly onto current `main`. No content was changed as part of the rebase.

FagnerMartinsBrack:

## Context
- **PR**: #1036
- **Branch**: claude/inline-css-bloat-6hnmqs
- **Attempt**: 1 of 5

FagnerMartinsBrack:
Re-rebased onto `main` (`5bfaefad`) after `main` advanced three commits, and force-pushed with `--force-with-lease`.

**Conflict resolved:** one, in `projects/hutch/src/runtime/web/pages/home/home.component.ts`. `main`'s `1fb8f536`/`5bfaefad` added a top-level `INSTALL_CTA_LABEL` const at the same position where this PR adds `HOME_STYLESHEET`. The two are independent declarations, so both are kept — `HOME_STYLESHEET` first, then `INSTALL_CTA_LABEL`. Neither side's change was dropped.

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **Run ID**: 30780580986
- **PR**: #1036
- **Branch**: claude/inline-css-bloat-6hnmqs

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #1036
- **Commit**: 6edf338f9ad17288d76943fcd83be7b44e0a9d9b

FagnerMartinsBrack:

None found

## Medium Priority Issues

### 1. `immutable` cache on a serve-by-name route can pin stale CSS well past the deploy window

The `GET /styles/:file` handler sends `Cache-Control: public, max-age=31536000, immutable` (`projects/hutch/src/runtime/server.ts:510`) but resolves the stylesheet **by name, ignoring the hash** in the URL (`server.ts:503-504`). Those two choices interact badly during the exact mixed-fleet window the design is meant to survive:

- A **new** server renders `/queue`; its HTML references `/styles/queue.<NEW>.css`.
- The browser's fetch for that URL is answered by a still-warm **old** container (or old Lambda version). It matches the name `queue`, finds *its own* (old) stylesheet, and returns the **old** bytes — under the `<NEW>` URL — with `immutable`.
- The browser now has old bytes pinned under the new-hash URL and, because of `immutable`, **will not revalidate**. The HTML keeps pointing at `<NEW>` until that page's CSS changes again, so the user is stuck on stale CSS *well beyond* the deploy window — not just during it.

So the design's stated goal — "serves current CSS rather than a 404/unstyled page" — isn't actually met in the cross-version case: it serves *a* valid stylesheet, but potentially the wrong version, and `immutable` makes that stick. Any shared cache in front of the origin amplifies this from per-browser to per-population.

The content-hash + `immutable` contract is normally safe *because* the server 404s unknown hashes, guaranteeing URL↔bytes. Serving-by-name deliberately breaks that guarantee to dodge 404s, but keeps `immutable` — voiding the invariant that makes `immutable` safe.

Surgical fix that keeps **both** properties (no 404 **and** self-healing): capture the requested hash and compare it to the served stylesheet's actual hash. Send `immutable` only on a match; on a mismatch, send a short `max-age` so the browser self-heals once the fleet settles.

```ts
const match = /^([a-z0-9-]+)\.([a-f0-9]{12})\.css$/.exec(String(req.params.file));
const stylesheet = match ? findPageStylesheetByName(match[1]) : undefined;
if (!stylesheet) { res.status(404).end(); return; }
const hashMatches = stylesheet.href.endsWith(`.${match[2]}.css`);
res.set("Content-Type", "text/css; charset=utf-8");
res.set("Cache-Control", hashMatches
  ? "public, max-age=31536000, immutable"
  : "public, max-age=60");
res.send(stylesheet.css);
```

If the stale window is judged acceptable instead, drop `immutable` outright (the per-view revalidation cost returns) — but `immutable` + serve-by-name *as written* is the combination to reconsider. [Fix this →](https://claude.ai/code?q=In%20projects%2Fhutch%2Fsrc%2Fruntime%2Fserver.ts%20the%20GET%20%2Fstyles%2F%3Afile%20route%20%28around%20lines%20502-512%29%20sends%20Cache-Control%20immutable%20but%20resolves%20the%20stylesheet%20by%20name%2C%20ignoring%20the%20hash%20in%20the%20URL.%20Capture%20the%2012-hex%20hash%20from%20the%20filename%20and%20send%20%27public%2C%20max-age%3D31536000%2C%20immutable%27%20only%20when%20it%20matches%20the%20served%20stylesheet%27s%20current%20hash%20%28stylesheet.href%20ends%20with%20%27.%3Chash%3E.css%27%29%3B%20otherwise%20send%20a%20short%20%27public%2C%20max-age%3D60%27%20so%20a%20stale%20cross-version%20response%20served%20during%20a%20mixed-fleet%20deploy%20self-heals%20instead%20of%20being%20pinned%20forever.%20Branch%20claude%2Finline-css-bloat-6hnmqs.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. First boosted swap to a not-yet-visited page type can flash unstyled

Moving page CSS from an inline `<style>` (which arrived atomically with a boosted `<main>` swap) to an external `<link>` inside `<main>` (`src/packages/web-shell/src/inject-page-styles.ts:11-20`) means the first boosted navigation to a page type the user hasn't loaded yet must fetch the stylesheet before it applies — a brief FOUC on the swapped content. The `<head>` preload (`base.template.ts:35`) only covers the *current* page, not a swap destination. Same-destination swaps (e.g. reader → queue when the user came from queue) are already cached, and `immutable` makes it once-per-user-per-page, so the blast radius is small — worth confirming during the wave-1 prod verification you've already planned.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #1036
- **Branch**: claude/inline-css-bloat-6hnmqs

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #1036
- **Commit**: 4de163fdabf130890e5f6eac35c4b09b1b5e8796

FagnerMartinsBrack:

## Context
- **PR**: #1036
- **Commit**: 8af779632b42bbbce7bd9cd201a8aea6f3423cc0

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found — the prior Medium (`immutable` + serve-by-name pinning stale CSS) remains correctly resolved at `8af77963`. Re-verified independently: `immutable` is sent only when `stylesheet.href === ${STYLES_MOUNT_PATH}/${file}` (`server.ts:510-515`), and because `href` is derived from `sha256(css)` at registration, that equality *is* a URL↔bytes guarantee. The cross-version deploy case serves the current CSS by name with `public, max-age=60`, so it self-heals within the deploy window instead of pinning stale bytes. Both ternary branches are exercised by `styles.route.test.ts`.

## Low Priority Suggestions

### 1. The removed `<main>`-not-`<head>` rationale was load-bearing — consider restoring it (or pinning it with a test)

This PR deleted the first paragraph of the doc comment on `injectPageStylesIntoMain` (`src/packages/web-shell/src/inject-page-styles.ts:1-10`), which explained *why* page CSS is injected **inside `<main>`** rather than `<head>`:

> *…so that htmx navigation (`hx-target="main" hx-select="main" hx-swap="outerHTML"`) swaps the page's CSS atomically with its content. Without this, htmx leaves the previous page's `<style>` stranded in `<head>`, defacing the new page's layout.*

That reasoning is still fully load-bearing under the new `<link>` variant — a stylesheet moved to `<head>` would be stranded on a boosted cross-page swap exactly as the old `<style>` would, and it's the linchpin of this PR's whole design ("Page CSS stays inside `<main>` … a `<head>` link would break cross-page swaps"). It's also precisely the non-obvious, external-constraint *why* the repo's comment policy says to keep even while removing *what/how* comments. With it gone, that knowledge now lives only in the PR description and git history, so a future "simplification" that hoists the link into `<head>` (the intuitive place for a stylesheet) would silently reintroduce the stranded-CSS regression. The `parseHTML`-vs-string-insert paragraph right below it was (rightly) kept for the same reason — this one deserves the same treatment. Restoring a one-line version covering both the `<style>` and `<link>` variants, or encoding the constraint in a test ("styles land in `<main>`, never `<head>`"), would preserve the guard. [Fix this →](https://claude.ai/code?q=In%20src/packages/web-shell/src/inject-page-styles.ts%20restore%20the%20rationale%20comment%20removed%20in%20this%20PR%20that%20explains%20WHY%20page%20CSS%20is%20injected%20inside%20%3Cmain%3E%20rather%20than%20%3Chead%3E%3A%20htmx%20boosted%20navigation%20%28hx-target/hx-select%3Dmain%2C%20hx-swap%3DouterHTML%29%20swaps%20%3Cmain%3E%20atomically%20with%20its%20content%2C%20so%20a%20stylesheet%20in%20%3Chead%3E%20would%20be%20stranded%20and%20break%20cross-page%20boosted%20swaps%20%28reader%20to%20queue%29.%20Update%20the%20wording%20to%20cover%20both%20the%20%3Cstyle%3E%20inline%20and%20%3Clink%3E%20href%20variants.%20Branch%20claude/inline-css-bloat-6hnmqs&repo=Readplace/readplace.com)

### 2. First boosted swap to a not-yet-visited page type can briefly flash unstyled

Unchanged from the prior review and already accepted for the wave-1 prod verification — repeated only so the count reflects it. The `<head>` preload (`base.template.ts:35` / `chromeless-page.template.ts:10`) warms the CSS for the *current* page, but a boosted `<main>` swap to a page type the user hasn't loaded yet must fetch `/styles/<name>.<hash>.css` before it applies. `immutable` makes this at most once-per-user-per-page and same-destination swaps are already cached, so the blast radius is small. No action needed pre-merge.

### 3. `stylesheetsByName` is process-global; `page-stylesheets.test.ts` relies on name-uniqueness, not isolation

Also unchanged from the prior review. `page-stylesheets.ts:9` holds a module-level `Map` that is never reset, and the unit test registers into it with fixed names (`alpha`/`beta`/`gamma`/`delta`) that are disjoint from the real page names — so it's safe today, and the duplicate-registration `assert` is a genuinely useful import-time fail-fast. The registry must be process-global for the route to resolve stylesheets, so a `beforeEach` reset isn't a clean option (it would wipe the real registrations the integration tests depend on). Flagged only as a latent order-dependence to keep in mind if a future test reuses a name — no change recommended.

PR #1040, opened 2026-08-02
perf(hutch): swap the card in place for a queue status change

FagnerMartinsBrack:
## What & why

A queue **card** mark-as-read / mark-as-unread re-shipped the entire `<main>` (~131 KB, up to ~2.5 s in production per the 2026-07-20 measurement) even though the only visible effect is that one card leaves the list. An htmx submit now answers with a **~2 KB card-scoped swap**: the card removes itself (`hx-target="closest .queue-article"`, empty primary body) and the toast + counts refresh **out of band**, off the mutation's critical path.

The no-JS, Undo, reader, Siren, and feature-tab paths stay byte-identical to before.

## How it works

- **Marker, not trusted state.** `toStatusActions` appends `swap=card` to the status action URL. The handler treats it only as a *representation hint* (`req.get("HX-Request") === "true" && req.query.swap === "card"`), then re-probes the page itself (metadata-only, `excludeContent: true`) and decides the response shape. Undo / reader / Siren callers build their own hrefs without the marker and keep the 303; a request on a feature tab keeps the 303 too.
- **Server-determined fallback.** When the DOM would drift — the page emptied, the requested page is now past the last page, or the change didn't apply — the handler re-renders the *full listing* via `HX-Retarget: main` / `HX-Reswap` / `HX-Reselect`, exactly as the 303 → GET path would. A still-full page just removes the card (the list holds its place, matching the iOS app).
- **Out-of-band toast + counts.** The toast swaps into a stable `#status-toast` mount; the counts loader span is re-armed with `hx-swap-oob` so its `load` trigger re-fires and the badge / "Page X of Y" refresh via `GET /queue/counts` — keeping the full-partition COUNT (removed from the render path in `faf29536`) off the mutation TTFB.
- **Focus reuses main's toast mechanism.** The status button now carries a stable `id`; main's `toast.client.ts` records it on submit and, finding it gone after the swap, lands focus on the confirmation toast (its comment already anticipated "a future flow that removes it"). The toast dismiss also rescans on `htmx:oobAfterSwap` so the OOB toast auto-dismisses.

## Scope note — status only

[...]

PR #1041, opened 2026-08-02
feat: hx-boost the article-open links and the header nav

FagnerMartinsBrack:
## What

hx-boost the most common CTA in the app — opening an article — and the header nav, so both swap only `<main>` in place instead of tearing down and re-parsing the whole document. This is a **perceived-latency** change (server time / TTFB is unchanged): it removes the ~300–900ms full-document teardown/re-parse on article open and preserves scroll and focus.

## How (three commits, sequenced so a regression is attributable)

1. **Script-delivery infrastructure** (`@packages/web-shell`, `hutch`) — the load-bearing change. `hx-select="main"` keeps only the swapped `<main>` and htmx re-executes only the scripts *inside* it, so a page bundle emitted at the end of `<body>` is dropped on a boosted arrival (the reader iframe would collapse; scripted nav destinations — account, import, inbox, install — would go inert). New `injectPageScriptsIntoMain` moves each page's own scripts (and per-request markup) inside `<main>` so they ride the swap; global scripts (htmx, toast, local-time, webmcp, trial-countdown, extension-suggestion-banner) stay outside and load once. Because in-`<main>` scripts now re-execute on every same-page swap (mark-read, filter), each bootstrap footer is guarded to run once per document (`window.__<outfile>Init`) — the factory's own swap listener handles later swaps — and DCL-gated footers convert to a `readyState` check that also fires post-boost (a bare `DOMContentLoaded` listener never runs after a boost).

2. **Card article-open links** (`hutch`) — the title, excerpt, and thumbnail anchors get `hx-boost hx-target="main" hx-select="main" hx-swap="outerHTML show:top"` (`show:top` — a fresh article starts at the top, unlike the mark-read forms' `show:none`).

3. **Header nav** (`@packages/web-shell`, `hutch`, `inbox`) — new `GlobalBoostedNav`; hutch and inbox point at it. GET links boost; **logout stays a plain POST** (a boosted logout would swap `<main>` yet leave the authenticated nav — now pointing at forbidden routes — stranded). blog-site and web-embed keep the unboosted `GlobalNav`.

## Decisions (the plan's open questions)

- **Analytics:** accepted the pageview loss for boosted GETs. The internal click (`utm_medium=internal`) and server-side reader-presence (`markArticleViewed`) are both retained, so "article opened" stays measurable.
- **Injection scope:** unconditional — matches how page styles already ride the swap; every page bundle is guarded.
- **Nav boost shape:** emitted per nav item only when `method === "GET"`, so logout is excluded with no new field.
- **Nav blast radius:** scoped to the app surfaces (hutch + inbox) via `GlobalBoostedNav`; the content surfaces (blog, embed) keep full-page navigation.

## Testing

- New unit tests: `injectPageScriptsIntoMain`; page-scripts-inside-`<main>` (globals stay outside) in the base shell; the 3 boosted card links; `GlobalBoostedNav` (GET items boosted, logout not, guest items). Updated the queue listing test that asserted the old "no htmx on title links" invariant.
- Green at their coverage thresholds: `@packages/web-shell`, `hutch` (3484 tests), `inbox`, `web-embed`, `blog-site`. All 19 client bundles syntax-checked after the guard rewrite.

## Notes for reviewers

- Rebased on latest `main`, which removed the `homepage-split` bundle, added view-paywall `dispatchDocumentEvent` + toast focus listeners, and introduced `state.requestScripts`. `requestScripts` now rides `<main>` alongside `body.scripts` (a per-request page script, so it should survive a boost); the two ordering tests main added still pass.
- `pnpm check` could not run `editorconfig-checker` in the dev session (its release binary can't be fetched under the session's GitHub repo scoping); every other check passes and the changed files were verified against `.editorconfig` by hand. CI runs the real check.
- The idempotency guards and the boosted open are best confirmed at runtime by the `queue-flow` E2E (reader iframe sized, back restores the queue, repeated swaps don't leak listeners).

https://claude.ai/code/session_017gw79crQph8jFfS8PdwK1U

---
_Generated by [Claude Code](https://claude.ai/code/session_017gw79crQph8jFfS8PdwK1U)_


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 30748379800
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb
- **Attempt**: 1 of 5

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **Run ID**: 30776689079
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:
Rebased onto `main` (`e54abc87`) and force-pushed with `--force-with-lease`.

No conflicts — the branch's commits replayed cleanly onto current `main`. No content was changed as part of the rebase.

FagnerMartinsBrack:

## Context
- **Attempt**: 3 of 5
- **Run ID**: 30780053515
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:
Picked up the CI failure directly — the attempt-3 run reached the push step and died there (run `30780275919` ended `failure` at 04:11 without pushing), so its verified work never landed.

I re-derived the root cause and fixed all three surfaces. **The queue-flow E2E now passes locally (1 passed, 52.1s)** — the same spec CI was failing.

One cause, three surfaces: `hx-select="main"` keeps only the swapped `<main>`, so everything the shell renders outside it kept the origin page's values.

1. **`<body class="page-*">` never changed.** queue → reader left `page-queue`. That is live page-scoped-CSS breakage, and it is why CI timed out on `waiting for locator('body.page-login')` after a boosted nav click.
2. **Extension-suggestion banner.** It lives in `.banner-area` outside `<main>`, so it kept the origin page's `data-show-extension-suggestion`; its `attach()` also only ever *added* the visible class, never removed it.
3. **`read-first-article` raced the swap.** It waited on `waitForLoadState('domcontentloaded')`, which returns immediately for an in-place swap.

**Approach:** both values now ride inside `<main>` via a new `injectPageChromeIntoMain` — the same carrier idea this PR already uses for page scripts and styles, which is what makes them survive the swap. A global script (outside `<main>`, bound once) restores the body class; the banner client re-evaluates on `htmx:afterSwap`, toggling both ways and binding its close button once, via the `addSwapListener` dep shape `local-time`/`toast`/`trial-countdown` already use. Both re-apply only when the swapped-in markup actually carries the marker, so a fragment response smaller than `<main>` cannot blank them.

`@packages/web-shell` is green at 362 tests / 100% coverage; committed through the pre-commit hook with the full `pnpm check`, no `--no-verify`.

---

### Separate critical issue this PR still has — `runOnce` permanently disables five bundles

Not fixed here, because the correct shape is your design call and it needs per-bundle teardown.

`runOnce` in `projects/hutch/scripts/build-client-bundles.js:88` latches on a per-outfile `window.__<name>Init` flag. Its comment justifies this as *"the factory's own swap listener, registered on that first run, handles every later swap"* — but that is false for the factories that take no `addSwapListener`. I checked all twelve latched footers; **five have no swap listener**:

`share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`

Because page bundles now re-execute inside `<main>` on every boosted arrival, the flag is already `true`, so the factory never runs against the freshly swapped DOM and the feature stays inert for the rest of the document's life. On `main` every article open is a full document load, so these always initialised — this is a regression the diff introduces.

Concretely: open article A (share balloon attaches), nav back to Queue, open article B — B's `[data-share-balloon-wrap]` keeps its server-rendered `hidden` and its buttons have no listeners. Worse, A's still-registered document-level listener fires against the detached node.

Two viable shapes: gate `runOnce` on the `<main>` instance rather than the document, or give the five factories a swap listener and let `runOnce` gate only the listener registration. `initShareBalloon` returns `{ attach, detach }`, so whichever you pick, the previous controller needs detaching before re-attaching or article A's scroll/htmx listeners leak.

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: b4f932f33c288c459d6a5f74151e1b798175beb3

FagnerMartinsBrack:
### Code Review — PR #1041

Solid, well-sequenced change. The carrier idea (page scripts/styles/chrome ride *inside* `<main>` so a boosted swap re-runs them) is applied consistently, the `injectPageChromeIntoMain`/`injectPageScriptsIntoMain` string-insert reasoning (no DOM round-trip so the double-escaped `srcdoc` survives) is sound, and the nav/banner/inject tests are thorough. One functional regression remains in the reviewed commit.

### 1. `runOnce` permanently disables the five no-swap-listener page bundles on the 2nd+ boosted arrival (share-balloon is squarely in the queue→reader path)

`runOnce` (`build-client-bundles.js` L40) latches each page bundle on a per-outfile `window.__<name>Init` flag. Its comment justifies this as *"the factory's own swap listener, registered on that first run, handles every later swap"* — but that only holds for factories that **do** register a `document.body` swap listener. Five latched bundles take no swap listener:

`share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`

Because these bundles are delivered *inside* `<main>` (e.g. `view.component.ts:235-244` puts `SHARE_BALLOON_SCRIPT`/`EXPIRY_COUNTER_SCRIPT`/`VIEW_PAYWALL_SCRIPT` into `body.scripts`, which `injectPageScriptsIntoMain` rides inside the swap), htmx re-executes them on every boosted arrival — but the flag is already `true`, so the factory never runs against the freshly-swapped DOM and the feature stays inert for the rest of the document's life. On `main` every article open was a full document load, so these always initialised — this PR introduces the regression.

**Concrete repro (core flow):** open article A (share balloon attaches) → nav back to Queue → open article B. B's `[data-share-balloon-wrap]` keeps its server-rendered `hidden` and its share/copy/close buttons have no listeners. Worse, `initShareBalloon.attach()` registered `document`-level `htmx:afterSwap` and `PAYWALL_REVEALED` listeners (`share-balloon.client.ts:229-234`) bound to A's now-detached nodes, and those keep firing on every later swap. `home` is also reachable this way — the guest **Features** item (`/#what-works`) boosts to `/`, so a second visit to `/` gets no headline rotator.

`initShareBalloon` already returns `{ attach, detach }` — the machinery for a fix exists, it just isn't wired for the boosted case. Two viable shapes:
- gate `runOnce` on the swapped `<main>` instance rather than the document, **or**
- give the five factories a `document.body` swap listener and have `runOnce` gate only the *listener registration*, with the listener calling the previous controller's `detach()` before re-attaching (so A's scroll/htmx listeners don't leak).

No test currently catches this class of regression (none assert a page bundle re-binds across a boosted swap), so a fix should add one. This was already surfaced in the thread and explicitly deferred as a design call; flagging it here because it is user-facing and unfixed at `b4f932f3`. [Fix this →](https://claude.ai/code?q=Fix%20the%20runOnce%20regression%20in%20projects%2Fhutch%2Fscripts%2Fbuild-client-bundles.js%20on%20branch%20claude%2Fhx-boost-article-nav-dk0mnb%3A%20the%20five%20page%20bundles%20with%20no%20swap%20listener%20(share-balloon%2C%20home%2C%20expiry-counter%2C%20save-error%2C%20view-paywall)%20latch%20on%20window.__%3Cname%3EInit%20and%20never%20re-init%20on%20a%202nd%20boosted%20swap%20into%20%3Cmain%3E%2C%20so%20a%20second%20opened%20article%20leaves%20its%20share%20balloon%20server-hidden%20with%20dead%20buttons%20and%20the%20first%20article%20controller%20document-level%20htmx%3AafterSwap%2FPAYWALL_REVEALED%20listeners%20keep%20firing%20on%20detached%20nodes.%20Either%20gate%20runOnce%20on%20the%20swapped%20%3Cmain%3E%20instance%20instead%20of%20the%20document%2C%20or%20give%20these%20factories%20a%20document.body%20swap%20listener%20and%20call%20the%20previous%20controller%20detach()%20(initShareBalloon%20returns%20%7Battach%2Cdetach%7D)%20before%20re-attaching.%20Add%20a%20test%20that%20a%20page%20bundle%20re-binds%20across%20a%20boosted%20swap.&repo=Readplace/readplace.com)

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`NAV_FEATURES` is `method: "GET"`, so `GlobalBoostedNav` boosts it with `hx-swap="outerHTML show:top"` (`nav.component.ts:61`). A boosted swap scrolls the new `<main>` to the top rather than to the `#what-works` anchor, so "Features" no longer jumps to the features section. Worth a quick check of whether the deep-link intent still holds (it may already be moot pre-PR, since the GET form's hidden `utm_*` inputs rewrite the query and can strip the fragment on submit). If it matters, exclude in-page-anchor items from boosting.

### 2. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
`base.component.ts:80-83` assigns `document.body.className = main.getAttribute('data-page-class')` on each `htmx:afterSwap`. Fine today (every full page sets a single `bodyClass`, and the guard skips when the swapped-in `<main>` carries no marker). Two edges to keep in mind if pages evolve: boosting to a page whose `bodyClass` is `undefined` leaves the *previous* page's body class in place (no `data-page-class` to read), and any JS-added body class would be wiped on the next swap. No action needed now.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Review to Address

FagnerMartinsBrack:

## Context
- **Attempt**: 4 of 5
- **Run ID**: 30791342451
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: 83f075363d926cf19ee540a606e602c82cc6cdeb

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: eb856f264b2e379e604679be6be55af41928c04a

FagnerMartinsBrack:
### Code Review — PR #1041 (`eb856f26`)

The previously-flagged HIGH regression is **resolved at this commit**, and the fix is clean and well-tested. The new `fix(hutch): re-create page bundles when a boosted swap replaces <main>` commit adds `initBoostedPageBundle` (`boosted-page-bundle.client.ts:8-23`) — the maintainer's sanctioned *option 2*: `runOnce` now gates only the one-time `document.body` swap-listener registration, and the listener tears down the previous controller (`cleanup()`) before re-creating against each freshly-swapped `<main>`. The five factories were migrated with proper teardowns (`initShareBalloon.detach()`, the new `initViewPaywall.detach()`, `rotator.stop()`, the two `stop()`s), and the helper's `target === currentMain` guard makes it robust to htmx's swap/script-execution ordering. Test coverage lands where the prior reviews said it was missing — `boosted-page-bundle.client.test.ts` asserts the create → cleanup → re-create cycle across swaps.

One instance of the **same defect class** was missed by the migration.

None found — the `runOnce` regression on the five enumerated bundles (`share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`) is fixed and verified.

## Medium Priority Issues

### 1. `initUploadProgress` (import) is a sixth un-migrated instance of the boosted re-init gap

The prior reviews enumerated "five bundles with no swap listener." `import.client.js` was not among them because, at the *bundle* level, it does register a swap listener — but only for one of the **two** factories it bootstraps under the shared `runOnce("__importInit")` (`build-client-bundles.js:211-227`):

- `initIndeterminateCheckboxes` registers its own `document.body` `htmx:afterSwap` listener (`import.client.ts:33-39`) — **self-heals** on each boosted arrival.
- `initUploadProgress` binds listeners only to `form.import__upload-form` and has **no** swap listener (`import.client.ts:73-108`).

`NAV_IMPORT` is `method: "GET"` (`banner-state.ts:169`), so `GlobalBoostedNav` boosts "Import Links". On the **2nd+** boosted arrival to `/import`, htmx re-executes the in-`<main>` bundle but `window.__importInit` is already `true`, so the whole footer is skipped — the freshly-swapped upload form gets no `htmx:xhr:progress` bar, no `importState="uploading"` toggle, and no `htmx:sendError`/`responseError`/`timeout`/`swapError` → native-submit fallback.

**Why MEDIUM, not HIGH:** the form still posts via its server-rendered `hx-post`, and the old form's listeners GC with the swapped-out `<main>` (no document-level leak, unlike `share-balloon`). The impact is a degraded upload (no progress UI) and a lost error-recovery fallback on the re-visited page — a real regression (`main` full-loaded `/import` every time), but the form is not dead.

**Fix:** wrap `initUploadProgress` in `initBoostedPageBundle` (selector `form.import__upload-form`) exactly like the five just-fixed bundles, so it re-creates against each swapped-in `<main>` (no teardown needed — the form is discarded with the swap, so `create` can return a no-op). Do **not** reach for a naive `addSwapListener(run)` as the sibling checkbox factory does: `initUploadProgress` uses `form.addEventListener`, so re-running it on a non-`<main>` swap would stack duplicate listeners on the same form — the main-identity gate in `initBoostedPageBundle` is what avoids that. No test asserts import re-binds upload progress across a boosted swap; the fix should add one. [Fix this →](https://claude.ai/code?q=Fix%20the%20remaining%20runOnce%20regression%20on%20branch%20claude%2Fhx-boost-article-nav-dk0mnb%3A%20import.client.js%20gates%20initIndeterminateCheckboxes%20and%20initUploadProgress%20under%20runOnce%28%27__importInit%27%29%2C%20but%20initUploadProgress%20%28projects%2Fhutch%2Fsrc%2Fruntime%2Fweb%2Fpages%2Fimport%2Fimport.client.ts%3A73%29%20binds%20only%20to%20form.import__upload-form%20with%20no%20swap%20listener%2C%20so%20a%202nd%20boosted%20arrival%20to%20%2Fimport%20%28NAV_IMPORT%20is%20GET%2C%20boosted%20by%20GlobalBoostedNav%29%20leaves%20the%20fresh%20upload%20form%27s%20progress%20bar%20and%20error-fallback%20unwired.%20Wrap%20initUploadProgress%20in%20initBoostedPageBundle%20%28selector%20form.import__upload-form%29%20like%20the%20five%20fixed%20bundles%3B%20do%20not%20use%20a%20naive%20addSwapListener%28run%29%20since%20form.addEventListener%20would%20stack%20duplicate%20listeners%20on%20a%20non-main%20swap.%20Add%20a%20test%20that%20import%20re-binds%20upload%20progress%20across%20a%20boosted%20swap.&repo=Readplace/readplace.com)

## Low Priority Suggestions

### 1. Stale comment reference in `PAGE_CLASS_SCRIPT`
`base.component.ts:74` says "see injectPageClassIntoMain", but the function is named `injectPageChromeIntoMain` (it was generalised from page-class to chrome). One-word fix.

### 2. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`NAV_FEATURES` is `method: "GET"`, so `GlobalBoostedNav` boosts it with `hx-swap="outerHTML show:top"`, scrolling the fresh `<main>` to the top rather than to `#what-works`. As the prior review noted, this is likely already moot pre-PR (the GET form's hidden `utm_*` inputs rewrite the query on submit and drop the fragment), so no action unless the deep-link intent still matters — in which case exclude in-page-anchor items from boosting.

### 3. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
`base.component.ts:80-82` assigns the whole `className` from `data-page-class`. Fine today (every page sets one `bodyClass`; the guard skips when the swapped-in `<main>` carries no marker). Two edges if pages evolve: boosting to a page whose `bodyClass` is `undefined` leaves the previous body class in place, and any JS-added body class is wiped on the next swap. No action needed now.

FagnerMartinsBrack:

## Context
- **Attempt**: 2 of 5
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: 339b40c1d9dba61c27508d5f6bac67f0f8172b43

FagnerMartinsBrack:
The `339b40c1` commit resolves the one MEDIUM issue from the `eb856f26` review — `initUploadProgress` is now wrapped in `initBoostedPageBundle` (selector `form.import__upload-form`), so a 2nd+ boosted arrival to `/import` re-wires the freshly-swapped upload form's progress bar and native-submit fallback. The fix is the maintainer-sanctioned *option 2* the five earlier bundles use, and the `create → no-op teardown` shape is correct here: `initUploadProgress` binds only to the form (`form.addEventListener`), which is discarded with the swapped-out `<main>`, so there is nothing to tear down and no document-level leak — unlike `share-balloon`. The added test (`import.client.test.ts` → *"initUploadProgress across a boosted swap"*) drives the real `initBoostedPageBundle` + `initUploadProgress` composition and asserts the **swapped-in** form (a node the first init never touched) toggles `data-import-state` and renders the progress fill, which proves re-init rather than stale wiring.

I independently re-traced all twelve hutch page bundles plus the shared global bundles to confirm no un-migrated instance of this defect class remains:

- **`initBoostedPageBundle` (6):** `share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`, and now import's `initUploadProgress`.
- **Self-heal by re-querying on swap (rest):** `progress-bar` (`findBars()` re-scans, WeakMap-keyed), `reader-nav` (`onSwap` re-queries `main`; `.header` is stable chrome), `reader-iframe` (`scan()` re-queries all iframes, binds/unbinds), `account-cards` (`mountElements` re-queries on `htmx:afterSettle`, guarded per-container by `data-card-mounted`), `crawl-bookmark` (`syncFrom` re-queries, guarded per-element by `crawl-bookmark--js`), `summary-toggle` (`bind` re-queries), import's `initIndeterminateCheckboxes` (`applyIndeterminate` re-queries). The global shared bundles (`toast`, `local-time`, `trial-countdown`, `extension-suggestion-banner`) carry `addSwapListener`; `webmcp` is a one-time registration; `clipboard-copy` (inbox/install) rides inside `<main>` via `INBOX_SCRIPT`/page `scripts` with a guard-free footer, so it re-executes and re-attaches to the fresh buttons with no document-level listener to stack.

The nav-boost shape is sound (`boostAttrsFor` emits `NAV_BOOST_ATTRS` only for `method === "GET"`, so logout stays a plain POST; `{{{boostAttrs}}}` is a constant, not user input), and `injectPageChromeIntoMain` correctly validates `bodyClass` against `/^[A-Za-z0-9 _-]+$/` before interpolating it into the `<main>` attribute (no injection vector). Coverage/test patterns match the repo conventions (real DOM + event-log assertions, no mocks, no negative assertions).

None found — the `runOnce` regression on the five enumerated bundles is fixed at `eb856f26`, and the sixth (`initUploadProgress`) is fixed and tested at `339b40c1`.

## Medium Priority Issues

None found.

## Low Priority Suggestions

### 1. Stale function reference in the `PAGE_CLASS_SCRIPT` comment (newly actionable)
`base.component.ts:74` still says *"(see injectPageClassIntoMain)"*, but that function does not exist — it was generalised to `injectPageChromeIntoMain` (it now carries both `data-page-class` and `data-extension-suggestion`). One-word fix; flagged in the `eb856f26` review and untouched by `339b40c1` (which only changed the import footer). [Fix this →](https://claude.ai/code?q=Fix%20the%20stale%20comment%20in%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.component.ts%20line%2074%20on%20branch%20claude%2Fhx-boost-article-nav-dk0mnb%3A%20the%20PAGE_CLASS_SCRIPT%20doc%20comment%20says%20%22(see%20injectPageClassIntoMain)%22%20but%20the%20function%20is%20named%20injectPageChromeIntoMain.%20Correct%20the%20reference.&repo=Readplace/readplace.com)

### 2. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`NAV_FEATURES` is `method: "GET"`, so `GlobalBoostedNav` boosts it with `hx-swap="outerHTML show:top"`, scrolling the fresh `<main>` to the top rather than to `#what-works`. As the prior two reviews noted, this is likely already moot pre-PR (the GET form's hidden `utm_*` inputs rewrite the query on submit and drop the fragment). No action unless the guest deep-link intent still matters — in which case exclude in-page-anchor items from boosting.

### 3. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
`base.component.ts:80-82` assigns the whole `className` from `data-page-class`. Fine today (every page sets one `bodyClass`; the guard skips when the swapped-in `<main>` carries no marker). Two edges if pages evolve: boosting to a page whose `bodyClass` is `undefined` leaves the previous body class in place, and any JS-added body class is wiped on the next swap. No action needed now.

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb
- **Attempt**: 1 of 5

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: 42291d28c215950ac038ed6e1719bf64a9434fee

FagnerMartinsBrack:
### Code Review — PR #1041 (`42291d28`)

The change is sound and thoroughly tested. `42291d28` is the rebased continuation of the line last reviewed at `339b40c1`, and every previously-flagged HIGH/MEDIUM is resolved:

- **Boosted re-init** — all six page bundles (`share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`, import's `initUploadProgress`) go through `initBoostedPageBundle`, whose `target === currentMain` gate re-creates against each swapped-in `<main>` and calls the previous controller's teardown first (`detach()` / `stop()`), so no document-level listener leaks onto a detached node. `runOnce` now gates only the one-time listener registration. The teardowns target the leak-prone listeners: `initHeadlineRotator.stop()` removes the `document` `visibilitychange` listener (the scroll-hint's click listener GCs with the swapped-out `<main>`), and `initViewPaywall.detach()` removes the `window` scroll listener + deadline timer.
- **Nav boost shape** — `boostAttrsFor` emits `NAV_BOOST_ATTRS` only for `method === "GET"`, so logout stays a plain POST (asserted at `nav.component.test.ts:336`). `{{{boostAttrs}}}` is a constant, not user input.
- **Chrome carry** — `injectPageChromeIntoMain` writes `data-page-class` (validated `/^[A-Za-z0-9 _-]+$/`) and `data-extension-suggestion` onto `<main>`; the once-bound `PAGE_CLASS_SCRIPT` re-applies the class and the extension-banner's `adoptSwappedState()` toggles both ways on `htmx:afterSwap`, each guarded so a fragment smaller than `<main>` can't blank them.

I re-traced the rebase itself, since that is what is new here: `main` (`36c11e99`) **removed the reader iframe** (article body now renders in the page's own document). The conflict resolution correctly dropped the deleted `reader-iframe` bundle entry from `build-client-bundles.js` while preserving every other footer's guard — the bundle set matches `main`, the reader/view pages still render through `Base` (`GlobalBoostedNav` shell) so the boosted card-open carries its chrome, no dangling import remains, and the reported `pnpm check` is green (hutch **E2E 26 passed**, functions **100%**). The `read-first-article` E2E step correctly moved to `clickAndWaitForPageReload`.

None found.

## Medium Priority Issues

None found.

## Low Priority Suggestions

### 1. The rebase left stale **reader-iframe** references in the new inject helpers and their tests
`main`'s removal of the reader iframe happened *after* these files were written, so the double-escaped-`srcdoc` example they cite no longer exists in the tree: comments at `inject-page-chrome.ts:20`, `inject-page-scripts.ts:9` and `:13`; test names + `data-reader-iframe` / `reader-iframe.client.js` fixtures in `inject-page-chrome.test.ts:85-86`, `inject-page-scripts.test.ts:8,11,35-39`, `base.component.test.ts:135-175`. The code is correct — string-insert (no parse/serialize round-trip) is still right and the tests still validate byte-for-byte preservation. Only the motivating example is stale; the surviving escaped-`srcdoc` case is now inbox's email iframe (`inbox-email-iframe-srcdoc.ts`, which flows through these helpers via `initBase`). Re-ground the wording, don't change the code.

### 2. Stale `injectPageClassIntoMain` reference in the `PAGE_CLASS_SCRIPT` comment
`base.component.ts:73` says *"(see injectPageClassIntoMain)"* but the function is `injectPageChromeIntoMain`. Flagged in prior reviews, untouched by the rebase — bundle with #1.

### 3. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`GlobalBoostedNav` boosts the GET item with `hx-swap="outerHTML show:top"`, scrolling to the top rather than `#what-works`. Likely moot pre-PR (the GET form's `utm_*` inputs rewrite the query and drop the fragment). No action unless the guest deep-link intent still matters.

### 4. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
`base.component.ts:80-82` assigns the whole `className` from `data-page-class`. Fine today; two edges if pages evolve (undefined `bodyClass` leaves the previous class; a JS-added body class is wiped on the next swap). No action needed now.

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb
- **Attempt**: 1 of 5

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: 07de07ba9b54d4f606cdb2c1482ecf29b47a29ec

FagnerMartinsBrack:
`07de07ba` is the rebase of the previously-approved line (last reviewed clean at `42291d28`) onto main's CSP-nonce commit (`66482188`), plus the earlier HIGH/MEDIUM fixes. The rebase is the only genuinely new surface, and it is integrated correctly:

- **CSP nonce** — the PR's new inline `PAGE_CLASS_SCRIPT` was converted to `pageClassScript(cspNonce)` and emits `<script nonce="…">`, satisfying main's "every inline script/style carries the request nonce" invariant; the invariant test's expected inline-script count moved 7 → 8 and passes, which is what proves `pageClassScript` is nonced. `injectPageStylesIntoMain({…, cspNonce})` (object form) and `htmxScripts(state.cspNonce)` are preserved, and `injectPageChromeIntoMain` writes only data-attributes (no nonce needed) after validating `bodyClass` against `/^[A-Za-z0-9 _-]+$/` (no injection vector).
- **Page scripts now ride inside `<main>`** (`body.scripts` + `state.requestScripts` moved out of the end-of-body slot into `injectPageScriptsIntoMain`). I traced every page's `scripts:` — the external bundle refs (`<script src>`) are nonce-exempt and `runOnce`-guarded where they register document-level listeners; the only inline nonced page scripts (`autoSubmitScript`, `uploadAutoSubmitScript`, `fromUrlAutoSubmitScript`) sit on `?save=`/`?url=` param variants that no boosted link targets, are `readyState`-guarded, and re-bind to the *fresh* form a `<main>` swap brings in (no listener stacking).
- **Boosted re-init** — all six bundles go through `initBoostedPageBundle`, whose `target === currentMain` gate re-creates against each swapped-in `<main>` and calls the previous controller's teardown first (`detach()`/`stop()`); the teardowns target the leak-prone listeners (view-paywall `window` scroll + deadline timer, home `document` `visibilitychange`). `boosted-page-bundle.client.test.ts` covers the full create → cleanup → re-create cycle.
- **Load-bearing htmx assumption verified** — against the vendored htmx 2.0.8, an `outerHTML` swap pushes the swapped-in `<main>` into `settleInfo.elts` and dispatches `htmx:afterSwap` on it, so the body-level listener sees `e.target` = the new `<main>` (`tagName === "MAIN"`) — exactly what `initBoostedPageBundle` keys on.
- **Nav boost** — `boostAttrsFor` emits the attrs only for `method === "GET"`, so logout stays a plain POST (asserted); **card links** carry `hx-boost/target/select/swap="outerHTML show:top"` (tested); **chrome carry** toggles both ways with a marker guard.

None found.

## Medium Priority Issues

None found.

## Low Priority Suggestions

### 1. Stale comment references left by the rebase (cosmetic; code is correct)
`base.component.ts:74` still says *"(see injectPageClassIntoMain)"* but the function is `injectPageChromeIntoMain`; and the *"collapse the reader iframe's double-escaped srcdoc"* motivation in `inject-page-chrome.ts:20`, `inject-page-scripts.ts:13`, `inject-page-styles.ts:11` (plus the `data-reader-iframe` / `reader-iframe.client.js` test fixtures) cites the reader iframe that main removed. The string-insert approach and byte-for-byte tests are still right — only the cited example is stale; the surviving escaped-`srcdoc` case is now inbox's `inbox-email-iframe-srcdoc.ts`. Flagged in the prior two reviews and untouched by the rebase. [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fhx-boost-article-nav-dk0mnb%20correct%20stale%20comments%20without%20changing%20behaviour%3A%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.component.ts%20line%2074%20says%20%22(see%20injectPageClassIntoMain)%22%20but%20the%20function%20is%20injectPageChromeIntoMain%3B%20and%20the%20reader-iframe%20references%20in%20inject-page-chrome.ts%3A20%2C%20inject-page-scripts.ts%3A13%2C%20inject-page-styles.ts%3A11%20and%20the%20data-reader-iframe%2Freader-iframe.client.js%20test%20fixtures%20cite%20the%20reader%20iframe%20removed%20on%20main%20-%20re-ground%20them%20to%20inbox%27s%20inbox-email-iframe-srcdoc.ts%20or%20generalize%20the%20wording.&repo=Readplace/readplace.com)

### 2. Forward-looking: inline page scripts now travel through the boosted-swap path
No live issue — CSP is not yet enforced (`helmet({ contentSecurityPolicy: false })` in `lambda.main.ts:43`), so the per-request nonce is stamped but nothing blocks a mismatched one. Worth noting for the in-progress CSP work: once a nonce-based policy is enforced, an htmx-reinserted inline script inside a boosted `<main>` carries the *destination request's* nonce, which won't match the *original document's* policy — so the eventual policy should use `'strict-dynamic'` (the standard htmx pairing), which trusts scripts inserted by already-trusted htmx regardless of nonce. Only the inline scripts are affected; the `<script src>` bundles are authorised by the source expression.

### 3. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`NAV_FEATURES` is `method: "GET"`, so it boosts with `hx-swap="outerHTML show:top"`, scrolling to the top rather than `#what-works`. As the prior reviews noted, likely already moot (the GET form's hidden `utm_*` inputs rewrite the query on submit and drop the fragment). No action unless the guest deep-link intent still matters.

### 4. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
Fine today (every page sets one `bodyClass`; the guard skips a marker-less `<main>`). Two edges if pages evolve: boosting to a page whose `bodyClass` is `undefined` leaves the previous class in place, and any JS-added body class is wiped on the next swap. No action needed now.

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Branch**: claude/hx-boost-article-nav-dk0mnb
- **Attempt**: 1 of 5

FagnerMartinsBrack:

## Context
- **PR**: #1041
- **Commit**: e3a3edf85929a9390edc33805b817714a36c993c

FagnerMartinsBrack:

None found — the `runOnce` regression on the six enumerated bundles (`share-balloon`, `home`, `expiry-counter`, `save-error`, `view-paywall`, import's `initUploadProgress`) is fixed and verified, and the rebase correctly extended the same boosted-swap machinery to main's newly-added `initSloganRotator`.

## Medium Priority Issues

None found.

## Low Priority Suggestions

### 1. Stale comment references left by main's reader-iframe removal (cosmetic; code is correct)
`base.component.ts:74` still says *"(see injectPageClassIntoMain)"* but the function is `injectPageChromeIntoMain`; and the *"collapse the reader iframe's intentionally double-escaped srcdoc"* motivation in `inject-page-chrome.ts:20` and `inject-page-scripts.ts:13` (plus the `data-reader-iframe` / `reader-iframe.client.js` test fixtures in `inject-page-chrome.test.ts:85-86` and `inject-page-scripts.test.ts:8,11,35-39`) cites the reader iframe that `main` (`36c11e99`) removed. The string-insert approach and byte-for-byte preservation tests are still correct — only the motivating example is stale; the surviving escaped-`srcdoc` case is now inbox's `inbox-email-iframe-srcdoc.ts`, which still flows through these helpers via `initBase`. Re-ground the wording (or generalize it) rather than change behaviour. Flagged in the prior two reviews and untouched by this rebase. [Fix this →](https://claude.ai/code?q=On%20branch%20claude%2Fhx-boost-article-nav-dk0mnb%20correct%20stale%20comments%20without%20changing%20behaviour%3A%20src%2Fpackages%2Fweb-shell%2Fsrc%2Fbase.component.ts%20line%2074%20says%20%22(see%20injectPageClassIntoMain)%22%20but%20the%20function%20is%20injectPageChromeIntoMain%3B%20and%20the%20reader-iframe%20references%20in%20inject-page-chrome.ts%3A20%2C%20inject-page-scripts.ts%3A13%20and%20the%20data-reader-iframe%2Freader-iframe.client.js%20test%20fixtures%20in%20inject-page-chrome.test.ts%3A85-86%20and%20inject-page-scripts.test.ts%3A8%2C11%2C35-39%20cite%20the%20reader%20iframe%20removed%20on%20main%20-%20re-ground%20them%20to%20inbox%27s%20inbox-email-iframe-srcdoc.ts%20or%20generalize%20the%20wording.&repo=Readplace/readplace.com)

### 2. Boosted **Features** nav item (`/#what-works`) loses its scroll-to-section
`NAV_FEATURES` is `method: "GET"`, so `GlobalBoostedNav` boosts it with `hx-swap="outerHTML show:top"`, scrolling the fresh `<main>` to the top rather than to `#what-works`. As the prior reviews noted, this is likely already moot pre-PR (the GET form's hidden `utm_*` inputs rewrite the query on submit and drop the fragment). No action unless the guest deep-link intent still matters — in which case exclude in-page-anchor items from boosting.

### 3. `PAGE_CLASS_SCRIPT` replaces `document.body.className` wholesale on every swap
`base.component.ts:80-82` assigns the whole `className` from `data-page-class`. Fine today (every page sets one `bodyClass`; the guard skips a marker-less `<main>`). Two edges if pages evolve: boosting to a page whose `bodyClass` is `undefined` leaves the previous body class in place, and any JS-added body class is wiped on the next swap. No action needed now.

FagnerMartinsBrack:
Closing this without a replacement PR. The change was built, verified in a browser, and measured — and the premise it rests on does not survive measurement.

## The premise

This PR's case was that opening an article pays a "~300–900 ms full-document teardown/re-parse" that an `hx-boost` `<main>` swap removes. That number appears in the description and in commit `8ec23781`, and nowhere else — not in the repo, not in a profile, not in a linked run.

## What was measured

A harness now lives on `main` (`pnpm nx run hutch:perf-article-open`, added in `7e71c32d`, corrected in `e6375a17`). It opens an article from `/queue` and times a capture-phase click on the card title to the frame the article body is fully laid out, across small/large articles at 1×/4×/6× CPU on loopback and on an emulated 150 ms / 4 Mbps link. It is arm-agnostic — it measures whatever build it points at — which is what let both arms be measured without either being merged.

The boosted arm is a local reconstruction of this PR: the card-title/excerpt `hx-boost`/`hx-target=main`/`hx-select=main`, plus `injectPageScriptsIntoMain` so the destination's bundles actually re-execute inside the swapped `<main>` (without that the boosted arm would skip real work and look unfairly fast), plus `data-page-class`.

Two runs per arm, interleaved (`baseline-a`, `boosted-a`, `baseline-b`, `boosted-b`), n=20 per cell after 3 discarded warm-ups, one macOS host, headless Chromium via Playwright 1.60. Inclusion rule fixed before results were read: both completed runs per arm, all samples, applied symmetrically.

## 1. The teardown this PR set out to remove costs 0.0 ms

Read off the destination's `PerformanceNavigationTiming` on every unboosted sample:

| | unload (ms) | pre-request (ms) |
|---|---:|---:|
| every baseline condition, both runs | **0.0** | 0.2–0.4 |

`unloadEventEnd - unloadEventStart` is **0.0 ms** in all eight conditions across both runs. (The field is calibrated: against a page carrying a 25 ms busy-loop unload handler it reads 25.0 ms, and 0.0 ms for the same page without one — so the zero is a real zero, not a broken probe.)

There is no ~300–900 ms teardown. Whatever a boost buys here, it is not what this PR said it was buying.

## 2. The CPU-side benefit is not resolvable by this instrument

| Condition | baseline → boosted (run A) | (run B) | verdict |
|---|---|---|---|
| loopback-cpu1x-small | 11.7 → 14.9 | 13.0 → 15.0 | below instrument resolution |
| loopback-cpu4x-small | 64.8 → 34.1 | 62.0 → 34.6 | below instrument resolution |
| loopback-cpu6x-small | 82.7 → 49.2 | 79.3 → 49.1 | below instrument resolution |
| loopback-cpu1x-large | 19.0 → 16.9 | 19.8 → 16.3 | below instrument resolution |
| loopback-cpu4x-large | 94.0 → 49.5 | 94.7 → 47.6 | below instrument resolution |
| loopback-cpu6x-large | 148.7 → 72.7 | 124.2 → 73.0 | A: right faster / **B: below resolution** |

The raw deltas lean toward the boost under CPU throttling (−27 to −76 ms), and that may well be a real effect — a full navigation re-parses and re-executes the whole shell where a swap does not. But the milestone is sampled with `requestAnimationFrame`, and under 4×/6× throttling a frame is 30–68 ms, which is larger than the deltas. Only one of twelve CPU cells cleared its own resolution gate, and **it did not replicate** in the second run. This harness cannot resolve the CPU question on this hardware, and I am not going to report a number it cannot support.

## 3. The boost is reproducibly slower on a slow link

| Condition | baseline → boosted | delta | resolution | 95% CI on the difference | verdict |
|---|---|---:|---:|---:|---|
| slow-mobile-small (run A) | 327.6 → 363.7 | +36.0 | 32.8 | 5.1 | boost slower |
| slow-mobile-small (run B) | 326.3 → 363.6 | +37.3 | 32.1 | 5.7 | boost slower |
| slow-mobile-large (run A) | 444.6 → 495.6 | +51.0 | 46.9 | 1.7 | boost slower |
| slow-mobile-large (run B) | 443.8 → 500.1 | +56.3 | 48.4 | 6.5 | boost slower |

Both cells, both runs, clear of both gates. The mechanism is structural: htmx cannot swap until the whole response has arrived, whereas the browser parses and lays out as bytes stream in — so the penalty scales with article size (+36 ms small, +51–56 ms large).

For a reading app whose users are largely on phones on cellular, that is the condition that matters most, and it is the one where the boost loses.

## A note on how this was arrived at

Two earlier measurement attempts got this wrong in opposite directions, which is worth recording:

- The first reported "never faster, ~61 ms slower on mobile". It measured no CPU-throttled conditions, so it could not have seen the CPU effect at all.
- The second reported "boost wins 25–43 ms when CPU-bound, loses 50–183 ms when network-bound". Its settle predicate was content-blind — a streaming navigation satisfied it off the first chunk while an htmx swap could only satisfy it with the whole body, so the arms were compared at different delivered states. The tell was that the unboosted arm cost the *same* for a 6-paragraph and a 120-paragraph article over the same slow link (312.3 vs 311.4 ms), i.e. it was measuring nothing about article size. With both arms settling on a sentinel at the end of the body, that arm costs 327.6 vs 444.6 ms, and the reported "+183 ms" shrinks to +51 ms.

Both are fixed in `e6375a17`. The harness now reports its own sampling resolution per condition and refuses to call a gap a difference when it is inside that resolution or inside the 95% interval on the difference of means.

## What is not being claimed

That boosting is wrong everywhere. The same-page boosts already on `main` — the save form, sort, filters, pagination, delete-confirm, the card status form, and #1040's card-scoped swap — are a different case and are untouched. This is specifically about a *cross-page* navigation, where the browser's streaming parse is competing against a JS parse that cannot start until the response is complete.

The structural observation in this PR is also still true and independently interesting: the shell delivers page **styles** inside `<main>` but page **scripts** outside it, so no cross-page boost can work without `injectPageScriptsIntoMain`. That asymmetry is real; it just isn't worth resolving for a change that measures slower where it counts.

Closing. The harness stays on `main` so the next person who proposes this can settle it in about two minutes instead of arguing about it.

PR #1057, opened 2026-08-09
perf(@packages/web-shell,hutch,inbox): serve htmx same-origin and defer it

FagnerMartinsBrack:
## What

htmx moves from a parser-blocking `cdn.jsdelivr.net` classic script to a same-origin, deferred `/client-dist/htmx.client.js`, and its one config value moves to htmx's native `<meta name="htmx-config">`.

## Why this shape

The old pair was a classic `<script src="https://cdn.jsdelivr.net/…">` immediately followed by an inline `<script>htmx.config.scrollBehavior='smooth';</script>`, first in the trailing script blob at the end of `<body>`.

Two costs followed. jsDelivr was the page's only third-party script origin and the repo carries no `preconnect`/`dns-prefetch` for it — the four preconnect lines that exist are all for fonts — so a first visit paid a cold DNS + TCP + TLS handshake to a third origin before the script could start downloading. And because the tag was classic rather than deferred, the parser blocked on that round trip, so `DOMContentLoaded` and every deferred bundle behind it waited on a third party. **No field measurement is claimed here** — this is the mechanism, not a number.

The inline config is what coupled delivery to ordering: it dereferenced the htmx global at parse time, so adding `defer` on its own throws `htmx is not defined`. Moving the setting to `<meta name="htmx-config">`, which htmx reads itself during init, removes that coupling and lets the tag be deferred.

## Delivery

`htmx.org` is pinned as an exact devDependency of the two deployables that serve `/client-dist`, and each `build-client-bundles.js` copies the package's prebuilt `dist/htmx.min.js` to `client-dist/htmx.client.js`. It is deliberately **not** added to the esbuild `BUNDLES` array: the file is already a minified IIFE, and re-bundling would only add sourcemap, coverage and knip surface. The `.client.js` name keeps the asset matching `TRIMMED_BUNDLE_PATH` in `static-asset-paths.ts`, so click attribution still classifies it as a static asset rather than a reader path.

The npm tarball's bytes are identical to what production served — 51,250 bytes, hashing to the same sha384 the deleted SRI attribute pinned:

```
$ wc -c < node_modules/.pnpm/htmx.org@2.0.8/node_modules/htmx.org/dist/htmx.min.js
51250
$ openssl dgst -sha384 -binary … | openssl base64 -A
/TgkGk7p307TH7EXJDuUlgG3Ce1UVolAOFopFekQkkXihi5u/6OCvVKyz1W+idaz   # == the deleted SRI
```

SRI and `crossorigin` go away because they carry no meaning same-origin.

That byte equality is also why the vendored e2e fixture goes: it was a copy of the same build, kept only so the hermetic CDN shim could answer the jsDelivr route. With nothing left to intercept, the route, the fixture and its knip entry are removed; the font routes are untouched. Worth knowing about the shim it replaces — it matched on hostname alone, so it was version-blind (its own test drove it with a 2.0.4 URL and still got the 2.0.8 fixture), and the staging suite never used it at all (`run.e2e-staging.ts` imports `test` from `@playwright/test`), so every staging deploy really did fetch from the CDN.

## Deferring htmx makes a dormant branch live

The reader capture bridge is an inline script, so the parser runs it — and registers its click listener — before any deferred script executes. Its handler read `window.htmx` synchronously and bailed out when the global was missing. While htmx was parser-blocking that bail-out could not fire; deferring htmx makes the branch live. Driving a WKWebView-shaped page with the htmx response held open shows the gap: at `readyState` `"interactive"` the bridge is armed while `window.htmx` is still `undefined`. A tap in that window disables the button and posts `captureBlocked`, but the reader-slot poll never starts and the native handler reloads only the reading list — the user is stranded on the blocked notice behind a disabled button. Registering a one-shot `htmx:load` listener when the global is absent keeps the immediate path unchanged and lets an early tap run its poll as soon as htmx initialises.

## htmx delivery is now a per-site decision

`/client-dist` is **hutch's** `express.static` mount. Four deployables render the shell — hutch, inbox, blog-site, web-embed — across six `BaseConfig` composition roots, and the first version of this change emitted the same-origin tag from all of them. blog-site and web-embed have neither the mount nor the copy step.

It did **not** 404 in production, and the reason is worth stating rather than assuming. Both are deployed behind hutch's API Gateway on the same origin (`GET /blog`, `ANY /blog/{proxy+}`, `GET /embed`, `ANY /embed/{proxy+}`); every path they do not claim falls through to hutch's `$default`. Verified against the live site: a `/blog` page already carries `<script src="/client-dist/toast.client.js" defer>`, and `https://readplace.com/client-dist/toast.client.js` returns `200 text/javascript`. What the tag *did* do is 404 on each project's own dev and e2e server, and turn a co-hosting arrangement into a runtime dependency — blog-site's infra program states the coupling to hutch is "deploy-time and one-way … There is no runtime code edge."

Neither site has any use for htmx. `grep -rn 'hx-[a-z]'` over `projects/blog-site/src` and `projects/web-embed/src` returns nothing; the only `hx-*` markup in the shared shell is `toast.template.ts`, and `renderToast` is called only by hutch and inbox. Both sites were paying a parser-blocking 51 KB CDN fetch on `main` for a library nothing on the page touched.

So `BaseConfig` gains a **mandatory** `htmx: HtmxDelivery`, and a site passes `HtmxLoaded` or `HtmxOmitted` — mandatory for the same reason `renderNav` is, so the next deployable to adopt the shell answers the question rather than inheriting an answer. Each value carries *both* fragments (the `<meta>` and the `<script>`), so neither the shell nor the template branches; the shape mirrors `GlobalNav`/`GlobalEmptyNav`. The chromeless shell keeps htmx unconditionally: hutch is its only consumer and its reader depends on it.

## The privacy post said htmx came from jsDelivr

`no-third-party-trackers.md` listed "the Inter web font from Google Fonts and the htmx library from jsDelivr" as the full set of what loads, and the section immediately after it is **"Check it yourself"**. After this change no jsDelivr request is made, so a reader taking up that invitation would find the accounting wrong — in the one post that can least afford it. The list now names the font alone, the intro no longer says "a few open-source assets from public CDNs" when one remains, and a short paragraph records that the library moved onto our own servers rather than quietly deleting the mention. Google Fonts is unchanged and still named as third-party.

## Verification

Runtime, against the compiled build:

| Check | Result |
|---|---|
| `GET /client-dist/htmx.client.js` (hutch) | `200`, `Content-Type: text/javascript; charset=utf-8`, `Cache-Control: public, max-age=300`, `Content-Length: 51250` |
| `GET /client-dist/does-not-exist.client.js` | `404` (mount is `fallthrough: false`) |
| `htmx.config.scrollBehavior` on a live reader page | `"smooth"`; `htmx.version` `"2.0.8"`; tag is `<script src="/client-dist/htmx.client.js" defer>`; config read from the `<head>` meta |
| blog-site `/blog` browser network log | **no htmx request at all** — neither same-origin nor jsDelivr; zero `htmx-config` metas |
| web-embed `/embed` browser network log | same |
| Boosted queue-tab swap (hutch) | `htmx:beforeRequest` → `afterSwap` → `afterSettle`, `<main>` replaced, URL via `pushState`, **0** full navigations |

`pnpm check` green across all 47 projects.

## Corrections to earlier claims in this description

Four statements in the first draft of this description were wrong. Recorded here so review does not inherit them.

1. **"Nothing fails at compile time if a _third_ deployable adopts the shell."** There were already **four** shell-consuming deployables (hutch, inbox, blog-site, web-embed) across six composition roots — and the failure was not hypothetical, it was realised: blog-site and web-embed really did emit the tag. `BaseConfig.htmx` is now mandatory, so this is a compile error rather than a silent inheritance.
2. **"`git grep 'window.htmx|htmx\.'` confirms `queue.page.ts` was the _only_ synchronous dereference."** That grep also returns `src/packages/web-shell/src/htmx-script.ts:4` on `main` — the inline `htmx.config.scrollBehavior='smooth'`, which is itself a synchronous dereference and which this PR deletes. Accurate version: **once the inline config tag is removed, `queue.page.ts:449-450` holds the only remaining synchronous dereference of the htmx global** — and that is exactly the one the second commit makes defer-safe.
3. **"The reader page alone ships 12 client bundles."** Measured: on `main` the reader page pulls **11** same-origin `/client-dist/` bundles plus htmx from jsDelivr. The 12th (`trial-countdown.client.js`) renders only for a user whose trial chip carries a live instant, so 12 was a conditional maximum, not the common case. After this change htmx becomes the 12th same-origin bundle (13 with a live trial chip) and the jsDelivr request goes away.
4. **"Two build scripts now share _four_ identical lines."** It is **five** — `fs.mkdirSync(…)` plus the four-line `fs.copyFileSync(…)`.

## Known, not fixed here

blog-site and web-embed still emit `/client-dist/toast.client.js` and `/client-dist/extension-suggestion-banner.client.js`, which the shell renders unconditionally for every site. Those 404 on each project's standalone dev/e2e server and resolve `200` in production through hutch's `$default` route. That is unchanged from `main` — this PR neither adds nor removes them — and deciding whether those two sites should render a toast or an extension banner at all is a separate change with its own product question.


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 31305183299
- **PR**: #1057
- **Branch**: fix/serve-htmx-same-origin

## Failed Jobs
- web-tests: Run pnpm check

FagnerMartinsBrack:

## Context
- **PR**: #1057
- **Commit**: 5950318895e8e52e80aae02694b48d1c7880739f

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The htmx copy block is duplicated verbatim across the two build scripts

`projects/hutch/scripts/build-client-bundles.js:362-366` and `projects/inbox/scripts/build-client-bundles.js:41-45` now carry the identical `fs.mkdirSync(OUT_DIR, …)` + `fs.copyFileSync(require.resolve("htmx.org/dist/htmx.min.js"), path.join(OUT_DIR, "htmx.client.js"))` block. The PR description already acknowledges this ("Two build scripts now share five identical lines"). It's a judgment call — the two scripts already duplicate their scaffolding (`OUT_DIR`, `rmSync`, esbuild wiring) and diverge elsewhere (watch mode, `ALL_BUNDLES` vs `BUNDLES`), so extracting one helper for five lines may not pay for the indirection. Flagging only so the choice is deliberate; leaving it as-is is defensible.

[Fix this →](https://claude.ai/code?q=In%20projects%2Fhutch%2Fscripts%2Fbuild-client-bundles.js%20and%20projects%2Finbox%2Fscripts%2Fbuild-client-bundles.js%20on%20branch%20fix%2Fserve-htmx-same-origin%2C%20the%20five-line%20mkdirSync%2BcopyFileSync%20block%20that%20copies%20htmx.org%2Fdist%2Fhtmx.min.js%20into%20client-dist%2Fhtmx.client.js%20is%20duplicated.%20Consider%20extracting%20it%20into%20a%20small%20shared%20workspace%20build%20helper%20both%20scripts%20import%2C%20or%20leave%20it%20if%20the%20indirection%20isn%27t%20worth%20it.&repo=Readplace/readplace.com)

PR #1063, opened 2026-08-12
fix(save-link): derive the tier snapshot's crawl status from the canonical enum

FagnerMartinsBrack:
Follow-up to the stale-check crawl-status incident (`aa377484`), shipped separately because it is unrelated to that fix.

## Why

`read-article-crawl-state-dynamodb.ts` restated the crawl-status enum inline — the same shape whose drift dead-lettered `stale-check-requested-handler` when `unsupported` was added canonically and the copy was never widened.

This copy happens to be in sync today. It is still worth closing:

- it parses the same column of the same table through the same throwing `table.get`;
- it is a thin AWS wrapper wrapped in `/* c8 ignore */`, so no unit test would ever catch it drifting;
- `crawlStatus` was typed against a *local* `CrawlStatus` union, so a new canonical member produced no compile error here.

## What

- `read-article-crawl-state-dynamodb.ts` — import `CrawlStatusSchema` from `@packages/article-state-types` instead of redeclaring it.
- `read-tier-snapshot.ts` — derive `CrawlStatus` from the canonical type (`PersistedCrawlStatus | "absent"`), mirroring how `ReaderStatusSchema` composes `CrawlStatusSchema` with `"unavailable"`.

Importing the schema alone fixes the parse but not the hazard. `tier1StatusFromCrawlStatus` ended in a catch-all `return "not_attempted"`, so a new member would have been absorbed silently and the tier dashboard would have under-reported rather than failed. It is now a `TIER_1_STATUSES` map with `satisfies Record<CrawlStatus, TierStatus>`, matching `TERMINAL_ACTIONS` in `decide-terminal-action.ts`.

Behaviour is unchanged: `ready → success`, `failed → failed`, everything else → `not_attempted`.

## Verification

`pnpm check` passes across all 47 projects.

The guard was proved rather than assumed — adding a fifth member (`"blocked"`) to the canonical enum now fails the build here:

```
read-tier-snapshot.ts:37:3 - error TS1360: Type '{ ready: "success"; failed: "failed"; pending: "not_attempted"; unsupported: "not_attempted"; absent: "not_attempted"; }' does not satisfy the expected type 'Record<CrawlStatus, TierStatus>'.
read-tier-snapshot.ts:28:9 - error TS7053: Element implicitly has an 'any' type ...
```

Before this change the ladder compiled clean and silently mapped the new member to `not_attempted`.

Worth noting from the same probe: the addition already breaks `@packages/article-state-types:compile` and `@packages/crawl-article:compile` first, via their own exhaustive switches. This closes save-link's gap in that set.

### DIFF
--- a/projects/save-link/src/runtime/domain/crawl-article-state/read-tier-snapshot.ts
+++ b/projects/save-link/src/runtime/domain/crawl-article-state/read-tier-snapshot.ts
@@ -1,7 +1,8 @@
 import type { TierName } from "@packages/hutch-infra-components";
+import type { CrawlStatus as PersistedCrawlStatus } from "@packages/article-state-types";
 
 export type PickedTier = TierName | "none";
-export type CrawlStatus = "ready" | "failed" | "pending" | "unsupported" | "absent";
+export type CrawlStatus = PersistedCrawlStatus | "absent";
 export type TierStatus = "success" | "failed" | "not_attempted";
 
 export type TierSnapshot = {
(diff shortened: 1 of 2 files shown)


FagnerMartinsBrack:

## Context
- **PR**: #1063
- **Commit**: be3d0ff306db648446e2510be1187c1db5d0fa4d

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `TIER_1_STATUSES` is declared *after* the function that indexes it

In `read-tier-snapshot.ts`, `tier1StatusFromCrawlStatus` (line 27) references `TIER_1_STATUSES`, which is declared below it (line 31). This is safe at runtime — the function is only invoked at request time, long after the module's `const` initialises, so there's no temporal-dead-zone hazard, and `pnpm check` (biome included) is clean. It's purely a readability nit: the pattern this PR cites as its model, `TERMINAL_ACTIONS` in `decide-terminal-action.ts`, declares the lookup table *before* the function that indexes it. Moving the `const` above `tier1StatusFromCrawlStatus` would make the two sites read identically. Optional.

PR #1091, opened 2026-08-31
docs(blog-site): add the readlists changelog post

FagnerMartinsBrack:
One changelog post, `sort-your-saves-into-readlists` ("Introducing readlists"), covering the readlist feature end to end. #1095's post is folded in here at the author's request rather than shipping as a second changelog post on the same feature, so this PR adds one file.

Written as a short feature announcement — what the feature is and how it works, without exact caps, field limits, or quoted UI strings.

## What the post announces

- The **All** tab still catches every save, and it replaced the "My Queue" label. It can't be renamed or deleted.
- Filing happens in the reader: the Add to readlist menu lists the lists the article isn't in yet, the article files at the top of the chosen one, a tag appears, the All copy stays put, and the tag takes it back out.
- **A list can start mid-article** — the menu's last row is a name field and a plus, so one submit creates the list with the open article in it. The row shows before you own any list, a name already in use files into that list whatever the casing, All is reserved, the row is a plain form that survives JavaScript being off, and it leaves the menu at the cap.
- The rail handles the housekeeping: create from the plus sign, rename with the pencil, and a per-list delete control whose confirmation offers to move the articles into another list. Deleting leaves the reader where they were.
- A list's address doesn't follow its name, so renaming keeps bookmarks working.
- One read state: marking read from any list flips every copy, with per-list To Read / Read counts.
- The boundary: the iPhone app got the server-driven status tabs (segmented control) but not the rail yet.
- `/queue` and every bookmark are unchanged.
- A closing line teasing automatic placement, with no date attached.

## Verified against code on current `main`

- The picker renders whenever the article holds an All copy — `buildReaderReadlistFiling` in `reader-readlist-filing.ts`, changed by `b59436a`. This corrected a line in the earlier draft claiming the picker stays hidden until you own a list.
- Case-insensitive name match files into the existing list, and the default readlist is refused — `decideReadlistCreate` in `readlist-create.ts`.
- The create row is omitted at the cap, with the write still enforcing it.
- Delete offers migration when another list exists, and keeps the reader in place unless the open list is the one removed — `readlist-delete-confirm.component.ts` and `readlist.delete.route.test.ts`.

## Editorial

- Structural Variety lookback re-run against the 8 newest posts, which now include `save-a-stack-overflow-question-keep-every-answer` (2026-09-01). The opening, TL;DR first sentence, the last two headers, and the closing call to action differ from all 8.
- Title "Introducing readlists" is a deliberate choice by the author, recorded earlier in this PR.
- `tags: ["changelog"]` with its banner, kept as asked. `slug` still equals the filename, so the post keeps its address.
- Same-origin links carry the per-post `utm_source` convention that `afe8a35` swept across the corpus.
- `date` is 2026-09-06: the earlier 2026-08-31 sorted below the 09-01 post and predated the 09-02 work described here.

## Verification

- `pnpm nx run blog-site:check`: 124 tests pass, coverage thresholds met.
- `main` is merged into the branch (96728fd) after it moved 64 commits. The `editorconfig-checker` sandbox failure noted earlier is gone — `41a2257` moved that check to the repo root.

https://claude.ai/code/session_01DEcsNzkKJUYzfRFV1bivft


projects/blog-site/src/runtime/web/pages/blog/posts/sort-your-saves-into-readlists.md
```diff
@@ -0,0 +1,63 @@
+---
+title: "Sort Your Saves Into Readlists You Name"
```
FagnerMartinsBrack: @claude change this to "Introducing readlists"

Also tease that we'll have automated AI placement on readlists soon

FagnerMartinsBrack: Both done in 2def5db.

The title is now "Introducing readlists". The slug stays `sort-your-saves-into-readlists`, so the post keeps its address and its search-facing keywords.

The tease closes the filing section:

> Filing is your decision today. The article and the names you picked are enough for a model to work out where a save belongs, so automatic placement into your readlists is what I'm building next.

It names what is coming without promising a date.

Unrelated note on the red check: the `claude` check on the previous head is the `claude-listener.yml` workflow fired by this review comment ([run 33508268487](https://github.com/Readplace/readplace.com/actions/runs/33508268487)). It failed 3 seconds in with no logs, before running any step, so it is a listener startup failure rather than anything in this diff. The test jobs (`web-tests`, `perf-tests`) were green on that same commit, and I have handled this comment directly, which is what the listener would have done.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1124, opened 2026-09-15
fix(@packages/inbox-store,@packages/test-fixtures): guard Gmail connection updates against partial rows

FagnerMartinsBrack:
## Problem

The ten Gmail-connection lifecycle updates in `dynamodb-gmail-connection.ts` issued bare `UpdateItem` calls with no condition. A `SET`/`REMOVE` on an absent partition key **creates** the row, so a write that arrives after the connection row is deleted (a disconnect, or account erasure from a different Lambda) resurrects a key-only row holding just `userId` plus one attribute.

The read schema requires `gatewayAddress` and `connectedAt`, and every read runs through `schema.parse` — so that partial row makes `findConnectionByUserId` **throw for that reader from then on**, with no way out through the product: the integrations index, the reconnect path, and the disconnect route all read the connection *before* the branch that could heal it, so each surfaces as an unhandled 500.

The write window is real: `RewriteGmailFilterCommand`, `GmailForwardingConfirmedEvent`, and `DisconnectGmailCommand` share one standard SQS queue and one Lambda with no ordering (`batchSize: 1`, no concurrency cap). Removing a sender publishes a rewrite and the Disconnect button publishes a disconnect seconds later, so a late `markRevoked`/`recordFilterError` (each after a Google round-trip) can interleave with the delete for one reader.

No partial row has been observed in production or staging — this shape was read out of the code. The change ships now as a **convention alignment** rather than speculative infrastructure: the repo already pins the identical table shape with `attribute_exists(userId)` (the subscription store), and the sibling discovery store in this same package already funnels writes through the same conditional-swallow helper.

## Change

- Add a `CONNECTION_EXISTS = "attribute_exists(userId)"` constant and an `updateWhenConnected(write)` helper that runs a write, swallows `ConditionalCheckFailedException` as a no-op, and rethrows every other error so the SQS handler still retries.
- Route all ten lifecycle updates (`markForwardingConfirmed`, `clearForwardingConfirmed`, `recordConfirmError`, `recordAccountEmail`, `recordFilter`, `clearFilter`, `recordFilterError`, `markRevoked`, `clearRevoked`, `markDisconnectRequested`) through the helper with the condition attached.
- Leave `createConnection`, `findConnectionByUserId`, `deleteConnection`, and `countConnected` untouched. The store contract is unchanged — every method still returns `Promise<void>`.
- Tests: three store tests (all ten writes carry the condition; a write to a deleted connection resolves; a non-conditional storage error still propagates) and one in-memory-fake test pinning the "never connected" contract across all lifecycle methods, so the two implementations cannot drift apart.

When the row exists — every real path today, including the connect page's `createConnection` → `recordAccountEmail` sequence — behaviour is identical. Only the absent-row path changes, from "resurrect a partial row" to "no-op".

## Testing

- `pnpm nx run @packages/inbox-store:check` — 132 tests pass, **100%** coverage on the store (all three helper branches).
- `pnpm nx run @packages/test-fixtures:check` — pass.
- `pnpm nx run hutch:compile` (typecheck) and `pnpm nx run hutch:lint` — pass; hutch's consumers still compile against the unchanged store signature.
- No infrastructure file is touched, so `pnpm check-infra` is not required.

CI runs the full aggregate checks on this PR (the local pre-commit `pnpm check` could not run in the authoring sandbox — `editorconfig-checker`'s binary download is blocked by proxy policy, and the runtime test suites need env vars the sandbox doesn't provision; `hutch:test-with-coverage` fails identically on the clean baseline there, confirming it is environmental).

https://claude.ai/code/session_01SKKQeQDCbKtL1KeBcPY22U

---
_Generated by [Claude Code](https://claude.ai/code/session_01SKKQeQDCbKtL1KeBcPY22U)_


FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **Run ID**: 34917058439
- **PR**: #1124
- **Branch**: rp/vibrant-bell-opbjmn

## Failed Jobs
- web-tests: Run pnpm check --exclude=ios,android

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Commit**: 1ab41495f497e52b9598ae96452c871cd8f4d0c3

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. The guard is added for an unobserved failure mode — confirm it clears the "Evidence Over Speculation" bar

The PR states plainly that **no partial row has been observed** in production or staging — the shape was read out of the code. CLAUDE.md's first architecture guideline, *Evidence Over Speculation*, discourages adding guards for problems that have not happened and asks that a worthwhile safeguard be proposed *as its own change with the evidence that motivates it*.

This PR reads as an exception to that rule rather than a violation of it, and the justification is strong — worth a human confirming the call, not worth blocking:

- **It is a standalone change** (the "own change" the rule prescribes), scoped to exactly the store + fakes.
- **It aligns with a load-bearing, already-shipped convention rather than inventing new infrastructure.** The subscription store pins the identical single-key table shape with `attribute_exists(userId)` on every lifecycle write, and its `setNextCharge` (`projects/hutch/src/runtime/providers/subscription-providers/dynamodb-subscription-writes.ts:192-209`) documents the *identical* "an `UpdateItem` `SET` on an absent key resurrects a partial row that fails every later read" failure this PR closes. The sibling discovery store already funnels writes through the same swallow-`ConditionalCheckFailedException` helper (`src/packages/inbox-store/src/dynamodb-gmail-discovery.ts:39-48`). So this is better read as removing an inconsistency across siblings than as net-new speculative infra.
- **The failure mode is real and permanent if it ever lands.** `GmailConnectionRow` requires `gatewayAddress` + `connectedAt` (`dynamodb-gmail-connection.ts:22-36`), and every read runs through `schema.parse` via `toConnection`, so one partial row makes `findConnectionByUserId` throw for that reader indefinitely — and the integrations, reconnect, and disconnect paths all read the connection before any branch that could heal it.

No code change requested — this is a policy judgment for the human reviewer to sign off on, which is why it is Low and not Medium (there is a clear, stated reason, so it does not meet the "critical guideline violation with no apparent reason" bar).

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Commit**: f5e78379912cf4855f9150d864b2251ee904ddd4

FagnerMartinsBrack:
### High Priority Issues (Must Fix)

None found

### Medium Priority Issues

None found

### Low Priority Suggestions

#### 1. "Evidence Over Speculation" — a standing judgment call for the human (already raised; no code change)

The PR states plainly that **no partial row has been observed**; the shape was read out of the code. That puts it under CLAUDE.md's *Evidence Over Speculation* guideline, which asks that a safeguard ship *as its own change with the evidence that motivates it*. This PR reads as the **exception the rule prescribes**, not a violation:

- **Behaviour-preserving on every real path.** The connect flow (`gmail-connect.page.ts:153-179`) reads the connection, `createConnection`s only when absent, then always `recordAccountEmail` — so the row exists before any lifecycle write, and the `attribute_exists(userId)` condition is evaluated strongly-consistently against the just-put item. Only the *absent-row* path changes, from "resurrect a partial row" to "no-op".
- **Closes a real divergence rather than adding speculative infra.** The in-memory fake already no-ops absent-row writes (`in-memory-gmail-connection.ts:9`), so DynamoDB's upsert-on-absent was the *outlier* no fake-backed test could exercise.
- **Matches load-bearing precedent** (`dynamodb-subscription-writes.ts:192-208`, `dynamodb-gmail-discovery.ts:39-48`).

Low, not medium: there is a clear, stated reason, so it does not meet the "critical guideline violation with no apparent reason" bar. A policy sign-off for the human, not a code change.

#### 2. `mcp.integration.ts` — the `addedNames` assertion drops *deterministic* order, not just the flaky part

The tie-break flakiness is only between **Work** and **Rust** (same-millisecond `createdAt`, random-slug `localeCompare`). "All" (earliest) and "Reading Group" (created last, in this call) have deterministic positions. The new assertion (`mcp.integration.ts:424-425`) drops *all* ordering; the `readlists` assertion above keeps `readlists[0] === "All"`, so for symmetry `addedNames` could pin the two deterministic ends while staying robust to the Work/Rust tie. Optional and non-blocking — the current form is correct, just slightly looser than the system guarantees.

FagnerMartinsBrack:
@claude address this low prio issue only: 2. mcp.integration.ts — the addedNames assertion drops deterministic order, not just the flaky part
The tie-break flakiness is only between Work and Rust (same-millisecond createdAt, random-slug localeCompare). "All" (earliest) and "Reading Group" (created last, in this call) have deterministic positions. The new assertion (mcp.integration.ts:424-425) drops all ordering; the readlists assertion above keeps readlists[0] === "All", so for symmetry addedNames could pin the two deterministic ends while staying robust to the Work/Rust tie. Optional and non-blocking — the current form is correct, just slightly looser than the system guarantees.

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Branch**: rp/vibrant-bell-opbjmn
- **Attempt**: 1 of 5

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Commit**: fbf96fd6e7ae8c341fd4d30e1fc4a1c86d43ad06

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Commit**: 387a0557b0da8caed2b4ef5ded91c6baa81cc1a8

FagnerMartinsBrack:
### Review of `387a0557` (tie-break tightening on top of the guarded store)

This commit is the `mcp.integration.ts` change requested as LOW #2 in the prior reviews — which the `fbf96fd6` re-review noted had been lost in the race between the tightening run and the conflict-fixer force-push. It has now landed. I re-verified the whole diff against `origin/main`.

**The tie-break fix is correct and now present** (`mcp.integration.ts:423-427`):
```ts
expect(addedNames[0]).toBe("All");
expect(addedNames.slice(1, 3)).toEqual(expect.arrayContaining(["Work", "Rust"]));
expect(addedNames.at(-1)).toBe("Reading Group");
expect(addedNames).toHaveLength(4);
```
It pins the two deterministic ends — **All** (earliest `createdAt`) and **Reading Group** (created last, in this very `add_to_readlist` call) — while leaving only the genuinely-tied **Work/Rust** middle order-independent. This mirrors the `readlists[0] === "All"` assertion above it (`:406`). Crucially, the fix correctly left the cross-array check at `:418` (`expect(article.readlists).toEqual(readlists)`) as an **exact** `toEqual`: both arrays are sorted from the same fixed slugs within a run, so that comparison is stable and does *not* need loosening. Right call — it loosened only the hard-coded literal expectations, not the live-vs-live cross-check.

**The store guard (unchanged since `fbf96fd6`) remains sound.** I re-confirmed the one thing that could have been a silent no-op: `defineDynamoTable.update` (`define-table.ts:136-140`) spreads its argument into `new UpdateCommand({ TableName, ...input })`, so `ConditionExpression` reaches the SDK — the guard is real in production, not just captured by the test fake. All **10** lifecycle writes carry `attribute_exists(userId)`; `updateWhenConnected` (`dynamodb-gmail-connection.ts:64-71`) swallows **only** `ConditionalCheckFailedException` and rethrows everything else, so a transient storage error still fails the SQS record and retries. Because the condition is purely `attribute_exists(userId)`, a swallowed rejection can only ever mean "row gone" — no ambiguity in what is no-op'd. Both helper branches, the ten-write condition pin, and the in-memory "never connected" contract across all ten methods are covered.

### High Priority Issues (Must Fix)

None found

### Medium Priority Issues

None found

### Low Priority Suggestions

#### 1. "Evidence Over Speculation" — standing human sign-off (already raised three times; no code change)

Unchanged from the prior reviews and noted here only for completeness. The PR states plainly that **no partial row has been observed**; the shape was read out of the code. That places it under CLAUDE.md's *Evidence Over Speculation* guideline, which asks that a safeguard ship *as its own change with the evidence that motivates it*. This PR reads as the **exception the rule prescribes**, not a violation: it is behaviour-preserving on every real path (only the absent-row path changes, from "resurrect a partial row" to "no-op"); it closes a real divergence (the in-memory fake already no-ops absent-row writes, so DynamoDB's upsert-on-absent was the outlier); and it matches load-bearing precedent (`dynamodb-subscription-writes.ts` pins the identical single-key shape with `attribute_exists(userId)`, and the sibling `dynamodb-gmail-discovery.ts` uses the same swallow helper). Low, not medium: there is a clear, stated reason, so it does not meet the "critical guideline violation with no apparent reason" bar. This is a policy judgment for the human to sign off, not a code change.

FagnerMartinsBrack:

## Context
- **PR**: #1124
- **Commit**: d0a34c1bb0585cacf4917263fa896e2c53128668

FagnerMartinsBrack:
### Review of `d0a34c1b` — tie-break ends now pinned on top of the guarded store

**The tie-break fix is correct** (`mcp.integration.ts:423-427`). Traced end to end: `readerReadlists` prepends `DEFAULT_READLIST` so `"All"` is structurally index 0; both list stores sort by `createdAt.getTime()` then `slug.localeCompare` (`dynamodb-readlist-definitions.ts:71-73`, `in-memory-article-store.ts:685`) with random slugs (`generate-readlist-slug.ts:5`), under a real wall-clock fixture (`fixture.ts:496`). So `"Work"`/`"Rust"` tie on the same millisecond (order-independent) while `"Reading Group"`, created ~9 round-trips later, reliably sorts last. Pinning `[0]` and `.at(-1)` and leaving `slice(1,3)` as `arrayContaining(["Work","Rust"])` pins exactly what the system guarantees; the live-vs-live `toEqual` at `:418` is correctly left exact.

**The store guard remains sound.** `defineDynamoTable.update` spreads into `new UpdateCommand({ TableName, ...input })` (`define-table.ts:137`) so `ConditionExpression` reaches the SDK; `ConditionalCheckFailedException` is the real re-export (`index.ts:13`). All 10 lifecycle writes carry `attribute_exists(userId)`; `updateWhenConnected` swallows only that exception and rethrows the rest (SQS still retries transient errors). Both helper branches, the ten-write pin, and the in-memory "never connected" no-op are covered; the two implementations agree.

### High Priority Issues (Must Fix)

None found

### Medium Priority Issues

None found

### Low Priority Suggestions

#### 1. "Evidence Over Speculation" — standing human sign-off (raised in prior reviews; no code change)

The PR states no partial row has been observed. This reads as the exception the rule prescribes, not a violation: a standalone change, behaviour-preserving on every real path (only the absent-row path changes to a no-op), closing a real divergence with the in-memory fake, and matching load-bearing precedent (`dynamodb-subscription-writes.ts:101-200`, `:193`; `dynamodb-gmail-discovery.ts:45`). A policy judgment for the human to sign off, not a code change.

FagnerMartinsBrack:
Closing after the local review: the conditional writes guard a partial-row race that has not been observed (Evidence Over Speculation), and the MCP readlist tie-break fix is unrelated to this PR's purpose and belongs in its own.

PR #1129, opened 2026-09-15
feat(hutch): run the Gmail filter, disconnect and scan-failure paths in local development

FagnerMartinsBrack:
## Why

Connecting Gmail is part of Readplace's self-serve acquisition surface, but a developer working the Gmail page locally was half-blind. With real `GMAIL_INTEGRATION_*` credentials exported, the dev composition ran the sender **scan** for real yet **stubbed every other Gmail command** (`publishRewriteGmailFilter: async () => {}`, `publishDisconnectGmail: async () => {}`). Three consequences, straight from the code:

1. **Disconnect never finished.** The route set `disconnectRequestedAt` and published the stub; only `initDisconnectGmail` clears the row and revokes the grant at Google — and it never ran. The page sat on "Disconnecting…" forever, and a restart dropped the in-memory row while Google still listed the grant.
2. **Mapping a sender changed nothing outside the app.** No filter was created/removed at Gmail, so the `filtering` and `filter-failed` states were unreachable locally.
3. **A dead scan stayed "running" forever.** In production a crashed scan is retried and finally marked failed by the dead-letter consumer with reader-facing copy; locally the rejection was only logged, so the page showed "Checking Gmail messages…" indefinitely.

This PR makes local development exercise the **same Gmail state machine as production**, in-process. No deployed behaviour changes.

## What changed

All changes are in the `hutch` project — the local composition and the `domain/gmail` modules it wires.

- **Shared the stalled-scan message** — `GMAIL_DISCOVERY_PAUSED_MESSAGE` is now exported from `discover-gmail-senders.ts` and used by the dead-letter handler in place of its literal, so the local and deployed paths can't drift (message text unchanged).
- **`run-gmail-discovery-locally.ts`** — runs the paged scan; on a mid-scan crash it re-reads the checkpoint and, only if still `running`, marks it failed under its current generation with the shared message, then logs. Settled scans are left alone (module guard + the store's generation fence). Start still returns immediately; the run stays detached, so the page polls as before.
- **`local-gmail-commands.ts`** — turns `RewriteGmailFilter`/`DisconnectGmail` into the `publishRewriteGmailFilter`/`publishDisconnectGmail` seams the Gmail bundle expects, running them in-process and logging at the **same severities the deployed handlers use** (rewrite: ok→info, unavailable→warn, else→error; disconnect: ok→info, unavailable→warn, not-connected→warn). The domain records outcome state on the connection row; the un-consumed events simply aren't published locally.
- **`confirm-on-connect-gmail-connection.ts`** — Option A from the plan: a thin `GmailConnectionStore` wrapper whose `createConnection` also marks the connection and gateway address confirmed (exactly the pair of writes the deployed forwarding-confirmed handler makes), so local dev reaches `ready-to-filter` / `filtering` / `filter-failed` instead of being stuck at "Step 2 of 2" waiting for a confirmation email that can't arrive locally.
- **`dev-providers.ts`** — replaced the two stubs and the inline scan block with real construction of `initGmailFilters` → `initRewriteGmailFilter` → `initDisconnectGmail`/`initRevokeGmailGrant` and the three new modules, sharing **one** sender store and **one** access token across rewrite, disconnect, and discovery. The composition root stays pure wiring (no branches) under its whole-file c8-ignore.

Every module with a branch has a unit test beside it. No deployed behaviour, infra, route, copy, command, or coverage/lint/knip exclusion changed. The two deployed files touched carry the shared message only: `discover-gmail-senders.ts` gains the exported `GMAIL_DISCOVERY_PAUSED_MESSAGE` constant, and the dead-letter handler (`gmail-discovery-handler.ts`, the `gmail-discovery-dlq` Lambda) reads it in place of its own literal — identical text, identical behaviour. No new comments.

## Notes for review

- **Reproduction precondition (decision gate).** The plan asks that all three behaviours be reproduced on a real local run before merge — that needs real `GMAIL_INTEGRATION_*` credentials (shipped empty in `.envrc`) and can't be done in the build sandbox. Worth confirming against a real Gmail account before merging; drop any step that doesn't actually reproduce.
- **Option A trade-off.** Gmail only forwards to an address verified inside that account, so a filter-create for a locally minted gateway may be **rejected**, landing the page on "Needs attention" with Gmail's own message. That's a faithful local error surface, not a working forward — as the plan flags.

## Verification

`pnpm nx run hutch:check` passes — tsc, biome lint, knip, unused-CSS, and jest under coverage (100% functions; statements/branches/lines above the hutch thresholds). Rebased onto the latest `main` (which shipped a Gmail-scan bounding change to `discover-gmail-senders.ts`) and re-verified green.

The native `android:check` / `ios:check` targets can't run in the Linux build environment (no Android SDK / `xcodegen`) and are unaffected by this change; CI runs those on their own runners.

https://claude.ai/code/session_01Nitr8W2ir3PgTJknCrDkQa

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nitr8W2ir3PgTJknCrDkQa)_


FagnerMartinsBrack:

## Context
- **PR**: #1129
- **Commit**: 682d06ad7b16d2e6806d996be156273f77466dce

FagnerMartinsBrack:

None found

## Medium Priority Issues

None found

## Low Priority Suggestions

### 1. `capturingLogger` / `LoggedCall` duplicated verbatim across the two new test files
`capturingLogger()` and the `LoggedCall` type are byte-identical in `local-gmail-commands.test.ts:10-19` and `run-gmail-discovery-locally.test.ts:12-21`. Extracting them into one shared test helper would keep them from drifting. This is genuinely optional — bespoke per-file capturing loggers are already the norm in this suite (`changelog-banner-source.test.ts:40`, `middleware/error-handler.test.ts:10`), so the duplication is consistent with existing convention rather than a new deviation.

FagnerMartinsBrack:

FagnerMartinsBrack:
Guideline fixes committed in cfeceb1:

**PR description** — corrected the false sentence at the end of "What changed". Replaced "No deployed code, infra, route, copy, command, or coverage/lint/knip exclusion changed." with a claim scoped to behaviour that names the deployed files the PR actually touches: `discover-gmail-senders.ts` (gains the exported `GMAIL_DISCOVERY_PAUSED_MESSAGE` constant) and `gmail-discovery-handler.ts` (the `gmail-discovery-dlq` Lambda, which now reads it instead of its own literal) — identical text, identical behaviour. I widened the finding's suggested wording from "the one deployed file touched" to both files, because `discover-gmail-senders.ts` is also deployed and also changed, so asserting "one" would have introduced a new inaccurate claim. Nothing else in the description was altered.

**Commit message** — amended the single PR commit's body: "No deployed behaviour changes: only the local composition and the domain/gmail modules it wires." became "No deployed behaviour changes: the local composition, the domain/gmail modules it wires, and the dead-letter handler, which now reads the same paused-scan message from a shared constant (text unchanged)." Message-only amend — the tree is identical to 7b97f3339 (`git diff --stat 7b97f3339 HEAD` is empty), the parent (86e633ace) and author metadata are unchanged, and the Co-Authored-By / Claude-Session trailers are preserved.

FagnerMartinsBrack:

## Context
- **PR**: #1129
- **Commit**: cfeceb17017d5e2c7e576aa1956024a746fd1c22

FagnerMartinsBrack:
Local Pulumi preview (read-only) of this branch at cfeceb1, rebased on main 86e633a, compared with the same previews run on main:

- **hutch staging and prod**: the only changes this PR causes are rebuilt Lambda code archives for `gmail-discovery-handler` and `gmail-discovery-dlq-handler`, because `GMAIL_DISCOVERY_PAUSED_MESSAGE` now lives in `discover-gmail-senders.ts` and esbuild does not inline it across modules. The message text and the behaviour are unchanged, so a deploy rolls both functions with no behavioural difference.
- No create, replace or delete on either stack.
- Every other update in the previews appears identically in the previews run on main (main commits not yet deployed, and environment values the local shell does not hold), so none of it comes from this PR.

The reproduction gate under "Notes for review" is still open: the three behaviours have not been exercised against a real Gmail account.

PR #1132, opened 2026-09-16
Make `pnpm check` runnable outside the self-hosted CI runner

FagnerMartinsBrack:
A bare `pnpm check` — what `.husky/pre-commit` runs — could not complete on any machine without the native toolchain, and could not start at all in a Claude Code web session. Both failures were reproduced on `main` before anything was changed; neither is a regression from recent work.

## Two observed failures

**1. Native check targets abort the run.** `pnpm check` fans out to all 51 projects, `ios` and `android` included. Reproduced on this Linux checkout at `main`:

```
nx run ios:check      → ./scripts/generate-project.sh: line 13: xcodegen: command not found
                        make: *** [Makefile:15: generate] Error 127   (exit 1)
nx run android:check  → ax.sh: no Android SDK found at $ANDROID_HOME or ~/Library/Android/sdk
                        make: *** [Makefile:20: test] Error 69        (exit 1)
```

`--exclude=ios,android` lives only in `ci.yml`, so every other caller of `pnpm check` inherits the problem: the pre-commit hook, and `claude-listener.yml`, whose prompt files instruct a bare `pnpm check` on an `ubuntu-latest` runner with no Xcode, no Android SDK and no JDK.

**2. `editorconfig-checker` cannot fetch its binary in a web session.** The npm package is a JS wrapper that resolves its Go binary at run time via Octokit against `api.github.com`:

```
GET /repos/editorconfig-checker/editorconfig-checker/releases/tags/v3.8.0 - 403
Failed to download binary: GitHub access to this repository is not enabled for this session.
 ELIFECYCLE  Command failed with exit code 1.
```

A web session's GitHub access is scoped to this repository alone. The wrapper hardcodes `new Octokit({auth: process.env.GITHUB_TOKEN})` against the default base URL and exposes no mirror or override, so `pnpm check` dies at its first gate before nx runs.

## Changes

**`projects/native-apps/{ios,android}/project.json`** — each `check` target runs its suite when the toolchain is present, and otherwise prints a `SKIPPED` line naming what was not tested.

- macOS is unchanged. `uname -s` is `Darwin`, so `make test` runs exactly as before and the native suites stay in pre-commit.
- CI is untouched. `web-tests` still passes `--exclude=ios,android`, so these targets are never scheduled there; `ios-tests` and `android-tests` invoke `make test` directly rather than through nx.

**`.claude/hooks/session-start.sh`** (new) — downloads the pinned `editorconfig-checker` release asset over plain HTTPS into the cache directory the wrapper reads. It returns immediately unless `CLAUDE_CODE_REMOTE` is `true`, so CI and local development are unaffected: CI already resolves the binary with a real `GITHUB_TOKEN` plus a persisted copy on the self-hosted runner, and a developer machine reaches `api.github.com` normally.

## One thing left for you to do

The hook is committed but **not wired up** — registering it was blocked here as a self-modification, which seems right: it makes Claude auto-execute a script every session, so the call belongs to you. To enable it, add this to `.claude/settings.json`:

```json
  "hooks": {
    "SessionStart": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "$CLAUDE_PROJECT_DIR/.claude/hooks/session-start.sh"
          }
        ]
      }
    ]
  },
```

Until then a web session must still seed the binary by hand. Everything else in this PR works as-is.

## A latent silent-failure found while building this

Worth flagging, because it nearly shipped. The wrapper decides the binary is ready by stat-ing the **version directory alone**, and its error path exits 0 through an unhandled rejection. So a half-populated directory makes `pnpm check` skip lint entirely *and still report success*. Reproduced directly:

```
$ mkdir -p <pkg>/bin/v3.8.0            # empty
$ pnpm exec editorconfig-checker
Error: ENOENT: no such file or directory, scandir '.../bin/v3.8.0/bin'
EXIT_CODE=0                            # ← green, having checked nothing
```

The first draft of the hook created exactly that state on an interrupted download. It now builds in a staging directory and renames it into place, and tests for the binary rather than the directory — which also recovers a checkout already poisoned this way. Verified: the old `[ ! -d dir ]` guard skips (silent green), the new `[ ! -x binary ]` guard re-downloads and recovers.

## Verification

- Bare `pnpm check`, no flags and no env overrides: **exit 0**, `Successfully ran target check for 51 projects and 64 tasks they depend on`.
- Both commits passed the real `.husky/pre-commit` gate, which runs that same bare `pnpm check` — the hook that previously could not run here.
- `editorconfig-checker` passes the repo with zero findings; hook recovery path exercised end to end against the poisoned state, leaving no staging directory behind.
- Confirmed the duplicate-flag trap that shaped this design: `--exclude=ios,android` passed **twice** silently re-includes both projects (51 projects, not 49), which is why the flag stays in one place only.

## Notes and trade-offs

- **The skip is loud, never silent.** A skipped native target prints what was not tested; nothing reports green for work it did not do.
- **Rejected alternative:** moving `--exclude=ios,android` into the root `check` script. It reads more cleanly, but drops the native suites from pre-commit on macOS, and would have required deleting the flag from `ci.yml` in the same commit or the exclusion silently evaporates.
- **Residual risk, deliberately not engineered around:** these targets are cached, and their inputs record `CLAUDE_CODE_REMOTE` but not the platform, so a skip could in principle replay after a toolchain is installed. No nx remote cache is configured and CI checks out clean per job, so a cache entry cannot cross machines. Building for this would be the speculative infrastructure `CLAUDE.md` rules out, so it is called out here instead.
- **Not addressed:** `pnpm check-infra` runs `nx run-many --target=compile --all`, and the `ios` `compile` target needs Xcode, so it has the same shape of problem. It is credential-gated and out of scope here.
- `.claude/hooks/session-start.sh` is a new script; `CLAUDE.md` asks for approval before adding one. Flagging it explicitly for that reason.

https://claude.ai/code/session_01QFkB6YUxFeR22nCZBRpMrq

---
_Generated by [Claude Code](https://claude.ai/code/session_01QFkB6YUxFeR22nCZBRpMrq)_


FagnerMartinsBrack:

## Context
- **PR**: #1132
- **Commit**: 65bc17a0b93a548db0aa11a11e30ba3652fd425e

FagnerMartinsBrack:
### Code Review — PR #1132

Solid PR: both failures were reproduced on `main` first, the editorconfig hook's staging-and-rename design correctly dodges the upstream silent-failure (verified against the wrapper source), and the trade-offs are disclosed honestly. The iOS gate is robust. One substantive gap on the Android side, plus low-priority notes.

None found.

## Medium Priority Issues

### 1. The Android skip does **not** fire on the `ubuntu-latest` runners the PR names as motivation

The PR's premise — *"claude-listener.yml … a bare `pnpm check` on an `ubuntu-latest` runner with no Xcode, no Android SDK and no JDK"* — is inaccurate for the Android half. GitHub-hosted `ubuntu-latest` ships the Android SDK and a JDK, so the skip is never taken there.

Evidence, gathered on this github-hosted `ubuntu` runner (same image family as `ubuntu-latest`):

```
ANDROID_HOME=/usr/local/lib/android/sdk        (platform-tools present)
JAVA_HOME=/usr/lib/jvm/temurin-17-jdk-amd64    (javac 17.0.x)
$ ./scripts/ax.sh true ; echo $?  →  0         ⇒ android:check runs `make test`, does NOT skip
```

`projects/native-apps/android/project.json:20` gates on `./scripts/ax.sh true`; `ax.sh:44-70` only checks that `$ANDROID_HOME/platform-tools` exists and *some* `javac` is present — both true on these runners. All four automation workflows that run bare `pnpm check` are `runs-on: ubuntu-latest` (`claude-listener.yml:54`, `claude-PR-CI-failure-fixer.yml`, `claude-PR-code-review-auto-apply.yml`, `claude-PR-conflict-fixer.yml`), so `android:check` still runs `make test` there — a no-op, not a fix. Compounding: `resolve_java_home` (`ax.sh:27-31`) is version-blind and accepts JDK 17, while the project pins JDK 21 (`ax.sh:32,62`), so a passing probe still doesn't guarantee `make test` passes. Net: the change reliably helps only where the SDK is genuinely absent (the web session you reproduced, and minimal dev machines). iOS is unaffected — `uname -s` is `Linux` on any ubuntu runner.

## Low Priority Suggestions

### 1. Cached skip can replay after a toolchain is installed
`check` is cached (`nx.json` `targetDefaults.check`: `cache:true`, inputs lack platform), so a same-machine skip→install→re-run hits the cache. nx replays the cached stdout so `SKIPPED` still prints (stays loud, just stale); `nx reset` or a source edit forces a real run. A `{ "runtime": "uname -sm" }` input would close it non-speculatively if desired.

### 2. `SKIPPED` message under-describes the guard
`android`'s message says *"no Android SDK"*, but the probe also fails without a JDK. Consider *"no usable Android toolchain."*

### 3. Pre-commit behavior change on macOS without the Android SDK
Previously a hard failure, now a skip — a macOS behavior change, though the PR frames macOS as unchanged.

### 4. Disclosed, flagging for the checklist
Editorconfig fix is inert until `.claude/settings.json` gets the `SessionStart` block; the new `.claude/hooks/session-start.sh` needs the human approval CLAUDE.md requires; minor: the `echo …@*/…` glob in `session-start.sh:14` assumes a single match (fails loud otherwise) — the `node_modules/editorconfig-checker` symlink is simpler.

FagnerMartinsBrack:

## Context
- **Attempt**: 1 of 5
- **PR**: #1132
- **Branch**: claude/pnpm-check-failures-66zcnh

## Review to Address

FagnerMartinsBrack:

## Context
- **PR**: #1132
- **Commit**: 2663f3cd99db76669eb6605a55afaa58518a6466

